PR #4, opened 2025-07-02
Modified flake8, fixed outstanding issues

allenwang28:
Fixing `pre-commit run --all-files`

- Copies .flake8 from torchtitan
- Fixed a few non-existing variables etc. (thanks DevMate)

### DIFF
--- a/forge/models/llama3/model/args.py
+++ b/forge/models/llama3/model/args.py
@@ -62,7 +62,7 @@ class TransformerModelArgs(BaseModelArgs):
             if isinstance(m, nn.Embedding)
         )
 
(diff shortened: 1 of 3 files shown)


forge/models/llama3/model/args.py
```diff
@@ -62,7 +62,7 @@ def get_nparams_and_flops(self, model: nn.Module, seq_len: int) -> tuple[int, in
             if isinstance(m, nn.Embedding)
         )
 
-        l, h, q, t = (
```
pbontrager: Why make this change?

PR #5, opened 2025-07-02
Adds monarch stack with examples and tests

allenwang28:
Changes:
- Adds in `monarch_utils` to forge with the `stack` implementation. I had a diff in phabricator for this, but was recommended to iterate directly in Forge to start with
- Adds in a unit test for monarch_utils, with some basic tests for `stack`
- Introduces an `examples` folder to showcase how the API works

Example Usage:
```
import asyncio

from forge.monarch_utils.stack import stack

from monarch.actor_mesh import Actor, endpoint
from monarch.proc_mesh import proc_mesh


class Environment(Actor):
    @endpoint
    async def step(self):
        print("env step")


class BrowserEnv(Environment):
    @endpoint
    async def step(self):
        print("browser step")


class CodingEnv(Environment):
    @endpoint
    async def step(self):
        print("coding step")


async def main():
    m = await proc_mesh(gpus=1)
    browser = await m.spawn("browser", BrowserEnv)
    coding = await m.spawn("coding", CodingEnv)
    # note that interface can be deduced, but adding it explicitly
    # helps the type checker
    envs = stack(browser, coding, interface=Environment)
    await envs.step.call()


if __name__ == "__main__":
    asyncio.run(main())
```

Tests:

```
$ pytest ./tests/unit_tests/test_monarch_utils.py 
============================================================================================================= test session starts ==============================================================================================================
platform linux -- Python 3.10.18, pytest-8.4.1, pluggy-1.6.0
rootdir: /home/allencwang/forge
configfile: pyproject.toml
plugins: typeguard-4.4.4, timeout-2.4.0, cov-6.2.1, asyncio-1.0.0
asyncio: mode=strict, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 10 items                                                                                                                                                                                                                             

tests/unit_tests/test_monarch_utils.py _1qst1euKrcAv
..........                                                                                                                                                                                        [100%]

============================================================================================================== 10 passed in 1.38s ==============================================================================================================
```

### DIFF
--- a/examples/stacked_envs.py
+++ b/examples/stacked_envs.py
@@ -6,7 +6,7 @@
 
 """Example of using forge.monarch APIs.
 
-Run this with python -m examples.monarch_stack
+Run this with python -m examples.stacked_envs
 
 """
 


examples/stacked_envs.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
pbontrager, line 1: Can you move this inside of a `forge/rl` folder so we can just stay contained inside of there?

forge/monarch_utils/README.md 
```diff
@@ -0,0 +1,31 @@
+# Monarch Utils
```
pbontrager, line 1: Do we want a whole folder here, or just components/monarch.py where you can put all of these components?

forge/monarch_utils/README.md 
```diff
@@ -0,0 +1,31 @@
...
+
+This folder contains experimental APIs built for Forge that may eventually graduate into Monarch proper.
```
pbontrager, line 3: I think we can include these examples either in the doc string to get picked up by whatever docs situation we setup. Or put a top level docs folder with md files for things the way titan does.

PR #6, opened 2025-07-03
Creates scaffolding for a runnable RL loop

allenwang28:
This PR was my (and devmate's) imagination of the implementation of https://colab.research.google.com/drive/11S8tVZuqgNFYIXihooGedhGfCMceJv0N#scrollTo=ng46micDhhwT.

This translates a few of the ideas into that notebook into something runnable, which you can see in examples/rl_main.py.

This is very WIP right now - just wanted to share the initial approach to get early feedback to hopefully land next week?

If you want to try it yourself, it runs with: 
```
python -m examples.rl_main
```

It's very noisy due to Monarch logs - a known thing that should hopefully be resolved soon(TM)

sample logs: P1858513059

Specifically, what this PR does is:
- Introduces common entities, and a toy implementation of them, i.e. a collector, policy, environment, rewarder, replay buffer, etc.
- Tests for those components
- Showcases, in examples/rl_main, an end-to-end running loop showcasing the dataflow and interaction between these components. No trainer yet - this just shows that samples can be obtained from the replay buffer.

### DIFF
--- a/forge/rl/interfaces.py
+++ b/forge/rl/interfaces.py
@@ -17,7 +17,7 @@ a generic orchestrator that can work with any component that implements these.
 """
 
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from monarch.actor import Actor, endpoint
 
@@ -32,20 +32,8 @@ from forge.rl.environments.base import Action, Observation, State
 class Trajectory:
     """A trajectory containing a sequence of states, actions, etc."""
 
-    states: list[Observation] | None = None
-    actions: list[Action] | None = None
-    dones: list[bool] | None = None
-    infos: list[dict] | None = None
-
-    def __post_init__(self):
-        if self.states is None:
-            self.states = []
-        if self.actions is None:
-            self.actions = []
-        if self.dones is None:
-            self.dones = []
-        if self.infos is None:
-            self.infos = []
+    states: list[Observation] = field(default_factory=list)
+    actions: list[Action] = field(default_factory=list)
 
 
 @dataclass
--- a/tests/unit_tests/test_rl_components.py
+++ b/tests/unit_tests/test_rl_components.py
@@ -134,8 +134,6 @@ class TestReplayBuffer:
             ToyObservation(step=0, data=torch.tensor([0.0]), text="Step 0, Value: 0.0")
         ]
         trajectory.actions = [ToyAction(data=torch.tensor([1.0]))]
-        trajectory.dones = [False]
-        trajectory.infos = [{"test": "info"}]
 
         # Add trajectory to buffer
         await buffer.extend.choose(trajectory)
@@ -174,8 +172,6 @@ class TestReplayBuffer:
                 )
             ]
             trajectory.actions = [ToyAction(data=torch.tensor([float(i + 1)]))]
-            trajectory.dones = [i == 4]
-            trajectory.infos = [{"trajectory_id": i}]
 
             await buffer.extend.choose(trajectory)
 
@@ -239,7 +235,6 @@ class TestCollector:
         # Check that trajectory has expected structure
         assert hasattr(trajectory, "states")
         assert hasattr(trajectory, "actions")
-        assert hasattr(trajectory, "dones")
 
         # Verify trajectory contains data (should have at least one step)
         assert trajectory.states and len(trajectory.states) > 0


examples/rl_main.py
```diff
@@ -0,0 +1,227 @@
...
+    # lets us customize our routing. But ultimately we will want to package this
+    # with our environment definition somehow.
+    # This toy rewarder just adds 1 to the next state value.
+    rewarder = await rewarder_procs.spawn("rewarder", ToyRewarder)
+
+    generator_config = GeneratorConfig(
+        max_generator_steps=5,
+    )
+
+    browser_generators = await browser_procs.spawn(
```
Darktex: Let's not use the word "generator". It's not a thing in RL, we were just influenced from the LlamaRL team. Add it to the banned words like "cabernet" ;)

examples/rl_main.py
```diff
@@ -0,0 +1,227 @@
...
+    # TODO - add in an example of a "vLLM executor" and "vLLM controller"
+    # This policy just generates something between -2. and 2.
+    policy = await policy_procs.spawn("policy", ToyPolicy, action_range=(-2.0, 2.0))
+
+    # Practically, we should create a single rewarder that's shared across the actors
+    # in its associated environment mesh. This provides cleaner decoupling and
+    # lets us customize our routing. But ultimately we will want to package this
+    # with our environment definition somehow.
+    # This toy rewarder just adds 1 to the next state value.
+    rewarder = await rewarder_procs.spawn("rewarder", ToyRewarder)
```
Darktex: We probably won't have this abstraction, but ok for now

examples/rl_main.py
```diff
@@ -0,0 +1,227 @@
...
+        sample_count = 0
+        while True:
+            try:
+                await asyncio.sleep(3)  # Wait a bit before sampling
+
+                # Check if buffer has enough data
+                buffer_length = await replay_buffer.len.choose()
+                if buffer_length < SAMPLES_PER_BATCH:
+                    print(
+                        f"📦 Replay buffer has {buffer_length} trajectories, waiting for at least {SAMPLES_PER_BATCH}..."
```
Darktex, line 122: Cool emojis

examples/rl_main.py
```diff
@@ -0,0 +1,227 @@
...
+                )
+
+                # Wait a bit before next episode
+                await asyncio.sleep(2)
+
+            except Exception as e:
+                print(f"❌ Error in episode {episode_count + 1}: {e}")
+                await asyncio.sleep(1)
+
+    async def replay_buffer_sampler_task():
```
Darktex, line 111: We probably want to run some structured concurrency and explicitly set how many copies from the get-go

forge/rl/interfaces.py
```diff
@@ -0,0 +1,152 @@
...
+class ForgeRequest:
+    """A request containing state/observation data."""
+
+    data: torch.Tensor | None = None
+    text: str | None = None
+    metadata: dict | None = None
+
+
+@dataclass
+class ForgeTrajectory:
```
Darktex: So, is this a batch?

forge/rl/interfaces.py
```diff
@@ -0,0 +1,152 @@
...
+import torch
+from monarch.actor_mesh import Actor, endpoint
+
+
+# ==== Data Structures ====
+# Of course, these data structures can be replaced by almost anything.
+
+
+@dataclass
+class ForgeRequest:
```
Darktex: What's the rationale for this class?

examples/rl_main.py
```diff
@@ -0,0 +1,227 @@
...
+        config=generator_config,
+        policy=policy,
+        replay_buffer=replay_buffer,
+        environment_creator=partial(
+            ToyEnvironment, name="coding", max_steps=5, rewarder=rewarder
+        ),
+    )
+
+    # Here's our stack API in action!
+    generators = stack(
```
Darktex: I find this API super confusing currently. I'm not really able to read this code.

1. Where do I specify how many resources *one* element needs? Eg **one** VLLM server for DeepSeek needs like 16 GPUs.
2. Similarly, where do I specify how many copies in the pool I have? I want to be able to see if I'm routing between copies or not.

examples/rl_main.py
allenwang28: That's fine by me - in the notebook we had called it `Env`, but IMHO that name is inaccurate because this construct that I call "Generator" is responsible for coordinating the data flow between the policy and the environment. Since its role is to "generate rollouts" (or trajectories) I called it a generator, but happy to change it if you have any other suggestions?

examples/rl_main.py
allenwang28: This is a reasonable concern, and I think the confusion is that there's a lack of documentation / transparency around how the proc/actor APIs fundamentally work. 

I'll take a pass at putting together a resource that explains it, putting the current `stack` API in context, and how we would do something like spin up multiple vLLM servers since I think this will be a common confusion.

examples/rl_main.py
allenwang28: I fear this may be a "Monarch's APIs are not immediately obvious" thing, but what we practically do in this code sample is create 14 processes for `generators` (2 for browsers, 4 for deep research, 8 for coding) 

Then when we say `results = await generators.run_episode.call()`, what we're doing is executing *all* of these processes in parallel. We want to do this because it's using Rust processes that bypass the GIL

That means this code snippet:
```
   async with TaskGroup(limit=2_000) as spawn_ep:  # <--- Run 2000 envs at the same time
```

is inherently accomplished by the `ActorMesh` we create

forge/rl/interfaces.py
allenwang28: This is an individual trajectory

forge/rl/interfaces.py
allenwang28: Tbh it's just a stand-in for whatever we want to call "state" - maybe I should've called it `EnvState` or something similar

examples/rl_main.py
allenwang28: Do you mean as a top-level abstraction in the main? If so, I agree that it's probably not needed to expose in the top-level

examples/rl_main.py
Darktex: Ah, yeah this would be a Collector (at least if we use the TorchRL terminology)

examples/rl_main.py
Darktex: Yeah, we may not do a rewarder class but that is a minor thing

examples/rl_main.py
Darktex: I'm willing to see it before judging of course, but IMHO keeping the confusing API and writing documentation around it doesn't seem all that worth it vs simply having a clearer API :D We can definitely bring this up at the next Monarch sync or we can land some syntactic sugar here for now while we discuss how to land it in Monarch (so by the time we launch we can use the Monarch version)

examples/rl_main.py
Darktex: Hmm, but it's not the same thing, right? 

If we go the TaskGroup route, we have an infinite loop where as soon as one env is done, the next env starts running. So this is going to implement our "infinite producer" loop with back-pressure to avoid producing too much.

Instead, if we run an await on the whole pool of workers, after one worker is done, we do not automatically get the next one in, so we don't implement the same infinite loop, right?

examples/rl_main.py
allenwang28: Switched the name to collector

examples/rl_main.py
allenwang28: (We discussed this live as a follow up to this PR)

forge/monarch_utils/stack.py
```diff
@@ -158,12 +158,12 @@ def _common_ancestor(*actors: ActorMeshRef) -> Optional[Type]:
 
 
 @overload
-def stack(*actors: Any, interface: Type[T]) -> StackedActorMeshRef[T]:
+def stack(*actors: Any, interface: Type[T]) -> StackedActorMeshRef:
```
pbontrager, line 161: Can we make this private? I like this implementation but want to limit the number of concepts we expose to the user. Right now we have:
- proc mesh
- actor mesh
- stacked actor mesh
- stacked actor mesh ref

In my mind, ideally this would all be one mesh. Or as close to that as we can get without mixing too many concepts.

examples/rl_main.py
```diff
@@ -0,0 +1,229 @@
...
+Run this with:
+    python -m examples.rl_main
+"""
+
+import asyncio
+from functools import partial
+
+from forge.monarch_utils.stack import stack
+from forge.rl.collector import Collector
+from forge.rl.config import CollectorConfig
```
pbontrager: I think we should leave config out of scope for now. Let's just write everything without a config and add that in once it's been implemented elsewhere in the library.

examples/rl_main.py
```diff
@@ -0,0 +1,229 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""A working example showcasing a practical example of forge with RL.
+
+Run this with:
+    python -m examples.rl_main
+"""
+
```
pbontrager, line 12: At a high level, I think what. I want as a user is a quick way to understand all of the concepts we're introducing and what are the main things I, as the user, am meant to extend. I think there are two main groups, the actors that we provide and the data adjacent stuff like environments and rewards. Maybe the clearest way to expose this is:

```
rl/
  collector.py
  replay_buffer.py
  interfaces.py
  trainer.py
  ...

data/
  types/
    ...
  datasets/
    ...
  environments/
    toy.py
  rewards/
    toy.py
```

examples/rl_main.py
```diff
@@ -0,0 +1,229 @@
...
+from forge.rl.replay_buffer import ReplayBuffer
+from forge.rl.rewards import ToyRewarder
+
+from monarch.actor import proc_mesh
+
+# Configuration constants
+SAMPLES_PER_BATCH = 4  # How many trajectories to sample at once
+
+
+async def main():
```
pbontrager, line 28: I'm mostly going to leave this section alone until we've got things working.

forge/rl/interfaces.py
```diff
@@ -0,0 +1,128 @@
...
+
+from forge.rl.environments.base import Action, Observation, State
+
+
+# ==== Data Structures ====
+# Of course, these data structures can be replaced by almost anything.
+
+
+@dataclass
+class Trajectory:
```
pbontrager, line 32: I think we should separate our types (envs, rewards, policy, etc) from our data types. Interfaces should be where users go to be able to see how to write their custom modules.

forge/monarch_utils/stack.py
allenwang28: Yes, this likely ends up being exposed as `monarch.actor.stack` instead of something kept in Forge. But I got feedback to keep it here in Forge so we can iterate on it quicker until we're confident in what its functionality should be. The intent of this is to just feel like a fused actor mesh

examples/rl_main.py
allenwang28: yeah good call, I removed the config altogether

examples/rl_main.py
allenwang28: I added data, which hosts environments and policies.

forge/monarch_utils/stack.py
pbontrager: My comment is less related to whether it's in forge or monarch and more about what public objects we're exposing to users. `stack` the function is simple, but what I'm concerned about is having 4 different mesh types. It would be simpler if `stack(List[ActorMesh]) -> ActorMesh`. From the Monarch side, I think it's a bit confusing having separate process and actor meshes, I'm not sure how much extra utility you get from the distinction.

This is not something I'm blocking on here. More of a note for something I think we should fix or improve on going forward.

forge/rl/environments/base.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
pbontrager, line 1: This file can be moved into inferfaces (data types I'd still prefer to be separated but I'll leave that) and the environments folder can be deleted.

forge/monarch_utils/stack.py
allenwang28: yes, I agree. My point is more that you interact with a `StackedActorMesh` the exact same way as you interact with an `ActorMesh` and we could hide it more elegantly when it's upstreamed into Monarch

forge/rl/policy.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
pbontrager, line 1: I misunderstood before that this wasn't the user defined policy (sampling logic, token insertion rules, etc) and is the generator itself. In general I think there's a distinction between our actors and the user defined logic they run. With current naming, we're building something like below:

```
Actor Class -> User Class
----------------------------
Collector -> Environments
Policy -> Model + Policy
Rewards -> Rewards
Trainer -> Model + Optimizer + Loss
```

It might be clearer to have more standardized naming and reuse the same structure for lots of these. Something like

```
EnvRunner -> Environment
PolicyRunner -> Model + Policy
RewardRunner -> Reward
Trainer -> Model + Optimizer + Loss
```

forge/data/environments/toy.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@dataclass
+class ToyAction(Action):
+    """Action for the toy environment."""
+
+    data: torch.Tensor
+
+
+class ToyRewarder(Actor):
```
pbontrager: We should have a `Scorer` Actor class that takes a Reward type class and runs it. But that can be left for a future PR.

forge/data/policies/toy.py
```diff
@@ -0,0 +1,40 @@
...
+
+import torch
+from monarch.actor import endpoint
+
+from forge.data.environments.toy import ToyAction
+from forge.rl.environments import Action, Observation
+from forge.rl.interfaces import PolicyInterface
+
+
+class ToyPolicy(PolicyInterface):
```
pbontrager, line 17: I realize now this is meant to be a stand-in for the vllm server not the user defined policy. Sorry for the misunderstanding, this would go under rl then.

forge/monarch_utils/stack.py
Darktex: I agree with this. Basically n-dimensional meshes should already have stack behavior

forge/data/environments/toy.py
allenwang28: ack and agree, I removed Rewarder altogether here for now.

forge/rl/policy.py
allenwang28: We discussed live - agreed that we want a consistent name to mark what is an actor and what isn't, follow up in another PR

examples/rl_main.py
allenwang28: IIUC, while asyncio runs concurrently, it doesn't actually run in parallel. But you're right, as is, after a worker is done it doesn't add the next one, and backpressure may be a concern.

I think we can come to a satisfying solution, would prefer to do this in a follow up

Darktex:
Let's land this one and start making changes on top. It doesn't make sense to block the earliest PRs in a project :)

forge/rl/replay_buffer.py
```diff
@@ -0,0 +1,56 @@
...
+
+import random
+
+from monarch.actor import endpoint
+
+from forge.rl.interfaces import ReplayBufferInterface, Trajectory
+
+
+# Silly replay buffer implementation for testing.
+# One nice thing if we implement our own Replay buffer is that
```
ebsmothers, line 17: Re implementing our own replay buffer, what's the alternative? I had kinda assumed we were doing this anyways

forge/rl/environments/base.py
```diff
@@ -1,23 +1,29 @@
...
 from __future__ import annotations
 
 import abc
+
 from dataclasses import dataclass, field
 from typing import Any, Callable, Dict, Optional, Union
 
 
-@dataclass
+@dataclass(kw_only=True)
```
ebsmothers, line 19: i like this

forge/rl/interfaces.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __post_init__(self):
+        if self.states is None:
+            self.states = []
+        if self.actions is None:
+            self.actions = []
+        if self.dones is None:
+            self.dones = []
+        if self.infos is None:
+            self.infos = []
```
ebsmothers: Can you use `field(default_factory=list)` to get around all these?

forge/rl/interfaces.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+@dataclass
+class Trajectory:
+    """A trajectory containing a sequence of states, actions, etc."""
+
+    states: list[Observation] | None = None
+    actions: list[Action] | None = None
+    dones: list[bool] | None = None
+    infos: list[dict] | None = None
```
ebsmothers: Noob q: why are these the right set of fields for a trajectory? (Also is `infos` just some generic metadata field?)

forge/rl/collector.py
```diff
@@ -0,0 +1,65 @@
...
+        policy: PolicyInterface,
+        replay_buffer: ReplayBufferInterface,
+        environment_creator: Callable,
+    ):
+        self.max_collector_steps = max_collector_steps
+        self.replay_buffer = replay_buffer
+        self.environment_creator = environment_creator
+        # maybe this is just the policy endpoint with a router?
+        self.policy = policy
+        self.environment = self.environment_creator()
```
ebsmothers, line 35: Should I think of this as environment.reset()?

examples/rl_main.py
```diff
@@ -0,0 +1,201 @@
...
+    print("Starting RL example with toy environment...")
+
+    # Process allocation
+    policy_procs = await proc_mesh(gpus=2)
+    replay_procs = await proc_mesh(gpus=1)
+
+    # Note - here is where we implement our "mixture" logic.
+    browser_procs = await proc_mesh(gpus=2)
+    deep_research_procs = await proc_mesh(gpus=4)
+    coder_procs = await proc_mesh(gpus=8)
```
ebsmothers, line 38: Sorry I am gonna ask some noob questions here.. from the notebook I thought we were actually keeping a CPU pool  and not really being opinionated about the resource allocation across different environment types. But seems like that's not happening here?

examples/rl_main.py
```diff
@@ -0,0 +1,201 @@
...
+        while True:
+            try:
+                print(f"🎮 Running episode {episode_count + 1}...")
+
+                # call() is essentially our "map" - every collector runs their own
+                # episode loop.
+                # What's pretty elegant here is if we wanted to control off policiness, we could
+                # easily counter on steps and call policy.update_weights.call() at our desired
+                # frequency.
+                results = await collectors.run_episode.call()
```
ebsmothers, line 97: General Monarch question: where does a noob like myself go to understand the contracts of call, choose, etc? (Obviously there's the code itself but wondering if there's anywhere else you'd recommend)

forge/rl/replay_buffer.py
allenwang28: In Cabernet times, we were using torchrl's Replay buffer which is what motivated this comment. I guess you're right in that there isn't really an alternative at this point lol

forge/rl/interfaces.py
allenwang28: Actually good catch, these were a remnant of assuming we were going to follow the `gymnasium` API exactly. IIUC the `infos` in the `gymnasium` API is like a catch all for additional data that might not be well captured in state/action/rewards

forge/rl/interfaces.py
allenwang28: Yep! Good catch, fixed

forge/rl/collector.py
allenwang28: ah good question, as architected right now each "collector" is associated with one "environment", sharing the same process. The `environment_creator()` is a generic function that instantiates the environment

examples/rl_main.py
allenwang28: We should definitely change the APIs in Monarch, but the `gpus` here doesn't actually mean anything haha. All `await proc_mesh(gpus=8)` does is creates 8 processes on CPUs.

examples/rl_main.py
allenwang28: Currently code is best, but https://docs.google.com/document/d/12JGL6TIdPzc9jVu_gO856bUCZYoC0qPA1lSqmoa2gQ4 outlines the motivation

PR #7, opened 2025-07-08
Iterable dataset + datamix + pack on the fly + data metrics

felipemello1:
image

What this PR does:
TLDR: You can do PackedDataset(InterleavedDataset([ds1, InterleavedDataset([d2, ds3])) and generate metrics for each dataset.

[x] Iterable infinite datasets
- Currently relies on HF, but can be swapped with any other library, e.g. torchdata/custom code. (PS: Everything downstream doesn't need HF)

[x] Data mix
- You can define sampling weights for multiple datasets
- You can nest InterleavedDatasets, e.g. For wikipedia, have 30% reasoning, 50% image, 20% high quality, but wikipedia is only 10% of all of your datasets.

[x] Packing on the fly
- greedy sampling until len(sample) > space left. Then searches the buffer for samples that fit. In a simple test with alpaca and orca, <1% of the tokens were padding.

[x] Data Metrics
- Easily add metrics to your datasets, e.g. num tokens seen, distribution of sequence lens, categorical count of sample types, etc.

TODOs:
[ ] Run recipe in main
[ ] Run recipe with this PR ([reference](https://github.com/pytorch/torchtune/pull/2819/files#diff-a5cbecce5528877f709e4e6152862af874d7699e8a832ab41d74f010ff23646f))
[ ] Adjust/delete other functions/tests that don't expect iterable

Follow ups not addressed in this PR:
[ ] Unit tests with dataloader with num_workers>0
[ ] we only support infinite datasets, but that doesnt work for validation, unless we have a way to stop after exactly one epoch
[ ] Unit test SFTOutputTransform
[ ] Resume from ckpt for PackedDataset (we drop buffer, so resuming is not exact)
[ ] collate_fn requires "device" arg so we can pass it to the mask_mod function that generates masks on the fly. There may be a better way to do this, e.g. pass it to PackingDataset -> Packer
[ ] Enable other packing strategy (binned packing) if padding levels are still high (i dont think that they should be)
[ ] User cannot do "finite" datasets. They cannot choose to iterate for exactly 1 epoch and stop it. Datasets are infinite because its simple -> they will never hang
[ ] Log at the start of training. E.g. print dataset.info
[ ] Packed cannot have exact resumption from checkpoint, because buffer_size needs to be >=1, and when resuming, buffer is discarded. Alternatives are to: a) checkpoint the buffer / intermediate packs; b) make it work with buffer=0.

### DIFF
--- a/forge/datasets/packed.py
+++ b/forge/datasets/packed.py
@@ -23,8 +23,11 @@ from torchdata.stateful_dataloader import Stateful
 logger = logging.getLogger(__name__)
 
 
+# Generic type so custom specific ones can be defined for new Packers
 SampleType = TypeVar("SampleType")
-PackType = dict[str, torch.Tensor | list[Metric]]
+
+# All purpose type for samples that can contain tensors, metrics, or any other data
+SampleDict = dict[str, Any]
 
 
 class Packer(ABC, Generic[SampleType]):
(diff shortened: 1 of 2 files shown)


scripts/download_tokenizer.py
joecummings, line 1: What's this doing here?

felipemello1: was testing things, this was not cloned from titan, so i just copied it over

PR #8, opened 2025-07-10
[1/3] Add ChatEnvironment

Darktex:
Creates the basic ChatEnv with batteries included. 

NOTE: This diff comes with the proposal to marry the HF way of handling messages. Note that we don't have to take a strict dependency on them (though of course we expect that most customers will happily use HF), so we implement a simple protocol that follows their own [way of doing templates](https://huggingface.co/docs/transformers/en/chat_templating).

I'm also not sure if we should just go with dictionaries all the way or not. For now, I put dataclasses...

### DIFF
--- a/forge/protocols/tokenizer.py
+++ b/forge/protocols/tokenizer.py
@@ -1,0 +1,88 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, Dict, List, Optional, Protocol, Union
+
+import torch
+
+
+class TokenizerProtocol(Protocol):
+    """Protocol for tokenizers used throughout Forge.
+
+    This protocol defines the interface for tokenizers that support Hugging Face's
+    apply_chat_template method with full signature support. This enables proper
+    handling of turn boundaries, generation prompts, and message continuation.
+
+    Key parameters for chat applications:
+    - add_generation_prompt: Adds tokens that indicate the start of a response
+    - continue_final_message: Removes EOS tokens to continue the final message
+    - return_assistant_tokens_mask: Returns a mask for assistant-generated tokens
+    """
+
+    def apply_chat_template(
+        self,
+        conversation: List[Dict[str, str]],
+        tools: Optional[List[Dict]] = None,
+        documents: Optional[List[Dict[str, str]]] = None,
+        chat_template: Optional[str] = None,
+        add_generation_prompt: bool = False,
+        continue_final_message: bool = False,
+        tokenize: bool = True,
+        padding: bool = False,
+        truncation: bool = False,
+        max_length: Optional[int] = None,
+        return_tensors: Optional[str] = None,
+        return_dict: bool = False,
+        return_assistant_tokens_mask: bool = False,
+        tokenizer_kwargs: Optional[Dict[str, Any]] = None,
+        **kwargs
+    ) -> Any:
+        """Apply chat template to messages with full Hugging Face signature.
+
+        Args:
+            conversation: A list of dicts with "role" and "content" keys, representing
+                the chat history so far.
+            tools: A list of tools (callable functions) that will be accessible to the model.
+            documents: A list of dicts representing documents for RAG applications.
+            chat_template: A Jinja template to use for this conversion.
+            add_generation_prompt: If True, appends tokens that indicate the start of
+                an assistant message. Useful for generating responses.
+            continue_final_message: If True, formats the chat so the final message is
+                open-ended without EOS tokens. Cannot be used with add_generation_prompt.
+            tokenize: Whether to tokenize the output. If False, returns a string.
+            padding: Strategy to pad returned sequences.
+            truncation: Whether to truncate sequences at maximum length.
+            max_length: Maximum length for padding or truncation.
+            return_tensors: Framework for returned tensors ('pt', 'tf', 'np', 'jax').
+            return_dict: Whether to return a dictionary with named outputs.
+            return_assistant_tokens_mask: Whether to return a mask of assistant tokens.
+            tokenizer_kwargs: Additional kwargs to pass to the tokenizer.
+            **kwargs: Additional kwargs to pass to the template renderer.
+
+        Returns:
+            A list of token ids or a dict of tokenizer outputs, depending on parameters.
+        """
+        ...
+
+    def decode(
+        self,
+        token_ids: Union[int, List[int], torch.Tensor],
+        skip_special_tokens: bool = False,
+        clean_up_tokenization_spaces: Optional[bool] = None,
+        **kwargs
+    ) -> str:
+        """Decode token ids to a string.
+
+        Args:
+            token_ids: List of tokenized input ids. Can be obtained using the __call__ method.
+            skip_special_tokens: Whether or not to remove special tokens in the decoding.
+            clean_up_tokenization_spaces: Whether or not to clean up the tokenization spaces.
+                If None, will default to self.clean_up_tokenization_spaces.
+            kwargs: Additional keyword arguments to pass to the underlying model specific decode method.
+
+        Returns:
+            The decoded string.
+        """
+        ...
--- a/forge/rl/environments/base.py
+++ b/forge/rl/environments/base.py
@@ -1,0 +1,125 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import annotations
+
+import abc
+from dataclasses import dataclass, field
+from typing import Any, Callable, Dict, Optional, Union
+
+
+@dataclass
+class Observation:
+    """Base class for environment observations.
+
+    Contract:
+    - Should contain all information needed by an agent to make decisions
+    - Should be serializable/deserializable
+    - Should be immutable (or treated as such)
+
+    Args:
+        done: Whether the episode/conversation is complete
+        reward: Optional reward signal (can be boolean, int, or float)
+        metadata: Additional data that doesn't affect agent decisions but may be useful
+                 for transforms, logging, evaluation, etc.
+    """
+
+    done: bool = False
+    reward: Optional[Union[bool, int, float]] = None
+    metadata: Dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class Action:
+    """Base class for environment actions.
+
+    Contract:
+    - Should contain all information needed to execute a step in the environment
+    - Should be serializable/deserializable
+    - Should be immutable (or treated as such)
+
+    Args:
+        metadata: Additional data that may be useful for logging, debugging, or transforms
+    """
+
+    metadata: Dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class State:
+    """Base class for environment state.
+
+    Contract:
+    - Should contain all information needed to restore the environment
+    - Should be serializable/deserializable
+    - May contain information not exposed in observations
+
+    Args:
+        metadata: Additional state information that may be useful for debugging or analysis
+    """
+
+    metadata: Dict[str, Any] = field(default_factory=dict)
+
+
+class Transform(abc.ABC):
+    """Abstract base class for observation transforms.
+
+    Transforms are first-class citizens that can modify observations,
+    typically to add rewards, compute metrics, or modify state.
+
+    They follow a functional interface where they take an observation
+    and return a (potentially modified) observation.
+    """
+
+    @abc.abstractmethod
+    def __call__(self, observation: Observation) -> Observation:
+        """Transform an observation.
+
+        Args:
+            observation: The input observation to transform
+
+        Returns:
+            The transformed observation (may be the same instance if no changes)
+        """
+        pass
+
+
+class Environment(abc.ABC):
+    """Abstract base class for environments.
+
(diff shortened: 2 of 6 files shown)


forge/rl/environments/chat.py
```diff
@@ -0,0 +1,222 @@
...
+from forge.rl.environments.base import Action, Environment, Observation, State
+
+
+@dataclass
+class ChatAction(Action):
+    """Action for ChatEnvironment.
+
+    Can contain either:
+    1. A message with role and content (strings) - will be tokenized by the environment
+    2. Pre-tokenized tokens - assumes chat template has already been applied
```
ebsmothers: It's good to be flexible, but I wonder if this class is trying to do a bit too much? Personally I would prefer to have a single clear contract for input format per action. But lmk if there are some obvious use cases I'm missing that necessitate this

forge/rl/environments/chat.py
```diff
@@ -0,0 +1,222 @@
...
+        # Type assertion is safe here because validation ensures both are not None
+        assert self.role is not None and self.content is not None
+        return {"role": self.role, "content": self.content}
+
+
+@dataclass
+class ChatState(State):
+    """State of the ChatEnvironment containing message history and tokenized messages."""
+
+    messages: List[Dict[str, str]]
```
ebsmothers: noob q: is the dict just a placeholder until we have the Message abstraction? Also super nit but since we're mostly starting from scratch here can we set the precedent to use builtin types instead of the typing ones?

forge/rl/environments/chat.py
```diff
@@ -0,0 +1,222 @@
...
+    """Observation returned by ChatEnvironment.
+
+    Contains the message history in Huggingface format (list of dicts with role/content)
+    and the tokenized representation of each message.
+
+    Example:
+    messages = [
+     {"role": "system", "content": "You are a helpful assistant"},
+     {"role": "user", "content": "How tall is the Eiffel Tower?"},
+    ]
```
ebsmothers, line 59: Missing tokens in the example? It's also not clear to me how this differs from `ChatState`

forge/protocols/tokenizer.py
```diff
@@ -0,0 +1,65 @@
...
+    apply_chat_template method with full signature support. This enables proper
+    handling of turn boundaries, generation prompts, and message continuation.
+
+    Key parameters for chat applications:
+    - add_generation_prompt: Adds tokens that indicate the start of a response
+    - continue_final_message: Removes EOS tokens to continue the final message
+    - return_assistant_tokens_mask: Returns a mask for assistant-generated tokens
+    """
+
+    def apply_chat_template(
```
ebsmothers, line 24: This seems like a reasonable approach to me and certainly avoids some of the challenges we had in tune of having to bake in chat templates after the fact. I want to understand whether we are making any implicit stance here through the use of this protocol. Iiuc the point is "you can use HF, you can also bring your own tokenizer, but you need to define this method". Is that right? And do we anticipate being in the business of providing our own tokenizers in the repo? Personally this is something I wish we had outsourced in tune from the get-go

forge/rl/environments/base.py
```diff
@@ -0,0 +1,68 @@
...
+    Contract:
+    - Should contain all information needed to restore the environment
+    - Should be serializable/deserializable
+    - May contain information not exposed in observations
+    """
+
+    pass
+
+
+class Environment(abc.ABC):
```
ebsmothers, line 88: Curious why ABC here vs Protocol for the tokenizer. Do we prefer ABC in general? (I understand Protocol is necessary for tokenizer because of the need for HF integration)

tests/unit_tests/rl/environments/test_chat.py
```diff
@@ -0,0 +1,590 @@
...
+            conversation=[{"role": "assistant", "content": "Hello"}],
+            continue_final_message=False,
+            tokenize=False,
+        )
+
+        # With continue_final_message, the end token should be removed
+        self.assertNotIn("<|end|>", result_continue)
+        self.assertIn("<|end|>", result_normal)
+
+    def test_generation_prompt_and_continue_message_conflict(self):
```
ebsmothers: That's a lot of test cases 👀

forge/rl/environments/chat.py
```diff
@@ -0,0 +1,222 @@
...
+        return ChatObservation(messages=self._state.messages, tokens=self._state.tokens)
+
+    def _tokenize_messages(self):
+        """Tokenize all messages individually and store as a list of tensors."""
+        self._state.tokens = []
+        for message in self._state.messages:
+            message_tokens = self.tokenizer.apply_chat_template(
+                conversation=[message], tokenize=True, return_tensors="pt"
+            )
+            self._state.tokens.append(message_tokens)
```
ebsmothers: I thought `apply_chat_template` was already intended to be used over a list of messages. In that case why do we need to do in a loop? Is it because here we are just tokenizing the messages and not actually applying the chat template? Personally I find that a bit confusing (maybe it's just me though)

forge/rl/environments/base.py
Darktex: This is where maybe I haven't caught up with the latest on how you are supposed to do this in Python, as this honestly keeps changing with the new versions. I'm still used to ABC so that's what I used (except when forced to use Protocol).

Should we switch to Protocol? No strong opinions tbh. I was also thinking that we could use TypedDict here which is basically a dataclass that you can use as a Python dict as well for extra flexibility

forge/rl/environments/chat.py
```diff
@@ -0,0 +1,222 @@
...
+     {"role": "system", "content": "You are a helpful assistant"},
+     {"role": "user", "content": "How tall is the Eiffel Tower?"},
+    ]
+    """
+
+    messages: List[Dict[str, str]]
+    tokens: List[torch.Tensor]
+
+
+class ChatEnvironment(Environment):
```
pbontrager, line 68: Can we have this inherit from https://gymnasium.farama.org/api/env/? This allows us to plug directly into the full ecosystem

forge/rl/environments/base.py
pbontrager: I think we should fully use ABC where we can in this library. This provides clear documentation for the user on how to extend things.

forge/rl/environments/base.py
```diff
@@ -0,0 +1,68 @@
...
+# LICENSE file in the root directory of this source tree.
+from __future__ import annotations
+
+import abc
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass
+class Observation:
```
pbontrager, line 13: I think we could use just "messages" for both the observation and actions and sync the design with the sft side since it makes the data code more uniform. Also it wouldn't be specific to just chat. I think you could do something like

```python
Role = Literal[
    "system",  # Origin is system prompt
    "user",  # Origin is user
    "assistant",  # Origin is the model output
    "agent", # Origin is generated
    "tool",  # Origin is return from a tool call
]


@dataclass
class Message:
	"""Generalization of existing torchtune message"""
	role: Role
	content: Tensor
	logprobs: Dict[str, Tensor]
	reward: Dict[str, Tensor]
	policy_version: int


@dataclass
class Rollout:
	"""Simplified  example rollout dataclass"""
	messages: List[Message]
	group_id: int

	@property
	def tokens(self, roles):
		toks = [m.content for m in self.messages if m.role in roles]
		return sum(toks, [])

	@property
	def logprobs(self, key, roles):
		probs = [m.logprobs[key] for m in self.messages if m.role in roles]
		return torch.cat(probs, 0)

	@property
	def rewards(self, key, roles):
		rewards = [m.reward[key] for m in self.messages if m.role in roles]
		return torch.cat(rewards, 0)
```

forge/protocols/tokenizer.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, Dict, List, Optional, Protocol
+
+
+class TokenizerProtocol(Protocol):
```
pbontrager, line 11: I think you want a `tokenize_messages` method too as well as probably `decode`

ebsmothers: This is kinda what I was alluding to with my other comment.. if we do that we are then starting to diverge from generic HF tokenizers, right? (Specifically talking about `tokenize_messages` here)

forge/protocols/tokenizer.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
pbontrager, line 1: HIgh level comment, do we want two parallel trees of folders for sft and rl or have everything mixed at the top level? I could imagine environments going under data for example. Don't really have a preference, both have downsides. @ebsmothers

forge/rl/environments/chat.py
Darktex: You are possibly right, it's one of the areas where I wanted to get some feedback. We should have a solid foundation for the rest, including keeping non-tokenized data around for ease of debugging...

What do people think?

forge/rl/environments/chat.py
Darktex: I blame Devmate for this typing AI slop :P 

The dict *is* the HF message abstraction though

forge/rl/environments/chat.py
Darktex: Ah, that is a good idea. There's a few more methods to implement but it seems worth it to me! Maybe this is a good first task for @DNXie to hop on?

forge/protocols/tokenizer.py
ebsmothers: Personally I kinda liked the amaia model: a core library with sft and rl as apps (props to @Darktex for converting me to it). That being said, if we then wind up with 90% of code going into apps/rl it'd make the library feel a bit imbalanced (and if we consider environments as part of our core offering we probably shouldn't relegate them to out-of-core). Not sure I answered your question but oh well 😃

forge/protocols/tokenizer.py
joecummings, line 1: Bit of a nit, but if we have "base" in other directories that enforce a certain contract, then we probably want to have a transforms or tokenizer directory that has a base.py that enforces some sort of tokenizer protocol.

forge/rl/environments/base.py
```diff
@@ -0,0 +1,68 @@
...
+class Environment(abc.ABC):
+    """Abstract base class for environments."""
+
+    @abc.abstractmethod
+    def reset(self) -> Observation:
+        """Reset the environment and return an initial observation."""
+        pass
+
+    @abc.abstractmethod
+    def step(self, action: Any) -> Observation:
```
allenwang28, line 111: Shouldn't the env return a reward here too?

Darktex: No, step() always returns an observation, and a reward can be part of that observation if/when there is one! Note that we have no rewards in this diff yet, look at the GSM8k PR

Darktex:
Answered comments

forge/protocols/tokenizer.py
Darktex: Honestly, imho adding both ABC and Protocols to Python has made things even more confusing :( here it needs to be protocol though because the HF code like their tokenizer doesn't depend on our ABCs. Otherwise, I'd very much prefer ABCs

forge/protocols/tokenizer.py
Darktex: image

forge/protocols/tokenizer.py
Darktex: Yes, this is exactly the spirit. They solved a bunch of annoying issues, with things like `add_generation_prompt` that actually helps a lot in RL. That said, no reason for us to have a hard dependency: sometimes people want to use TikToken directly, in which case they would be able to do so by writing a few methods -- but I suspect that the vast majority of our customers will use HF and be happy with it

forge/rl/environments/base.py
Darktex: Yeah agreed

forge/rl/environments/chat.py
Darktex: ... You are right, will fix it

forge/rl/environments/chat.py
Darktex: Will fix the docstring.

In this case, there is no difference: the whole state is observable!

forge/rl/environments/base.py
Darktex: I'd rather not have a fixed enum for roles because different models will require different roles so it's hard to be prescpritive -- plus, it's not like we gain a lot in doing so.

For Messages, I'd stick to the simple HF class. We will also need to have all those extra things like rewards and logprobs, but they can belong to the Observation class, right?

forge/protocols/tokenizer.py
Darktex: Not necessarily diverge: we just need to identify what exactly we need from them and put that in the Protocol, but otherwise if they expose extra methods we may not care about them.

Darktex:
Implemented suggestions, PTAL!

forge/protocols/tokenizer.py
Darktex: So, here we are in a `Protocols/` directory together with training protocols and whatnot, so they are all base in a way? How would you structure this?

forge/rl/environments/base.py
joecummings: > different models will require different roles

The roles I've seen it are pretty set independent of model. I haven't seen any for a message outside of "system",  "user",  "assistant", or "tool".

PR #9, opened 2025-07-13
[3/3] GSM8K dataset (with DatasetChatEnv)

Darktex:
Not sure about the ColumnMapping part -- this came from DevMate lol. It does seem quite decent, but I'd like to get feedback on it.

### DIFF
--- a/forge/rl/environments/gsm8k.py
+++ b/forge/rl/environments/gsm8k.py
@@ -1,0 +1,203 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import re
+from typing import Optional
+
+from forge.rl.environments.base import Observation, Transform
+from forge.rl.environments.chat import ChatObservation
+
+from forge.rl.environments.dataset import ColumnMapping, DatasetChatEnvironment
+
+
+def extract_gsm8k_answer(text: str, method: str = "flexible") -> Optional[float]:
+    """Extract the final numerical answer from GSM8K text.
+
+    For GSM8K, answers are typically marked with #### followed by the number.
+    In strict mode, only the #### format is accepted.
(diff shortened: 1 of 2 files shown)


forge/rl/environments/gsm8k.py
```diff
@@ -0,0 +1,203 @@
...
+            reward = 0.0
+
+        # Create new observation with reward
+        observation.reward = reward
+        observation.done = True  # GSM8K is typically single-turn
+
+        return observation
+
+
+class GSM8KEnvironment(DatasetChatEnvironment):
```
ebsmothers, line 133: Given that I was just pondering about deleting our Alpaca dataset builder I am gonna give you a hard time about the same thing 😃. In my mind they're analogous: GSM8KEnvironment : DatasetChatEnvironment :: alpaca_iterable_dataset : sft_iterable_dataset. Do you agree? If so, I think we should treat both types of "convenience implementations" similarly

forge/rl/environments/gsm8k.py
```diff
@@ -0,0 +1,203 @@
...
+            return observation
+
+        # Get the expected answer from metadata
+        expected_answer = None
+        model_response = observation.messages[-1]["content"]
+
+        # Look for expected answer in metadata
+        if hasattr(observation, "metadata") and observation.metadata:
+            # Try common metadata keys for the expected answer
+            for key in ["answer", "expected_answer", "solution"]:
```
ebsmothers, line 107: This feels kinda arbitrary (maybe I just need to get used to reward modeling code). But wonder if we can clamp down here in any way so we're not just constantly searching for random metadata keys

forge/rl/environments/gsm8k.py
Darktex: Maybe. The issue is whether how you construct it from config. I think that the Alpaca dataset can be entirely configured, but I don't think we can do it for envs because of the reward functions... But open to ideas

PR #10, opened 2025-07-14
[wip] Trainer Actor

allenwang28:
This PR creates a simple Monarch actor wrapper for the training loop. This is loosely based off of parallel work taking place in Monarch-land to guide what will need to be accomplished. 

Plan is not necessarily to check this in - I think this will be most useful as a reference starting point - we may want to e.g. pull functionality out of the titan trainer itself longer term.


examples/train_actor.py
```diff
@@ -0,0 +1,139 @@
...
+import os
+
+import time
+from typing import Optional
+
+from forge.train import ConfigManager, init_logger, Trainer
+from monarch.actor import Actor, current_rank, current_size, endpoint, proc_mesh
+
+
+class TrainerActor(Actor):
```
pbontrager, line 27: This isn't really what the train actor is. I think we need to wait for the SFT version to land first, but it should inherit from the Titan trainer and we'll need a custom train step/loop, dataset setup, loss setup, and weight sync logic.

allenwang28: yeah I agree, things will be clearer once there is an SFT version

I guess this PR is to show that once the SFT version is landed, converting to an actor isn't a heavy lift

Darktex: Yep, pretty much! Should just thinly wrap titan and we should be good with minimal changes compared to SFT (mostly changing loss, reading from replay buffer instead of disk etc)

PR #11, opened 2025-07-15
[2/3] Datasetchatenv

Darktex:
This turned out to be far less trivial than I had anticipated. What do people think of this API?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -185,5 +185,7 @@ cover/
 # VSCode
 .vscode/
 
+PREFERENCES.md
+
 # wandb
 wandb/
--- a/forge/rl/environments/dataset.py
+++ b/forge/rl/environments/dataset.py
@@ -1,0 +1,371 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import random
+from dataclasses import dataclass, field
+from typing import Any, Callable, Dict, List, Optional
+
+import torch
+from datasets import Dataset, load_dataset
+
+from forge.protocols.tokenizer import TokenizerProtocol
+from forge.rl.environments.base import Transform
+from forge.rl.environments.chat import (
+    ChatAction,
+    ChatEnvironment,
+    ChatObservation,
+    ChatState,
+    Message,
+)
+from torch.utils.data import DataLoader
+
+
+@dataclass
+class DatasetChatState(ChatState):
+    """State for DatasetChatEnvironment that includes metadata.
+
+    Extends ChatState to include metadata from the dataset.
+    """
+
+    metadata: Dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class ColumnMapping:
+    """Configuration for mapping dataset columns to chat roles or metadata.
+
+    This class defines how to extract content from dataset columns and either:
+    1. Assign to chat roles (for conversation messages)
+    2. Store as metadata (for reference data like expected answers)
+    """
+
+    column: str
+    role: Optional[str] = None  # None means store as metadata
+    content_template: Optional[str] = None
+    metadata_key: Optional[str] = None  # Key to store in metadata dict
+
+    def __post_init__(self):
+        """Validate that either role or metadata_key is provided."""
+        if self.role is None and self.metadata_key is None:
+            raise ValueError("Either 'role' or 'metadata_key' must be provided")
+        if self.role is not None and self.metadata_key is not None:
+            raise ValueError("Cannot specify both 'role' and 'metadata_key'")
+
+        # If metadata_key not specified but role is None, use column name
+        if self.role is None and self.metadata_key is None:
+            self.metadata_key = self.column
+
+    @property
+    def is_metadata(self) -> bool:
+        """Check if this mapping is for metadata storage."""
+        return self.role is None
+
+    def extract_content(self, row: Dict[str, Any]) -> str:
+        """Extract content from a dataset row.
+
+        Args:
+            row: A single row from the dataset
+
+        Returns:
+            The extracted content as a string
+        """
+        content = row[self.column]
+        if self.content_template:
+            # Allow template formatting with row data
+            return self.content_template.format(**row)
+        return str(content)
+
+
+class DatasetChatEnvironment(ChatEnvironment):
+    """A ChatEnvironment that loads prompts from HuggingFace datasets.
+
+    This environment follows the standard RL environment API:
+    - reset() loads a new episode (conversation) from the dataset
+    - step() adds a new turn to the current conversation
+
+    Most datasets will be single-turn (reset() loads a prompt, done), but this
+    design supports multi-turn conversations like TauBench where step() can
+    extend the conversation.
+
+    Example usage for single-turn datasets:
+        env = DatasetChatEnvironment(
+            tokenizer=tokenizer,
+            dataset_name="gsm8k",
+            column_mappings=[
+                ColumnMapping(column="question", role="user")
+            ]
+        )
+
+        obs = env.reset()  # Loads new question from dataset
+        # obs.messages = [{"role": "user", "content": "What is 2+2?"}]
+
+        # Agent generates response
+        action = ChatAction(role="assistant", content="4")
+        obs = env.step(action)  # Adds assistant response
+
+    Example usage for multi-turn datasets:
+        env = DatasetChatEnvironment(
+            tokenizer=tokenizer,
+            dataset_name="taubench",
+            column_mappings=[
+                ColumnMapping(column="user_message", role="user"),
+                ColumnMapping(column="assistant_message", role="assistant")
+            ]
+        )
+
+        obs = env.reset()  # Loads initial conversation from dataset
+        # Continue conversation with additional turns via step()
+    """
+
+    def __init__(
+        self,
+        tokenizer: TokenizerProtocol,
+        dataset_name: Optional[str] = None,
+        dataset: Optional[Dataset] = None,
+        column_mappings: Optional[List[ColumnMapping]] = None,
+        system_prompt: Optional[str] = None,
+        system_role: str = "system",
+        split: str = "train",
+        shuffle: bool = True,
+        seed: Optional[int] = None,
+        filter_fn: Optional[Callable[[Dict[str, Any]], bool]] = None,
+        transform: Optional[Transform] = None,
+        **dataset_kwargs,
+    ):
+        """Initialize DatasetChatEnvironment.
+
+        Args:
+            tokenizer: Tokenizer for processing messages
+            dataset_name: Name of HuggingFace dataset to load (e.g., "gsm8k")
+            dataset: Pre-loaded dataset (alternative to dataset_name)
+            column_mappings: List of ColumnMapping objects defining role assignments
+            system_prompt: Optional system prompt
+            system_role: Role for system messages
+            split: Dataset split to use ("train", "test", "validation")
+            shuffle: Whether to shuffle the dataset
+            seed: Random seed for reproducibility
+            filter_fn: Optional function to filter dataset rows
+            transform: Optional transform function to apply to observations
+            **dataset_kwargs: Additional arguments for load_dataset
+        """
+        # Initialize with a custom state class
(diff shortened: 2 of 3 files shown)


forge/rl/environments/dataset.py
```diff
@@ -0,0 +1,371 @@
...
+            The extracted content as a string
+        """
+        content = row[self.column]
+        if self.content_template:
+            # Allow template formatting with row data
+            return self.content_template.format(**row)
+        return str(content)
+
+
+class DatasetChatEnvironment(ChatEnvironment):
```
pbontrager, line 81: I don't understand why there needs to be a separate DatasetChatEnv and ChatEnv. Don't we just want a single chat env that wraps a dataloader and we can add tool support down the line?

forge/rl/environments/base.py
```diff
@@ -0,0 +1,197 @@
...
+# LICENSE file in the root directory of this source tree.
+from __future__ import annotations
+
+import abc
+from dataclasses import dataclass, field
+from typing import Any, Callable, Dict, Optional, Union
+
+
+@dataclass
+class Observation:
```
pbontrager: If this is a standin for the Message type (I think @joecummings  proposed OAI message with extra metadata?) then I'm fine with this, but I think we should put in the comments a todo or warning so we remember to remove these.

forge/rl/environments/base.py
```diff
@@ -0,0 +1,197 @@
...
+                 for transforms, logging, evaluation, etc.
+    """
+
+    done: bool = False
+    reward: Optional[Union[bool, int, float]] = None
+    metadata: Dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class Action:
```
pbontrager: This should be able to be the same message type, not separate from the observation

forge/rl/environments/base.py
```diff
@@ -0,0 +1,197 @@
...
+    - May contain information not exposed in observations
+
+    Args:
+        metadata: Additional state information that may be useful for debugging or analysis
+    """
+
+    metadata: Dict[str, Any] = field(default_factory=dict)
+
+
+class Transform(abc.ABC):
```
pbontrager: This is the tokenizer right? In torchtune we had tokenizers as a special case of transforms but that ended up causing a lot more confusion than what it was worth. I think we should just use the tokenizer API and assume they can be multimodal. Unless this is transforming the raw output to the OAI format? I think for the latter case, we should just require that the env writer handles that themselves. For datasets we already handle this for you

forge/rl/environments/base.py
```diff
@@ -0,0 +1,197 @@
...
+        """Take a step in the environment and return an observation."""
+        pass
+
+    @property
+    @abc.abstractmethod
+    def state(self) -> State:
+        """Get the current state of the environment."""
+        pass
+
+    def _apply_transform(self, observation: Observation) -> Observation:
```
pbontrager: I think we should be "dumb" here and just make the user responsible for this. We should just enforce the output type for step and input type for action and let them implement this without a standard pattern. I think it's faster and less error prone to start simpler and dumber and add more later.

forge/rl/environments/chat.py
```diff
@@ -0,0 +1,221 @@
...
+from forge.rl.environments.base import Action, Environment, Observation, State
+
+
+class Message(TypedDict):
+    role: str
+    content: str
+
+
+@dataclass
+class ChatAction(Action):
```
pbontrager: I think these types can't be env specific right? They will be passed to and from the policy model which will always expect the same types no matter what env is used. Should we not make these data classes and just implement one version for now?

forge/rl/environments/chat.py
```diff
@@ -0,0 +1,221 @@
...
+            )
+            self._state.history_tokens.append(system_tokens)
+
+    def reset(self) -> ChatObservation:
+        """Reset the environment to initial state.
+
+        Returns:
+            ChatObservation: Initial observation with system prompt (if any)
+        """
+        self._state.history_messages = []
```
pbontrager: I think the state data class should handle the full trajectory so you can encapsulate all of this logic inside of state and keep things simpler here.

forge/rl/environments/chat.py
```diff
@@ -0,0 +1,221 @@
...
+    ]
+    tokens = tensor([1, 2, 3, 4, 5, ...])  # tokenized entire conversation
+    """
+
+    messages: List[Message] = field(default_factory=list)
+    tokens: torch.Tensor = field(default_factory=lambda: torch.tensor([]))
+    # Inherited fields from Observation ABC: reward, done, metadata
+
+
+class ChatEnvironment(Environment):
```
pbontrager: I think right now you've actually implemented this as a dataset and the DatasetChatEnv as the env? If you look at torchtune sft dataset it already handles appending system prompts etc. I think ChatEnv should just wrap stateful dataloader and take in a dataset (might want to mock it until sft lands) and apply the tokenizer.

The question I have for you is if you think a ChatEnv should return a batch of data and then unbatch the data when we send to the policy or to always have a bs of 1 but have to spin up a lot more monarch processes (which I don't think cost anything on cpu)

forge/protocols/tokenizer.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers: This is from the previous PR, right? Can you merge main so we have the minimal set of changes in this one?

.gitignore
```diff
@@ -185,5 +185,7 @@ cover/
 # VSCode
 .vscode/
 
+PREFERENCES.md
```
ebsmothers, line 188: What's this about?

forge/rl/environments/dataset.py
```diff
@@ -0,0 +1,371 @@
...
+        Args:
+            row: A single row from the dataset
+
+        Returns:
+            The extracted content as a string
+        """
+        content = row[self.column]
+        if self.content_template:
+            # Allow template formatting with row data
+            return self.content_template.format(**row)
```
ebsmothers, line 77: How should I think about content_template here? Just an f-string to format a single message?

forge/rl/environments/base.py
```diff
@@ -0,0 +1,197 @@
...
+        Examples:
+            >>> target = MyObservation(done=False, reward=None)
+            >>> source = DifferentObservation(done=True, reward=1.0, extra="ignored")
+            >>> new_obs = env._convert_observation_type(target, source)
+            >>> print(type(new_obs), new_obs.done, new_obs.reward)
+            <class 'MyObservation'> True 1.0
+        """
+        new_obs = type(target_obs)()
+
+        for attr_name, attr_value in vars(target_obs).items():
```
ebsmothers: noob q: what's the difference between this and just doing a copy here?

forge/rl/environments/base.py
Darktex: That's not the tokenizer, this is the env transforms like we had in the notebook. For example, you can do rewards here

forge/rl/environments/chat.py
Darktex: I went back and forth on this. Maybe having a single standardized class for these is the way to go. The thing is, depending on the env people may want to dump very different things in there, so I'm not sure how to handle that

forge/rl/environments/chat.py
Darktex: I would just do batch size 1. We can introduced batched envs as a generalization later on, but I don't see a need to have them as a p0 exactly for the reason you are saying. Where we care about batching (eg in reward computation and in VLLM generation) we have batching anyway, so this is not costing us anything

forge/protocols/tokenizer.py
Darktex: yea

.gitignore
Darktex: Ehm trying to give Devmate my code style preferences but it doesn't remember them so I had to write something to reference :D

forge/rl/environments/chat.py
Darktex: It is handling the full trajectory though, right? We have the whole history and the trajectory as tokens. Or did you mean smt else?

forge/rl/environments/dataset.py
Darktex: Yup. This is useful if you want the user turn to be a combination of multiple columns (it would happen in Alpaca where you need to handle a potentially missing input column)

forge/rl/environments/base.py
Darktex: Hmmm, we could do a copy of the target obs and go back and fill the values from source and it would be the same, yeah.

Maybe Philip is right that we should avoid this problem altogether and have a single Observation/Action type with a structure that allows you to add whatever you want. I'd like to point out that this path is starting to look quite similar to having a TensorDict so it feels like we are rediscovering TorchRL...

forge/rl/environments/dataset.py
Darktex: Having a separate ChatEnv is useful, because given the interactive nature of this framework, we may end up needing a basic env that accumulates state in a convo between two agents for example, or interactively between agent and a user. It's also the simplest and most basic env to evaluate. We treat datasets as an extra delta on top where now the dataloader is providing one side of this data

forge/rl/environments/dataset.py
```diff
@@ -0,0 +1,371 @@
...
+            batch_size=1,  # Always batch size 1 for episode loading
+            shuffle=self.shuffle,
+            generator=self.generator if self.shuffle else None,
+            collate_fn=self._identity_collate_fn,
+        )
+
+        # Initialize the iterator
+        self._data_iterator = iter(self.dataloader)
+
+        return self
```
ebsmothers, line 371: Don't actually need to return anything here, right?

forge/rl/environments/dataset.py
```diff
@@ -0,0 +1,371 @@
...
+
+        obs = env.reset()  # Loads initial conversation from dataset
+        # Continue conversation with additional turns via step()
+    """
+
+    def __init__(
+        self,
+        tokenizer: TokenizerProtocol,
+        dataset_name: Optional[str] = None,
+        dataset: Optional[Dataset] = None,
```
ebsmothers, line 126: Kind of a small thing (and similar in spirit to a comment I left on the 1/3 PR): personally I prefer to just have a single blessed path for stuff like this rather than trying to support multiple paths via an if/else. It'll just keep the signature cleaner and the code easier to reason about imo

forge/rl/environments/dataset.py
ebsmothers: This is actually kinda related to Philip's comment (not to pile on too much here): I wonder if we should really inherit from ChatEnvironment at all. I know it's natural because all environments have the same methods. But really I think of the dataset chat environment as relying on ChatEnvironment as a subcomponent; fundamentally they are kind of different entities, no? Personally unless we are very clearly going to use ChatEnvironment as an environment (I'll defer to you on that, if you say we will I believe you), there is some cost to doing things like this -- we've now split the implementation in a way that's harder to understand and maintain (Btw if I am just being an anti-inheritance extremist please tell me 😃)

forge/rl/environments/dataset.py
```diff
@@ -0,0 +1,371 @@
...
+        dataset_columns = set(self.dataset.column_names)
+        for mapping in self.column_mappings:
+            if mapping.column not in dataset_columns:
+                raise ValueError(
+                    f"Column '{mapping.column}' not found in dataset. Available columns: {', '.join(dataset_columns)}"
+                )
+
+        # Set up random seed for reproducibility
+        if seed is None:
+            seed = int(torch.empty((), dtype=torch.int64).random_().item())
```
ebsmothers, line 191: In this case isn't it not reproducible though? Shouldn't we hardcode it? (Lmk if I'm missing the point here)

forge/rl/environments/dataset.py
```diff
@@ -0,0 +1,371 @@
...
+
+        # Extract metadata and store in state
+        self._state.metadata = self._extract_metadata(row)
+
+        # Construct conversation from row
+        conversation = self._construct_conversation(row)
+
+        # Add all messages from the dataset to the conversation
+        for message in conversation:
+            action = self.message_to_action(message)
```
ebsmothers, line 270: Related to my "no inheritance" comment: imo this is too much indirection for something that's basically just "tokenize the message"

PR #12, opened 2025-07-16
Initial SFT

ebsmothers:
A merge of various tune components + titan components + stuff from https://github.com/pytorch/torchtitan/pull/1387 + some local hacks to titan (https://github.com/ebsmothers/torchtitan/tree/forge).

Tons of feature gaps, tons of not-thought-out design decisions, but it does run.

```
forge run --nproc_per_node 2 apps/sft/sft.py --config apps/sft/configs/llama3_8b.yaml
```

Enumerating some of the biggest gaps (roughly stack-ranked). Will try to keep this updated, but I claim pretty much all of these can be done in follow-ups.
- [ ] No checkpoint saving
- [ ] No gradient accumulation
- [ ] Currently using causal mask for attention instead of block causal
- [ ] Currently requires packed seq len == model max seq len (discussed offline and we will keep this assumption for now)
- [ ] No normalization of loss by # of unmasked tokens
- [ ] All data setup is currently hardcoded
- [ ] No metric logging
- [ ] Hack to make config system work with titan's literal fields
- [ ] No validation support

### DIFF
--- a/apps/sft/configs/llama3_8b.yaml
+++ b/apps/sft/configs/llama3_8b.yaml
@@ -14,7 +14,7 @@ model:
 
 optimizer:
   name: AdamW
-  lr: 3e-4
+  lr: 1e-5
   eps: 1e-8
 
 lr_scheduler:


apps/sft/sft.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+        dataset = alpaca_iterable_dataset(
+            model_transform=tokenizer,
+            path="yahma/alpaca-cleaned",
+        )
+        min_seq_len_divisor = 2 * self.parallel_dims.tp * self.parallel_dims.cp
+        collate_fn = partial(padded_collate_sft, pad_to_multiple_of=min_seq_len_divisor)
+        # TODO: check this
+        dp_dim = self.parallel_dims.dp_replicate * self.parallel_dims.dp_shard
+        dataloader = StatefulDataLoader(
```
Darktex, line 107: You should import Felipe's code for this so we do packing from the get-go

apps/sft/sft.py
```diff
@@ -0,0 +1,232 @@
...
+            total=self.num_training_steps,
+            desc=f"{self.current_step}",
+        )
+
+        while self.current_step < self.num_training_steps:
+            batch = next(dataloader)
+            # Move tensors to the appropriate device
+            for k, v in batch.items():
+                if isinstance(v, torch.Tensor):
+                    batch[k] = v.to("cuda")  # TODO: hardcoded for now
```
Darktex, line 207: How does multinode work?

forge/data/collate.py
```diff
@@ -6,6 +6,87 @@
 from typing import Any, Callable
 
 import torch
+import torch.nn.functional as F
+from torch.nn.utils.rnn import pad_sequence
+
+CROSS_ENTROPY_IGNORE_IDX = -100
+
+
+def padded_collate_sft(
```
Darktex: This we should completely remove. No padding, we should support only packing -- and we already have it!

forge/data/transforms/utils.py
```diff
@@ -0,0 +1,180 @@
...
+]
+
+
+class MaskingStrategy(Enum):
+    TRAIN_ON_ALL = "train_on_all"
+    TRAIN_ON_ASSISTANT = "train_on_assistant"
+    TRAIN_ON_LAST = "train_on_last"
+
+
+class Message:
```
Darktex, line 25: Weren't we taking the Huggingface messages over?

Darktex: This is the Tune messages right?

forge/data/utils.py
```diff
@@ -0,0 +1,104 @@
...
+Role = Literal[
+    "system",  # Origin is system prompt
+    "user",  # Origin is user
+    "assistant",  # Origin is the model output
+    "ipython",  # Origin is return from a tool call
+    "tool",  # Origin is return from a tool call
+]
+
+
+def padded_collate_sft(
```
Darktex: There are two copypastas of this same function. The other one is in data/collate.py

And we should nuke both! :D

forge/datasets/alpaca.py
```diff
@@ -0,0 +1,76 @@
...
+
+from typing import Any, Callable, Optional
+
+from forge.data.transforms.tokenizers import ModelTokenizer
+from forge.data.transforms.transforms import AlpacaToMessages
+from forge.datasets.hf import HfIterableDataset
+from forge.datasets.sft import sft_iterable_dataset
+
+
+def alpaca_iterable_dataset(
```
Darktex: Works but use Felipe's class instead

apps/sft/sft.py
```diff
@@ -0,0 +1,232 @@
...
+        model_parts = self.model_parts
+        parallel_dims = self.parallel_dims
+
+        # apply context parallelism if cp is enabled
+        # ensure CP handles the separate freqs_cis buffer for each pp stage
+        inputs = input_dict["tokens"]
+        optional_context_parallel_ctx = (
+            dist_utils.create_context_parallel_ctx(
+                cp_mesh=parallel_dims.world_mesh["cp"],
+                cp_buffers=[inputs, labels] + [m.freqs_cis for m in model_parts],
```
Darktex, line 133: Can you do these calculations inside the function and just send down model_parts?

apps/sft/sft.py
```diff
@@ -0,0 +1,232 @@
...
+                    self.pp_schedule.step(
+                        inputs, target=targets, losses=losses, input_batch=inputs
+                    )
+                else:
+                    self.pp_schedule.step(
+                        target=targets, losses=losses, input_batch=inputs
+                    )
+
+            # accumulate losses across pipeline microbatches
+            # TODO: PP+FSDP unexpectedly puts the loss back to the CPU
```
Darktex, line 158: Does this also happen on Titan? This code should be super close to a copypasta of what Titan is doing

apps/sft/sft.py
ebsmothers: Yeah this is one of the first things I want to do. We will need to integrate this with titan components (basically hotswap flex and do a check that CP is not enabled)

apps/sft/sft.py
ebsmothers: Doesn't yet. I just ran this on my dev machine, nothing on MAST yet

forge/data/collate.py
ebsmothers: Yeah agreed, responded to your other comment about this above

forge/data/transforms/utils.py
ebsmothers: Yeah this is just to unblock. I think we can redesign a bit here

forge/datasets/alpaca.py
ebsmothers: You mean [this one](https://github.com/pytorch-labs/forge/blob/4ea5b2c657bdb1d6793f59633d7a9403643d8121/forge/datasets/sft.py#L15), right? This is basically just a wrapper, but I am OK to take it out if you think it's not providing much value

apps/sft/sft.py
ebsmothers: Yes this is a copypasta for now. I think we need to handle loss normalization + padding tokens a bit differently though.

apps/sft/sft.py
```diff
@@ -0,0 +1,232 @@
...
+        return loss
+
+    def train_step(self, batch) -> None:
+        # TODO
+        # with GradientAccumulation(
+        #     self.gradient_accumulation_steps,
+        #     self.model,
+        #     self.data_parallel_size,
+        # ) as grad_acc:
+        labels = batch.pop("labels")
```
felipemello1, line 184: batch may have other keys nows like metrics and masks to unpack sentences, eg doc_id. Need to make sure that 'forward_backward' is ok with it. In tune, i just added '**kwargs' to the 'forward' call, so those were no opp.

apps/sft/sft.py
```diff
@@ -0,0 +1,232 @@
...
+        self.optimizers.zero_grad()
+
+        self.pbar = tqdm(
+            initial=0,
+            total=self.num_training_steps,
+            desc=f"{self.current_step}",
+        )
+
+        while self.current_step < self.num_training_steps:
+            batch = next(dataloader)
```
felipemello1, line 203: batch will have 'metrics'. Prob need to add logger here too. There are references in my tune PR.

apps/sft/sft.py
felipemello1, line 1: pretty nice that stf is only 230 lines. Hopefully its still intuitive to the user how to hack it if they want to.

forge/_cli/download.py
felipemello1, line 1: might want to delete /scripts/download.py, if this is here

apps/sft/sft.py
ebsmothers: Yeah the one thing is that many instance variables are defined inside of titan, so the user will need to know and understand those

apps/sft/sft.py
ebsmothers: Yeah haven't touched logging yet, but that can come soon after this initial version

forge/datasets/alpaca.py
Darktex: Yeah I'd just use that one for now, but it's a minor thing either way

apps/sft/sft.py
ebsmothers: You mean the fact that we have to explicitly pass the buffers like this? If so I agree, this is just what the [titan API](https://github.com/pytorch/torchtitan/blob/d69a7378cc70e93b1061c9d68ff1c8d54e1594d9/torchtitan/distributed/utils.py#L155-L161) requires. Personally I modified it a bit when reimplementing in tune, is [this](https://github.com/pytorch/torchtune/blob/a036e66842c18d08ee3638c55ed4bc45e944cb4d/torchtune/training/_distributed.py#L807-L812) more what you have in mind?

ebsmothers: More importantly there is also this bug in tune related to CP: https://github.com/pytorch/torchtune/pull/2879. Basically if the seq len in the current batch does not match the model's max seq len (which will happen if we're not packing, and even if we are packing depending on the exact strategy), if we naively shard the RoPE buffers identically to the model inputs we will wind up with a misalignment of RoPE positions. The tune user fixed by leaving the RoPE buffers unsharded and passing input_pos to the model forward to do the correct indexing. However, input_pos is not an arg in titan's transformer implementation, so this won't work here. The way I see it, there are a few paths to handle this:

1) Go all in on packed, always require model seq len == batch seq len
2) Try to extend titan's model implementation to have an equivalent fix to the tune PR
3) Maybe some changes to CP APIs to support custom user-specified sharding of buffers?

(1) somewhat limits the generalizability of some of the packing APIs that @felipemello1 had worked on. It also further raises the visibility of the CP + flex issue -- without support for naively padded batches we likely wind up forcing people to use the technically incorrect pure causal mask until flex CP support lands. (2) sounds like a pain for titan, and it's not yet obvious to me that (3) is even feasible/advisable. Tagging a few folks for thoughts: @Darktex @tianyu-l @XilunWu @pbontrager (who also has an old dynamic batching proposal from last year as an alternative to packing that we could resuscitate, but that's a whole other can of worms).

forge/datasets/alpaca.py
ebsmothers: OK deleted this so marking as resolved

forge/data/transforms/utils.py
ebsmothers: @Darktex do you think this is a blocker for this PR? Personally I think there are enough headaches with titan interface, getting the training loop right, etc that I'd personally punt this refactor for a follow-up (since things work with this, that's kinda my top priority atm). Lmk if that sounds reasonable to you though

forge/datasets/packed.py
```diff
@@ -630,7 +637,7 @@ def finalize_pack(
                 if pack["input_pos"]
                 else torch.empty(0, dtype=torch.long)
             ),
-            "metrics": pack["metrics"],
+            # "metrics": pack["metrics"],
```
felipemello1, line 640: why comment it out? i think its fine to keep it. But in the recipe you have to pop them. IMO a better solution.

apps/sft/sft.py
```diff
@@ -85,18 +86,24 @@ def setup_data(self):
...
             model_transform=tokenizer,
+            message_transform=AlpacaToMessages(),
             path="yahma/alpaca-cleaned",
+            split="train",
+        )
+        packer = TextPacker(padding_idx=0)
+        dataset = PackedDataset(
+            dataset=dataset,
+            packer=packer,
+            target_tokens_per_pack=1024,
```
felipemello1: do we have it in the cfg?

forge/datasets/packed.py
ebsmothers: For the initial version it was just one less thing to test. Assuming it works smoothly I can add it back in

apps/sft/sft.py
ebsmothers: Discussed this a bit over chat on Friday, I think we will currently need to hardcode this to match model max seq len

forge/_cli/download.py
```diff
@@ -0,0 +1,148 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
joecummings, line 1: Let's not bring this over yet?

forge/_cli/run.py
```diff
@@ -0,0 +1,123 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
joecummings, line 1: Lets just launch with torchrun for now?

README.md
```diff
@@ -8,13 +8,27 @@
...
-pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu126 --force-reinstall
-[For AMD GPU] pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/rocm6.3 --force-reinstall
 ```
+
+## Quick Start
+
+To run SFT for Llama3 8B, run
+
+```bash
+forge download meta-llama/Meta-Llama-3.1-8B-Instruct --output-dir /tmp/Meta-Llama-3.1-8B-Instruct --ignore-patterns "original/consolidated.00.pth"
```
joecummings, line 28: huggingface-cli download ...

README.md
```diff
@@ -8,13 +8,27 @@
...
+forge run --nproc_per_node 2 apps/sft/sft.py --config apps/sft/configs/llama3_8b.yaml
```
joecummings, line 29: torchrun --nproc-per-node apps/sft/main.py --config apps/sft/llama3_8b.yaml

joecummings: No more CLI system :/

ebsmothers:
@joecummings are the CLI comments blockers? It runs, right? I think we can separately figure out the CLI experience (e.g. @allenwang28 is already working on some of the Monarch/MAST stuff anyways so I expect there to be some nontrivial changes here regardless)

PR #13, opened 2025-07-17
[WIP] Proposed directory structure

joecummings:


joecummings:
> High-level question: are we considering packaging here? The inclusion of src/ seems to at least implicitly indicate that we are. That can definitely impose some additional constraints (and would require figuring out how we want to handle apps). Setting that aside, this looks reasonable to me.

Shore, lets consider it.

proposed_structure.md
```diff
@@ -0,0 +1,46 @@
...
+            monarch_utils.py
+            logging.py
+            profiling.py
+            ...
+    tests/
+    README.md
+    ...
+
+Assumptions:
+* `uv` for package management (Eli and Ed believe it's ready for primetime)
```
Darktex, line 40: +1

proposed_structure.md
```diff
@@ -0,0 +1,46 @@
...
+* Users will work with a fork of the repo
+    `uv run python apps/sft/main.py --config=apps/sft/example.yaml`
```
Darktex, line 42: we can also make a CLI command to make a new app

proposed_structure.md
```diff
@@ -0,0 +1,46 @@
+forge/
+    apps/
+        sft/
+            main.py
+            example.yaml
```
Darktex, line 5: This is fine for now. In the future we may want to consider hydra's hierarchical configs so users could configure data in like data_config.yaml, models in models.yaml etc.

proposed_structure.md
```diff
@@ -0,0 +1,46 @@
...
+        grpo/
+            main.py
+            example.yaml
+        dpo/
+            main.py
+            example.yaml
+        ...
+    src/
```
Darktex, line 13: This works for me! Actually better than what we have now, I agree that it's best to NOT have a `rl/` dir

joecummings: Yeah the thinking is to keep things as flat as possible right now and only subfolder when necessary :)

Let the emergent patterns reveal themselves!

proposed_structure.md
joecummings: Interesting - what would the CLI command do exactly? Create a new folder in `apps` with the chosen name and a default config + add it to the registry? Or something more complex?

proposed_structure.md
pbontrager: Where would be the best place to put actors? We'll have an Env actor (that runs envs), a Policy (vllm) Actor, an Actor to run rewards and likely an actor for the Replay buffer. Should their be a "src/actors" folder or should their just be an actors file under the apps?

joecummings: Apps should only contain code that's NOT to be duplicated across other apps or useful as building blocks. So, if actors might be re-used (I think they will be), then they should live under `src/`.

Now how should they live under `src/`? 

**Option 1: actors/ directory**

```
src/
    actors/
		env.py
        policy.py
        reward.py
        replay_buffer.py
```

Double clicking into e.g. `policy.py` would give you:

```python
class VllmActor(Actor):
	...
```

Pros: All actors go here, easy to understand
Cons: might cause some indirection b/c these actors touch other parts of src/
***

**Option 2: co-located w/ use case**

```
src/
	envs/
		env_actor.py  < ---- HERE
		chat.py
		code.py
	training/
		replay_buffer.py < ---- HERE
		monarch_utils.py
	...
```

Pros: Co-located w/ the concepts used
Cons: No immediate place to put `policy.py` right now, up to each person to decide where their actor goes

joecummings: cc @Darktex for thoughts here

pbontrager: Let's make an actor folder for now (can put the trainer in there). We can revisit it later.

joecummings:
Completed in #19 #16 #15

PR #16, opened 2025-07-23
[2/n] [CLEANUP] Rename `_scratch` and update recipe file names to be inline with proposal

joecummings:


apps/sft/llama3_8b.yaml
pbontrager, line 1: I think we shouldn't go in the direction of configs per model again. It'd be better if the model name had a default value like `<model-name>` and the user had to override it in the cli or config before running. Also, by default, users should be able to change models with a one line name change.

joecummings: Excellent feedback that should have a little RFC :)

PR #17, opened 2025-07-23
[don't review] just demoing commit flow

ebsmothers:


apps/sft/main.py
```diff
@@ -11,6 +11,7 @@
 
 import torch
 
+print("hi")
```
ebsmothers: don't print this

PR #19, opened 2025-07-23
[3/n] [CLEANUP] Aka refactor aka the big kahuna

joecummings:
```
.
├── actors
│   ├── collector.py
│   ├── __init__.py
├── _cli
│   ├── config.py
│   ├── download.py
│   ├── forge.py
│   ├── __init__.py
│   ├── run.py
│   └── subcommand.py
├── controller
│   ├── __init__.py
│   ├── README.md
│   └── stack.py
├── data
│   ├── collate.py
│   ├── dataset_metrics
│   │   ├── __init__.py
│   │   ├── metric_agg_handlers.py
│   │   ├── metric_aggregator.py
│   │   ├── metric_transform.py
│   │   └── readme.md
│   ├── datasets
│   │   ├── dataset.py
│   │   ├── hf_dataset.py
│   │   ├── packed.py
│   │   └── sft_dataset.py
│   ├── __init__.py
│   ├── replay_buffer.py
│   ├── tokenizer.py
│   └── utils.py
├── envs
│   ├── chat.py
│   └── __init__.py
├── __init__.py
├── interfaces.py
└── types.py
```

**Notable changes**:

1. Consolidating RL-related interfaces and types into top-level files called types.py and interfaces.py. The rationale here is that these types and interfaces are important to letting users build their own RL training loops. Unfortunately, they don't fit cleanly within their own lanes e.g. Env needs to know about the Observation, State, and Action types and Actors do as well. Therefore, we put them at the top level so that everyone is always very clear as to what contracts they should follow. I have not touched all the dataclasses / interfaces in `data/` yet b/c SFT data work is much more straightforward and I don't want to confuse people. Open to thoughts here though.

2. I moved @allenwang28's stuff to a proper `toy_rl` app. Made more sense to keep all Environments, Policies, etc in the same project place. IMO this is how things should be structured when we go to make the first RL recipe.

**This is not the end-all be-all format of forge!** But hopefully this provides us with a good North Star so that all contributors can jump in and be immediately productive.

### DIFF
--- a/apps/sft/llama3_8b.yaml
+++ b/apps/sft/llama3_8b.yaml
@@ -7,6 +7,11 @@
 #   enable_tensorboard: true
 #   save_tb_folder: "tb"
 
+# TODO: required by torchtitan
+# https://github.com/pytorch/torchtitan/blob/2f1c814da071cc8ad165d00be6f9c1a66f8e1cce/torchtitan/distributed/utils.py#L265
+comm:
+  trace_buf_size: 0
+
 model:
   name: llama3
   flavor: 8B
--- a/apps/sft/main.py
+++ b/apps/sft/main.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
 import sys
 from dataclasses import asdict
 from functools import partial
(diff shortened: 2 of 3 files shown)


allenwang28:
nice, the big kahuna! Mostly lgtm, wondering if there's a better spot to put monarch_utils? IMO it's not necessarily train or RL specific

joecummings:
> nice, the big kahuna! Mostly lgtm, wondering if there's a better spot to put monarch_utils? IMO it's not necessarily train or RL specific

Good callout, here are some other options:

1. `distributed/` subfolder with relevant monarch utils and maybe some titan stuff
2. `controller/` subfolder with relevant monarch utils
3. Throw it into `data/`

I'm not a fan of 3 to be fair.

forge/interfaces.py
```diff
@@ -0,0 +1,174 @@
...
+        Returns:
+            str: The decoded text.
+        """
+        pass
+
+
+class ModelTokenizer(ABC):
+    """
+    Abstract tokenizer that implements model-specific special token logic in
+    the ``tokenize_messages`` method. See :class:`~torchtune.models.llama3.Llama3Tokenizer`
```
ebsmothers, line 151: o god, i bet i left quite a few of these in here
```suggestion
    the ``tokenize_messages`` method. See :class:`~torchtune.models.llama3.Llama3Tokenizer`
```

joecummings: Yeah, not gunna clean those up :)

allenwang28:
> > nice, the big kahuna! Mostly lgtm, wondering if there's a better spot to put monarch_utils? IMO it's not necessarily train or RL specific
> 
> Good callout, here are some other options:
> 
> 1. `distributed/` subfolder with relevant monarch utils and maybe some titan stuff
> 2. `controller/` subfolder with relevant monarch utils
> 3. Throw it into `data/`
> 
> I'm not a fan of 3 to be fair.

I like `controller/` personally

PR #21, opened 2025-07-23
vLLM Policy

pbontrager:
vLLM Policy (vLLM workers) + Router Actor (vLLM scheduler) to run generation.

- Aiming for basic working implementation
- Run output [here](https://www.internalfb.com/phabricator/paste/view/P1887136730)


forge/rl/policy.py
```diff
@@ -0,0 +1,287 @@
...
+import pdb
+
+logger = logging.getLogger(__name__)
+
+@dataclass
+class PolicyRouter(Actor):
+    # TODO: Add dp support
+    policy: Actor
+    sampling_params: SamplingParams = None
+    lora_request: LoRARequest = None
```
ebsmothers: Is this strictly required?

forge/rl/policy.py
```diff
@@ -0,0 +1,287 @@
...
+    prompt = "Tell me a joke"
+    response = await router.generate.call(prompt)
+    print(f"User: {prompt}\nAssistant: {response}")
+
+    await router.shutdown.call()
+
+if __name__ == "__main__":
+    config = {
+        "model": "meta-llama/Llama-3.1-8B-Instruct",
+        "tensor_parallel_size": 2,
```
ebsmothers: Dumb q: how are we able to use TP? Iiuc from the policy_mesh definition you are setting CUDA_VISIBLE_DEVICES=1?

forge/rl/policy.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+    @endpoint
+    async def generate(self, prompt: str, priority: int = 0):
+        self.request_id += 1 % sys.maxsize
+        request_id = str(self.request_id) # implement from a counter
+
+        prompt = convert_input(prompt)
+        if self.sampling_params is None:
+            self.sampling_params = get_default_sampling_params(self.vllm_args)
```
ebsmothers: Why isn't this just done in setup?

forge/rl/policy.py
```diff
@@ -0,0 +1,287 @@
...
+            request_id, prompt, self.sampling_params, arrival_time, self.lora_request,
+            tokenization_kwargs, trace_headers, self.prompt_adapter_request,
+            priority)
+
+        if self.sampling_params.n == 1:
+            self.output_processor.add_request(request, prompt_str, None, 0)
+            request = Request.from_engine_core_request(request)
+            # TODO: mm_hash and sturcutured_output
+            self.requests["request_id"] = asyncio.Future()
+            self.scheduer.add_request(request)
```
ebsmothers:
```suggestion
            self.scheduler.add_request(request)
```

forge/rl/policy.py
```diff
@@ -0,0 +1,287 @@
...
+            #                                     parent_req, idx)
+            #     child_request = Request.from_engine_core_request(child_request)
+            #     self.scheduer.add_request(chile_request)
+        
+        return await self.request[request_id]
+
+    @endpoint
+    async def run(self):
+        # TODO: add support for `iteration_stats`
+        # TODO: move postprocessing out of loop to not block
```
ebsmothers: noob q: why can't you just add await on L122?

forge/rl/policy.py
```diff
@@ -0,0 +1,287 @@
...
+from vllm.v1.core.sched.output import SchedulerOutput
+
+
+import pdb
+
+logger = logging.getLogger(__name__)
+
+@dataclass
+class PolicyRouter(Actor):
+    # TODO: Add dp support
```
ebsmothers: Just for my own understanding, what's the issue with DP currently?

forge/rl/policy.py
```diff
@@ -0,0 +1,287 @@
...
+    policy: Actor
```
allenwang28: I guess this is an ActorMesh right?

forge/rl/policy.py
```diff
@@ -0,0 +1,287 @@
...
+        
+        return await self.request[request_id]
+
+    @endpoint
+    async def run(self):
+        # TODO: add support for `iteration_stats`
+        # TODO: move postprocessing out of loop to not block
+        self.running = True
+        while self.running:
+            scheduler_output = self.scheduler.schedule()
```
allenwang28: to solve your async woes, here's what I'd do:

- on `generate`, put request in the PolicyRouter's owned `asyncio.Queue` 
- in this loop, you `await` on waiting for `generate` to be ready to relinquish control

forge/rl/policy.py
pbontrager: No, I was just exposing the options that vLLM was taking, but we don't need to for now

forge/rl/policy.py
pbontrager: More for organization. All of the preproc/setup request stuff should go with the environment (to not block generation) not the policy, I just grouped it all there for now to get it working before I work on the environment.

forge/rl/policy.py
pbontrager: The way vLLM is designed is that this loop runs continuously and requests just get added in by the scheduler. So when you call generate, it'll batch that request in for the workers and every time you call model.execute_model it'll do a fixed amount of work on whatever is currently in the batch or unfinished decoding. So we await on generate, which waits on run adding a finished request in the self.requests dictionary whenever it gets to it.

forge/rl/policy.py
pbontrager: vLLM doesn't support it very well (basic ray implementation). We'll just do the same thing by spinning up multiple meshes and sending work to whoever is least busy.

forge/rl/policy.py
pbontrager: Is this still true? I couldn't import it.

forge/rl/policy.py
pbontrager: I'm currently using a dictionary for this (requests dict with futures as values) since requests don't necessarily finish in the order they're submitted. I guess I don't understand how I can use a queue for this task.

PR #22, opened 2025-07-24
Initial uv builds

joecummings:
1. Change to follow src/ format: https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/
2. Relies on `uv` to handle package management
3. Automatically relies on torchtitan nightly and torch nightly and monarch-nightly


22quinn:
Adding uv is nice! I'm having trouble setting up dev env so tried this PR but got an error. Somehow it's looking for pyproject.toml in the second level forge folder?
> error: Workspace member `/data/users/quinnzhu/gitrepos/forge/forge` is missing a `pyproject.toml` (matches: `forge`)

joecummings:
> Adding uv is nice! I'm having trouble setting up dev env so tried this PR but got an error. Somehow it's looking for pyproject.toml in the second level forge folder?
> 
> > error: Workspace member `/data/users/quinnzhu/gitrepos/forge/forge` is missing a `pyproject.toml` (matches: `forge`)

Interesting, can you share your exact commands? Are you just trying to build with `uv sync`?

src/forge/data/replay_buffer.py
```diff
@@ -18,7 +18,7 @@
 # Silly replay buffer implementation for testing.
 # One nice thing if we implement our own Replay buffer is that
 # we can wrap RDMA calls / torchstore calls here.
-class ReplayBuffer(ReplayBuffer):
```
pbontrager, line 21: Not a big deal, but I feel it'd be easier to just call this ReplayBuffer and replace it with the real one in the future. If we don't expect to support multiple ones, we don't need special names.

pyproject.toml
```diff
@@ -10,22 +10,16 @@ authors = [
 ]
 keywords = ["pytorch", "training", "llm"]
 dependencies = [
-    # Stateful Dataloader
+    # PyTorch
     "torchdata>=0.8.0",
-
+    "torchtitan",
+    "torch",
+    "torchmonarch-nightly",
```
pbontrager, line 17: Should nightlies here be handled the same way as with titan and torch? Or we basically never want stable monarch until we've developed more?

README.md
```diff
@@ -8,25 +8,20 @@
...
 ```bash
-pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu126 --force-reinstall
-[For AMD GPU] pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/rocm6.3 --force-reinstall
-git clone https://github.com/pytorch/torchtitan
-pip install -e ./torchtitan
+pip install uv
 git clone https://github.com/pytorch-labs/forge
 cd forge
-pip install -r requirements.txt
+uv sync
```
pbontrager, line 15: This is awesome! Does this create an new env or maybe we should include instructions for a new uv env here too?

README.md
```diff
@@ -8,25 +8,20 @@
...
 ```
 
 ## Quick Start
 
 To run SFT for Llama3 8B, run
 
 ```bash
-forge download meta-llama/Meta-Llama-3.1-8B-Instruct --output-dir /tmp/Meta-Llama-3.1-8B-Instruct --ignore-patterns "original/consolidated.00.pth"
-forge run --nproc_per_node 2 apps/sft/main.py --config apps/sft/llama3_8b.yaml
+uv run forge download meta-llama/Meta-Llama-3.1-8B-Instruct --output-dir /tmp/Meta-Llama-3.1-8B-Instruct --ignore-patterns "original/consolidated.00.pth"
```
pbontrager, line 23: More of a status question, but are we expecting automatic downloads from DCP to be working soon? It's even more important now since vllm will also grab the model from the same cache

pbontrager:
Could you add [vllm](https://docs.vllm.ai/en/latest/getting_started/installation/gpu.html#install-the-latest-code_1) to pyproject too?

22quinn:
> Interesting, can you share your exact commands? Are you just trying to build with `uv sync`?

Yes, just a simple `uv sync`, without any change to this PR.

pyproject.toml
joecummings: Defaulting everything pretty much to nightlies while we're developing since it seems like there's important changes being pushed daily across all three repos (titan, core, and monarch)

README.md
joecummings: Yep, this should create the new env for you.

joecummings:
> > Interesting, can you share your exact commands? Are you just trying to build with `uv sync`?
> 
> Yes, just a simple `uv sync`, without any change to this PR.

What happens if you remove the uv.lock file and create your own environment based on the specs in pyproject.toml? Does that work?

I'm unable to repro the error on my side

joecummings:
> Could you add [vllm](https://docs.vllm.ai/en/latest/getting_started/installation/gpu.html#install-the-latest-code_1) to pyproject too?

Regular vLLM or do you want nightlies there as well?

README.md
joecummings: Deflection: @ebsmothers Is checkpointing supposed to live in Titan SMR or in Forge?

joecummings:
> > > Interesting, can you share your exact commands? Are you just trying to build with `uv sync`?
> > 
> > 
> > Yes, just a simple `uv sync`, without any change to this PR.
> 
> What happens if you remove the uv.lock file and create your own environment based on the specs in pyproject.toml? Does that work?
> 
> I'm unable to repro the error on my side

I'll go ahead and merge but this is on my radar and I'll make sure the uv commands work well for the repo.

pyproject.toml
ebsmothers: So this is just an issue of Monarch publishing nightlies differently than titan and core do? Is it worth trying to consolidate to a single shared approach there?

README.md
ebsmothers: Mostly in titan, though we may want to inherit from their checkpointer if we need our own handling (e.g. I have not gotten to dataloader checkpointing yet, not sure if that will be a 1:1 mapping). But definitely for @pbontrager's question, I think it's more of a titan thing

PR #27, opened 2025-07-28
Fix the Monarch APIs for toy_rl example

allenwang28:
Updates the Monarch APIs in `toy_rl` to use the latest APIs.

A lot of the APIs *should* be a bit more stable after [this](https://fb.workplace.com/groups/996292674996363/posts/1247681839857444) has landed, fingers crossed


apps/toy_rl/test_toy_rl.py
```diff
@@ -280,7 +286,7 @@ async def test_full_rl_pipeline_simulation(self):
         max_collector_steps = 5
         proc = await local_proc_mesh(gpus=1)
         policy = await proc.spawn("policy", ToyPolicy, action_range=(-2.0, 2.0))
-        replay_buffer = await proc.spawn("replay_buffer", ReplayBuffer)
+        replay_buffer = await proc.spawn("replay_buffer", SimpleReplayBuffer)
```
pbontrager: This is going to have to be changed again in the future when we remove SimpleReplayBuffer. I'm fine with keeping this change, but an alternative would be to just rename SimpleReplayBuffer to ReplayBuffer and not have to rename test stuff so often.

apps/toy_rl/test_toy_rl.py
```diff
@@ -4,23 +4,29 @@
 # This source code is licensed under the BSD-style license found in the
```
pbontrager: I think tests should be under the tests folder

apps/toy_rl/main.py
```diff
@@ -10,121 +10,15 @@
 """
```
pbontrager: I don't feel strongly about this at the moment, but I would just leave main and components combined and introduce less structure for now

allenwang28:
@pbontrager thanks for the review, I agree with your comments and applied them. Note: I did have to modify some stack APIs due to Monarch update. Once we get things working on MAST we should pin...

PR #32, opened 2025-07-31
ReplayBuffer v0

joecummings:
### Context

We need a replay buffer. See internal doc (dm me for it) for implementation details.

**Everything needing thorough review will be inline called out.**

Tests:
```
(forge-dev-6549113) [jrcummings@devvm4767.pnb0 ~/projects/joe-forge (init-replay-buffer)]$ python -m pytest tests/test_replay_buffer.py::TestReplayBuffer
================================================================================================================ test session starts ================================================================================================================
platform linux -- Python 3.10.18, pytest-8.4.1, pluggy-1.6.0
rootdir: /home/jrcummings/projects/joe-forge
configfile: pyproject.toml
plugins: typeguard-4.2.1, hypothesis-6.136.6, asyncio-1.1.0
asyncio: mode=strict, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 6 items

tests/test_replay_buffer.py .[-]I0804 11:10:42.277547 4128248 fbcode/monarch/monarch_hyperactor/src/actor_mesh.rs:386] Dropping PythonActorMesh: replay_buffer
.[-]I0804 11:10:45.207152 4128248 fbcode/monarch/monarch_hyperactor/src/actor_mesh.rs:386] Dropping PythonActorMesh: replay_buffer
.[-]I0804 11:10:48.097360 4128248 fbcode/monarch/monarch_hyperactor/src/actor_mesh.rs:386] Dropping PythonActorMesh: replay_buffer
.[-]I0804 11:10:50.940287 4128248 fbcode/monarch/monarch_hyperactor/src/actor_mesh.rs:386] Dropping PythonActorMesh: replay_buffer
.[-]I0804 11:10:54.075980 4128248 fbcode/monarch/monarch_hyperactor/src/actor_mesh.rs:386] Dropping PythonActorMesh: replay_buffer
.[-]I0804 11:10:58.281658 4128248 fbcode/monarch/monarch_hyperactor/src/actor_mesh.rs:386] Dropping PythonActorMesh: replay_buffer
                                                                                                                                                                                                            [100%]
[-]I0804 11:10:58.295372 4128248 fbcode/monarch/monarch_hyperactor/src/actor_mesh.rs:386] Dropping PythonActorMesh: debug_manager
[-]I0804 11:10:58.295596 4128248 fbcode/monarch/monarch_hyperactor/src/actor_mesh.rs:386] Dropping PythonActorMesh: debug_manager
[-]I0804 11:10:58.318732 4128248 fbcode/monarch/monarch_hyperactor/src/actor_mesh.rs:386] Dropping PythonActorMesh: debug_manager

================================================================================================================ 6 passed in 21.33s =================================================================================================================
```

### DIFF
--- a/src/forge/data/replay_buffer.py
+++ b/src/forge/data/replay_buffer.py
@@ -48,12 +48,12 @@ class ReplayBuffer(Actor):
         # Evict old trajectories
         self._evict(curr_policy_version)
 
-        if bsz > len(self):
+        if bsz > len(self.buffer):
             print("Not enough trajectories in the buffer.")
             return None
 
         # TODO: Make this more efficient
-        idx_to_sample = self.sampler(range(len(self)), k=bsz)
+        idx_to_sample = self.sampler(range(len(self.buffer)), k=bsz)
         sorted_idxs = sorted(
             idx_to_sample, reverse=True
         )  # Sort in desc order to avoid shifting idxs
@@ -78,19 +78,12 @@ class ReplayBuffer(Actor):
         ]
 
     @endpoint
-    async def get_item(self, idx: int) -> Trajectory:
-        """Get the trajectory at the given index from Actor mesh"""
-        return self[idx]
-
-    def __getitem__(self, idx: int) -> Trajectory:
+    async def _getitem(self, idx: int) -> Trajectory:
         return self.buffer[idx]
 
     @endpoint
-    async def get_length(self) -> int:
-        """Get the length of the replay buffer from Actor mesh"""
-        return len(self)
-
-    def __len__(self) -> int:
+    async def _numel(self) -> int:
+        """Number of elements (trajectories) in the replay buffer."""
         return len(self.buffer)
 
     @endpoint
--- a/tests/test_replay_buffer.py
+++ b/tests/test_replay_buffer.py
@@ -27,8 +27,8 @@ class TestReplayBuffer:
     async def test_add(self, replay_buffer: ReplayBuffer) -> None:
         trajectory = Trajectory(policy_version=0)
         await replay_buffer.add.call_one(trajectory)
-        assert replay_buffer.get_length.call_one().get() == 1
-        assert replay_buffer.get_item.call_one(0).get() == trajectory
+        assert replay_buffer._numel.call_one().get() == 1
+        assert replay_buffer._getitem.call_one(0).get() == trajectory
         replay_buffer.clear.call_one().get()
 
     @pytest.mark.asyncio
@@ -37,9 +37,9 @@ class TestReplayBuffer:
         trajectory_1 = Trajectory(policy_version=1)
         await replay_buffer.add.call_one(trajectory_0)
         await replay_buffer.add.call_one(trajectory_1)
-        assert replay_buffer.get_length.call_one().get() == 2
-        assert replay_buffer.get_item.call_one(0).get() == trajectory_0
-        assert replay_buffer.get_item.call_one(1).get() == trajectory_1
+        assert replay_buffer._numel.call_one().get() == 2
+        assert replay_buffer._getitem.call_one(0).get() == trajectory_0
+        assert replay_buffer._getitem.call_one(1).get() == trajectory_1
         replay_buffer.clear.call_one().get()
 
     @pytest.mark.asyncio
@@ -48,9 +48,9 @@ class TestReplayBuffer:
         await replay_buffer.add.call_one(trajectory)
         state_dict = replay_buffer.state_dict.call_one().get()
         replay_buffer.clear.call_one().get()
-        assert replay_buffer.get_length.call_one().get() == 0
+        assert replay_buffer._numel.call_one().get() == 0
         await replay_buffer.load_state_dict.call_one(state_dict)
-        assert replay_buffer.get_length.call_one().get() == 1
+        assert replay_buffer._numel.call_one().get() == 1
         replay_buffer.clear.call_one().get()
 
     @pytest.mark.asyncio
@@ -59,9 +59,9 @@ class TestReplayBuffer:
         trajectory_1 = Trajectory(policy_version=1)
         await replay_buffer.add.call_one(trajectory_0)
         await replay_buffer.add.call_one(trajectory_1)
-        assert replay_buffer.get_length.call_one().get() == 2
+        assert replay_buffer._numel.call_one().get() == 2
         await replay_buffer.evict.call_one(curr_policy_version=2)
-        assert replay_buffer.get_length.call_one().get() == 1
+        assert replay_buffer._numel.call_one().get() == 1
         replay_buffer.clear.call_one().get()
 
     @pytest.mark.asyncio
@@ -70,7 +70,7 @@ class TestReplayBuffer:
         trajectory_1 = Trajectory(policy_version=1)
         await replay_buffer.add.call_one(trajectory_0)
         await replay_buffer.add.call_one(trajectory_1)
-        assert replay_buffer.get_length.call_one().get() == 2
+        assert replay_buffer._numel.call_one().get() == 2
 
         # Test a simple sampling w/ no evictions
         samples = await replay_buffer.sample.call_one(curr_policy_version=1)
@@ -99,7 +99,7 @@ class TestReplayBuffer:
         trajectory_1 = Trajectory(policy_version=1)
         await replay_buffer.add.call_one(trajectory_0)
         await replay_buffer.add.call_one(trajectory_1)
-        assert replay_buffer.get_length.call_one().get() == 2
+        assert replay_buffer._numel.call_one().get() == 2
         samples = await replay_buffer.sample.call_one(
             curr_policy_version=2, batch_size=1
         )


src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,88 @@
...
+        self.max_policy_age = max_policy_age
+        if seed is not None:
+            random.seed(seed)
+        self.sampler = random.sample
 
     @endpoint
-    async def extend(self, sample: Trajectory):
-        """Add a trajectory to the replay buffer."""
-        self.buffer.append(sample)
+    async def add(self, trajectory: Trajectory) -> None:
```
joecummings, line 29: Currently no option to extend with multiple trajectories. Figured we could add that easily in later.

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,88 @@
...
+        self.buffer.append(trajectory)
 
     @endpoint
-    async def sample(self, batch_size=1) -> list[Trajectory] | None:
+    async def sample(
+        self, curr_policy_version: int, batch_size: int | None
```
joecummings: In the training loop, I figure we should be able to have the current policy version but wanted to confirm with @pbontrager ?

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,88 @@
...
             return None
 
-        if batch_size == 1:
-            return [random.choice(self.buffer)]
-        else:
-            return random.choices(self.buffer, k=batch_size)
+        await self.evict(curr_policy_version)
+
+        # TODO: Make this more efficient
+        idx_to_sample = self.sampler(range(len(self.buffer)), k=bsz)
```
joecummings, line 56: Really basic random sampling while deleting the instances we picked.

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,88 @@
...
+        sorted_idxs = sorted(
+            idx_to_sample, reverse=True
+        )  # Sort in desc order to avoid shifting idxs
+        sampled_trajectories = [self.buffer.pop(i) for i in sorted_idxs]
+        return sampled_trajectories
 
     @endpoint
-    async def len(self) -> int:
-        """Return the length of the replay buffer."""
+    async def evict(self, curr_policy_version: int) -> None:
```
joecummings, line 64: We'll have a more complicated eviction policy eventually, but right now it's purely based on the difference between curr policy and the policy version of each trajectory.

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,88 @@
...
 
-    @endpoint
-    async def is_empty(self) -> bool:
-        """Check if the replay buffer is empty."""
-        return len(self.buffer) == 0
+    def clear(self) -> None:
+        """Clear the replay buffer immediately - dropping all trajectories."""
+        self.buffer.clear()
+
+    def state_dict(self) -> dict[str, Any]:
```
joecummings: These don't need to be async, right @allenwang28 ?

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,88 @@
...
-            return random.choices(self.buffer, k=batch_size)
+        await self.evict(curr_policy_version)
+
+        # TODO: Make this more efficient
+        idx_to_sample = self.sampler(range(len(self.buffer)), k=bsz)
+        sorted_idxs = sorted(
+            idx_to_sample, reverse=True
+        )  # Sort in desc order to avoid shifting idxs
+        sampled_trajectories = [self.buffer.pop(i) for i in sorted_idxs]
+        return sampled_trajectories
```
joecummings, line 61: There's no collation right now b/c I think that will depend on the Message updates for tokenization.

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,88 @@
...
+            curr_policy_version (int): The current policy version.
+            batch_size (int, optional): Number of trajectories to sample. If none, defaults to batch size
+                passed in at initialization.
 
         Returns:
-            A list of sampled trajectories or None if buffer is empty.
+            A list of sampled trajectories or None if there are not enough trajectories in the buffer.
         """
-        if batch_size > len(self.buffer):
+        bsz = batch_size if batch_size is not None else self.batch_size
```
ebsmothers, line 46: It feels a bit weird to me that this is an instance variable but can also be dynamic. Why not just go all the way on dynamic? (Are there some performance implications where defining a fixed size up front can benefit us?)

src/forge/data/replay_buffer.py
ebsmothers: Tell me if I'm prematurely generalizing here, but should we provide some generic plugin for user-defined eviction policies?

src/forge/data/replay_buffer.py
ebsmothers: nit but if you already defined self.len() you might as well use it here

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,88 @@
...
-    def __init__(self):
+    def __init__(
+        self, batch_size: int, max_policy_age: int, seed: int | None = None
+    ) -> None:
         self.buffer: list[Trajectory] = []
+        self.batch_size = batch_size
+        self.max_policy_age = max_policy_age
+        if seed is not None:
+            random.seed(seed)
+        self.sampler = random.sample
```
ebsmothers, line 26: parametrize this?

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,88 @@
...
+class ReplayBuffer(Actor, Stateful):
     """Simple in-memory replay buffer implementation."""
 
-    def __init__(self):
+    def __init__(
+        self, batch_size: int, max_policy_age: int, seed: int | None = None
+    ) -> None:
         self.buffer: list[Trajectory] = []
+        self.batch_size = batch_size
+        self.max_policy_age = max_policy_age
```
ebsmothers, line 23: kinda similar to my other comment, but this feels overly-tied-to one specific eviction policy

tests/test_replay_buffer.py
```diff
@@ -0,0 +1,77 @@
+"""Test for data/replay_buffer.py"""
+
+import re
+
+from src.forge.data.replay_buffer import ReplayBuffer
+from src.forge.interfaces import Trajectory
```
ebsmothers: Why do we need src in the import paths here?

src/forge/data/replay_buffer.py
joecummings: Hey, you're prematurely generalizing.

src/forge/data/replay_buffer.py
joecummings: I was just thinking of the UX perspective that it would be nice to be able to specify a batch size once and then call `buffer.sample()` unless you wanted to override it but open to thoughts here.

I would think that compile might play nicer with a fixed size batch size but perhaps we can just mark as dynamic?

src/forge/data/replay_buffer.py
joecummings: Look at M2

src/forge/data/replay_buffer.py
joecummings: You are correct, I didn't want to over generalize to start.

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,108 @@
...
 
         Returns:
-            A list of sampled trajectories or None if buffer is empty.
+            A list of sampled trajectories or None if there are not enough trajectories in the buffer.
         """
-        if batch_size > len(self.buffer):
+        bsz = batch_size if batch_size is not None else self.batch_size
+
+        # Evict old trajectories
+        self._evict(curr_policy_version)
```
Jack-Khuu, line 49: QQ: Why do we evict on sample instead of based on buffer state ("full")?

tests/test_replay_buffer.py
```diff
@@ -0,0 +1,109 @@
...
+
+import pytest
+import pytest_asyncio
+from forge.data.replay_buffer import ReplayBuffer
+from forge.types import Trajectory
+
+from monarch.actor import proc_mesh
+
+
+class TestReplayBuffer:
```
Jack-Khuu, line 17: Do we want to test for (anti-)behavior of multiple Trajectory's with the same policy?

src/forge/data/replay_buffer.py
joecummings: Good question. 

This current implementation does not assume a fixed length buffer, therefore there is no concept of "full". We could go that way in which we evict only based on length of time in buffer. However, you can imagine more complex eviction policies to come (even if they are not supported right now). Therefore, the standard way those would be called is right before the actual sampling takes place.

tests/test_replay_buffer.py
joecummings: I believe it's fine to have multiple trajectories with the same policy.

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,108 @@
...
+        return len(self)
+
+    def __len__(self) -> int:
         return len(self.buffer)
 
     @endpoint
-    async def is_empty(self) -> bool:
-        """Check if the replay buffer is empty."""
-        return len(self.buffer) == 0
+    async def clear(self) -> None:
```
pbontrager, line 90: What is the purpose of this method?

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,108 @@
...
-        """Return the length of the replay buffer."""
+    async def get_item(self, idx: int) -> Trajectory:
+        """Get the trajectory at the given index from Actor mesh"""
+        return self[idx]
+
+    def __getitem__(self, idx: int) -> Trajectory:
+        return self.buffer[idx]
+
+    @endpoint
+    async def get_length(self) -> int:
```
pbontrager: Unclear if this returns the size of the buffer or the number of items in the buffer

src/forge/data/replay_buffer.py
```diff
@@ -4,55 +4,108 @@
...
+        self.buffer = [
+            trajectory
+            for trajectory in self.buffer
+            if (curr_policy_version - trajectory.policy_version) <= self.max_policy_age
+        ]
 
     @endpoint
-    async def len(self) -> int:
-        """Return the length of the replay buffer."""
+    async def get_item(self, idx: int) -> Trajectory:
```
pbontrager: Maybe we should stick to a generator interface similar to streamed datasets

src/forge/data/replay_buffer.py
pbontrager, line 1: I like this. Around the interface I think we should keep the number of endpoints low. A generator method (runs sample and evict each time), an add method, and state_dict methods are probably enough.

src/forge/types.py
pbontrager, line 1: I wonder if there should be a dataloader type/protocol that the buffer and existing dataloader conform too? Not a priority.

src/forge/data/replay_buffer.py
joecummings: In this implementation, those are the same thing.

src/forge/data/replay_buffer.py
joecummings: It clears the buffer - evicts everything. The use case I'm thinking of here is that you get a new policy and want to drop everything from the old policies. 

Another way of doing this would be to already expand the eviction policies to include something like "ALL" that will evict everything.

src/forge/data/replay_buffer.py
joecummings: This was mainly for testing purposes I suppose.

src/forge/data/replay_buffer.py
joecummings: The number of policies in the replay buffer (and other stats) will be helpful eventually.

PR #34, opened 2025-08-01
Adds a ForgeActor and a MAST proc mesh allocator

allenwang28:
This PR introduces two pieces:
1. `ForgeActor`
2. a `forge.single_controller.get_proc_mesh` functionality

and I guess 3rd, a minimal example that runs the MAST workload in _scratch/mast.py

# ForgeActor
Logging doesn't work by default in Monarch Actors - I figure it would be useful to add our own logging support in Forge through a custom Forge actor.

ForgeActor could be the location to add more common actor functionality, like torch.distributed init stuff, etc. - but for now it's just handling logging.

Right now it will translate any `logger.info` calls into something like this:
```
[TestActor-1/8] 2025-08-01 07:36:58 INFO Initializing process group `gloo`:
  MASTER_ADDR = localhost
  MASTER_PORT = 12345
  RANK        = 1
  WORLD_SIZE  = 8
```

# MAST support
This will only work with the fbcode internal build of Monarch. The vision: just by toggling the config from `local` to `mast`, you should be able to run the same workload on MAST. It should give you an illusion that you're just developing on a big cluster as if it was local!

However there are a few papercuts!
- Under the hood, we're using conda:pack. Conda pack does not play well with editable pip packages - this is a problem because I imagine that we would want to make changes to e.g. torchtitan, forge, etc. - and it's possible that these are not packaged correctly and fail silently
- Log spam - Monarch and NCCL both like to spam logs, there's currently not a way to turn that down to allow us to focus on Forge specific logs.
- OilFS/Manifold integration - if you run on MAST, you will have to change your paths to use OilFS. That breaks the illusion if you're used to using `/tmp`. Maybe we can just always run on OilFS?
- MAST allocator needs async APIs for now - Monarch has capabilities to run in both sync and async mode, but the MAST APIs don't support this yet. For now we have to use async, but fortunately if we want to switch to sync it won't be that heavy of a lift.
- Please be mindful of MAST allocations - go to bunnylol mast to teardown your workloads.

Note - I've raised these issues to the devX team, but we could benefit from an owner on the Forge side to own the Forge DevX.

# How to run
```
$ cd _scratch
$ python -m mast
```

Currently editing the config to switch `config.scheduler` between `local` and `mast`, here are example logs:
- Local: [P1889111967](https://www.internalfb.com/phabricator/paste/view/P1889111967)
- MAST: [P1889114125](https://www.internalfb.com/intern/everpaste/?handle=GC-pRh9psRQX4FIEADxFWI7PX89kbsIXAAAB&phabricator_paste_number=1889114125)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ pip install uv
 git clone https://github.com/pytorch-labs/forge
 cd forge
 uv sync
+
+# Or for dev install:
+uv sync --all-extras
 ```
 
 ## Quick Start
--- a/_scratch/stacked_envs.py
+++ b/_scratch/stacked_envs.py
@@ -12,9 +12,8 @@ Run this with python -m examples.stacked_envs
 
 import asyncio
 
-from forge.monarch_utils.stack import stack
-from monarch.actor_mesh import Actor, endpoint
-from monarch.proc_mesh import proc_mesh
+from forge.controller.stack import stack
+from monarch.actor import Actor, endpoint, proc_mesh
 
 
 class Environment(Actor):
@@ -42,7 +41,9 @@ async def main():
     # note that interface can be deduced, but adding it explicitly
     # helps the type checker
     envs = stack(browser, coding, interface=Environment)
-    await envs.step.call()
+    # Temporary hack due to Monarch changes
+    res = envs.step.call()
+    [await r for r in res]
 
 
 if __name__ == "__main__":
--- a/apps/sft/llama3_8b.yaml
+++ b/apps/sft/llama3_8b.yaml
@@ -27,7 +27,7 @@ lr_scheduler:
 
 training:
   local_batch_size: 1
-  seq_len: 8192
+  seq_len: 2048
   max_norm: 1.0
   steps: 1000
   compile: false
--- a/apps/sft/main.py
+++ b/apps/sft/main.py
@@ -16,7 +16,7 @@ import torchtitan.experiments.forge.train_spec as forge_train_spec
 from forge.cli.config import parse
 from forge.data.collate import collate_packed
 from forge.data.datasets.packed import PackedDataset, TextPacker
-from forge.data.datasets.sft_dataset import sft_iterable_dataset
+from forge.data.datasets.sft_dataset import AlpacaToMessages, sft_iterable_dataset
 from forge.data.tokenizer import HuggingFaceModelTokenizer
 
 from omegaconf import DictConfig, OmegaConf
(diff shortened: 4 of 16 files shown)


src/forge/controller/proc_mesh.py
```diff
@@ -0,0 +1,87 @@
...
+        AllocSpec,
+    )
+    from monarch._src.actor.meta.allocator import MastAllocator, MastAllocatorConfig
+    from monarch.tools.components.meta import hyperactor
+
+    MAST_SUPPORTED = True
+except ImportError:
+    logger.warning(
+        "MAST is not supported on this platform. You can ignore this if you do not work at Meta."
+    )
```
ebsmothers, line 37: I hope this is only temporary?

joecummings:
> Logging doesn't work by default in Monarch Actors

... Can we make it work in Monarch Actors? Or is it intentional that logging doesn't always work? If intentional, why?

src/forge/controller/proc_mesh.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+async def get_proc_mesh(scheduler_config: DictConfig) -> ProcMesh:
+    if scheduler_config.scheduler == "local":
+        if scheduler_config.num_hosts != 1:
+            raise ValueError("Local scheduler only supports 1 host")
+        return await proc_mesh(gpus=scheduler_config.num_gpus)
+    elif scheduler_config.scheduler == "mast":
```
ebsmothers, line 45: OK tell me if I'm crazy but why are we putting all this stuff on Github? Imo it'd be better to just put somewhere inside fbcode.. we are already plugging in there for the MAST launch script, right?

src/forge/controller/actor.py
```diff
@@ -0,0 +1,36 @@
...
+        formatter = logging.Formatter(
+            f"{BLUE}[{self.__class__.__name__}-{self._rank}/{self._size}] %(asctime)s %(levelname)s{RESET} %(message)s",
+            datefmt="%Y-%m-%d %H:%M:%S",
+        )
+        stdout_handler = logging.StreamHandler(sys.stdout)
+        stdout_handler.setLevel(logging.INFO)
+        stdout_handler.setFormatter(formatter)
+
+        self.logger.root.setLevel(logging.INFO)
+        self.logger.root.addHandler(stdout_handler)
```
ebsmothers, line 35: Similar question around logging not working by default in Monarch Actors.. is this by design? Or is it something we should expect to change down the line

allenwang28:
> ... Can we make it work in Monarch Actors? Or is it intentional that logging doesn't always work? If intentional, why?

I think we could, but for velocity I would like to start with controlling it in Forge and upstreaming it once we're happy with the implementation.

src/forge/controller/proc_mesh.py
allenwang28: So it's a bit tricky - Monarch won't run like SPMD workloads. The way it works is by spinning up Monarch listeners in place of your old `torchrun` workloads, and having something on the client side connect to those listeners. In some way, you'll need some type of MAST specificity.

I agree with the general sentiment that we won't want to expose MAST in the final iteration of forge, and that we should start here and iterate as the story evolves.

ebsmothers: Noob question about this: what is the difference between this approach and the one in D77302505? Iiuc that diff is still using torchx to kick off a MAST job, but the job only constructs a Monarch hyperactor which can then be leveraged by a controller defined elsewhere to run training actors. So in that sense I guess your version is a more self-contained implementation. But would it be possible to more explicitly separate your `get_proc_mesh` from your dummy Actor implementation and main loop in a way that hides this MAST-specific logic (or at least makes the client code more agnostic to it)? Maybe I'm barking up the wrong tree here (if so let me know), still getting used to this model of doing things.

allenwang28: No worries at all on the questions lol, I want to land this right as well!

Yeah, the APIs I'm using here is the vision for how we want users to use Monarch - it pulls torchx into Monarch and runs the same command that your linked diff also contains. If we were to follow that diff, we would still need MAST specific logic to tell Monarch which MAST job to connect to.

What I want to point out is that the dummy Actor here is runnable on a proc_mesh spawned locally or on MAST - proc_mesh itself is the abstraction for the scheduling layer. As we get closer to running Monarch on OSS, I expect the abstraction will become more generalizable and changeable by config

PR #35, opened 2025-08-01
Replace checkpoint save/load with proper APIs

DNXie:
This PR updates the checkpointing behavior as follows:

- Replace `dcp_save` and `dcp_load` with `save` and `load`.
- Intermediate checkpoints are now saved in the Titan format.
- The final checkpoint is saved in Hugging Face (HF) format if the `last_save_in_hf` option is enabled in the YAML configuration.
- Note that the folder specified by `initial_load_path` must not already exist to avoid loading issues.

**Test Plan:**  
- Verified saving and loading of intermediate and final checkpoints in both Titan and HF formats.

### DIFF
--- a/apps/sft/main.py
+++ b/apps/sft/main.py
@@ -58,6 +58,7 @@ class ForgeSFTRecipe(ForgeEngine):
 
     def __init__(self, job_config: ForgeJobConfig):
         self.current_step = 0
+        self.num_training_steps = job_config.training.steps
         self.gradient_accumulation_steps = 1  # Example value, adjust as needed
         super().__init__(job_config)
         self.metric_logger = None  # TODO: fix this


apps/sft/main.py
```diff
@@ -215,18 +212,10 @@ def train(self) -> None:
...
-            ):
-                self.checkpointer.dcp_save(
-                    ModelWrapper(self.model_parts).state_dict(),
-                    checkpoint_id=self.checkpointer._create_checkpoint_id(
-                        self.current_step
-                    ),
-                    async_mode=self.checkpointer.async_mode,
-                    to_hf=True,
-                )
+            self.checkpointer.save(
```
ebsmothers, line 215: Don't we want to keep the check for `self.checkpointer._should_save` around? Iiuc this will only save on the last step, right?

apps/sft/main.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, job_config: ForgeJobConfig):
         self.num_training_steps = 1000  # Example value, adjust as needed
```
ebsmothers: This is my fault, but we should remove this hardcoding. Otherwise if we set e.g. `training.steps=10` we will error out on step 11

apps/sft/main.py
DNXie: This is implemented in the function [save](https://github.com/pytorch/torchtitan/blob/ed288bc9f28700b992cb7e50465648cc21aced28/torchtitan/components/checkpoint.py#L466)

apps/sft/main.py
DNXie: @ebsmothers I have removed the hardcoded training steps.

apps/sft/main.py
```diff
@@ -58,9 +58,9 @@ class ForgeSFTRecipe(ForgeEngine):
 
     def __init__(self, job_config: ForgeJobConfig):
         self.current_step = 0
-        self.num_training_steps = 1000  # Example value, adjust as needed
```
DNXie, line 61: @ebsmothers  I have removed this hard coded training steps.

ebsmothers: Is this defined somewhere else then? I checked in the titan engine and I don't see it there. If not we should define it ourselves (can just pull from the relevant config field). We use it during the training loop so if it's not defined we'll hit an error

apps/sft/main.py
```diff
@@ -58,9 +58,10 @@ class ForgeSFTRecipe(ForgeEngine):
...
+        self.num_training_steps = job_config.training.steps
```
DNXie, line 61: @ebsmothers add the `self.num_training_steps` init back. Read from config file.

PR #36, opened 2025-08-01
Fix OSS us Build

pbontrager:
- Add vllm requirement
	- Add torch nightly override for vllm	 
- Pin to working Monarch-nightly version
	- pin to python <= 3.12 (Monarch doesn't support 3.13)
- fix torch install for cpu issue
- [TODO] get Monarch install to work with uv (can't find static library libcudard12

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,9 +14,10 @@ dependencies = [
     "torchdata>=0.8.0",
     "torchtitan",
     "torch",
-    "torchmonarch-nightly==2025.7.26",
+    "torchmonarch-nightly==2025.8.1",
     # vLLM
-    "vllm",
+    # TODO: pin specific vllm version
+    #"vllm==0.10.0",
     # Hugging Face integrations
     "datasets>=2.21.0",
     "tokenizers",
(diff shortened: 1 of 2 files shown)


Jack-Khuu:
fyi; This unexpectedly grabs an old version of torch as-is `(+ torch==2.8.0.dev20250627+cpu)` which causes mismatches with forge code

uv.lock
allenwang28, line 1: is this supposed to be gitignored?

pyproject.toml
```diff
@@ -14,7 +14,9 @@ dependencies = [
     "torchdata>=0.8.0",
     "torchtitan",
     "torch",
-    "torchmonarch-nightly",
+    "torchmonarch-nightly==2025.7.26",
+    # vLLM
+    "vllm",
```
allenwang28: I feel like this should be versioned, no?

uv.lock
joecummings: Okay so technically [it IS supposed to be added](https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile) to ensure that runs are consistent across builds and platforms. HOWEVER, we are doing something rather unusual where we commonly want to pick up the newest nightly builds. Here are the options:
1. Do not commit uv.lock, which will always build a new version (with new nightlies) when users run `uv sync`. 
2. Cron job to update uv.lock every day. 


Option 1 seems best and least intrusive IMO. When we DO have a stable release, we should have pins to stable torchtitan and stable monarch and stable vllm with an option for nightlies. But right now, there's nothing that would work with a "stable" version of our upstream libraries :)

PR #37, opened 2025-08-05
[No longer for land] Data types and examples in ToyRL app

joecummings:


src/forge/types.py
```diff
@@ -71,6 +75,24 @@ def __post_init__(self):
...
 
 
+class Message(TypedDict):
+    role: Role
+    content: str | dict[str, Any]
+    tools: dict[str, Any] | None
+
+
+@dataclass
+class Conversation:
```
ebsmothers: Is this schema based on something in particular?

src/forge/actors/__init__.py
```diff
@@ -3,8 +3,3 @@
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
-
-from .collector import Collector
-from .policy import Policy, PolicyRouter
-
-__all__ = ["Collector", "Policy", "PolicyRouter"]
```
pbontrager, line 10: Why are we deleting this?

PR #38, opened 2025-08-06
Remove torch/monarch from dep to enable MAST install

Jack-Khuu:
There are a few ways to get a dev workflow working, with MAST being one of the most updated ones.
This PR consciously break OSS install support in favor of enabling the MAST flow.

See `forge env setup: MAST enviroment` for more details


Jack-Khuu:
@pbontrager Thoughts? Since I think this will impact your flow

PR #39, opened 2025-08-06
Fix tests (test_toy_rl, TestReplayBuffer) broken by Trajectory and ReplayBuffer changes

Jack-Khuu:
Misc changes to get tests working again
- Fixed broken `Trajectory` and `SimpleReplayBuffer` import signature
- Add missing test deps
- Removed redundant `TestSimpleReplayBuffer` test which is covered by `TestReplayBuffer`

---
```
pytest tests/test_replay_buffer.py::TestReplayBuffer
pytest tests/unit_tests/rl/test_toy_rl.py
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,10 +40,6 @@ dev = [
     "tomli>=1.1.0",
     "anyio",
     "pytest-asyncio",
-    "pytest-tornasync",
-    "pytest-trio",
-    "pytest-twisted",
-    "twisted",
 ]
 
 [project.scripts]
--- a/src/forge/actors/collector.py
+++ b/src/forge/actors/collector.py
@@ -44,7 +44,8 @@ class Collector(Actor):
         state = self.environment.reset()
 
         # Initialize trajectory storage
-        trajectory = Trajectory()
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -38,6 +38,12 @@ dev = [
     "pytest-cov",
     "tensorboard",
     "tomli>=1.1.0",
+    "anyio",
+    "pytest-asyncio",
```
joecummings, line 42: I think this is the only one we need

Jack-Khuu: The others are used by test_toy_rl i think

```
async def functions are not natively supported.
You need to install a suitable plugin for your async framework, for example:
  - anyio
  - pytest-asyncio
  - pytest-tornasync
  - pytest-trio
  - pytest-twisted
```

But I'll give it another look

joecummings: Yeah you can just install one of them (not all), so for the tests I opted for `pytest-asyncio`

Jack-Khuu: Whoops reading is hard

Unrelated: Looks like the tests were passing when they shouldn't be; setting PR back to draft while i look

PR #40, opened 2025-08-07
RLTrainer

allenwang28:
This implements the RLTrainer actor following the pattern from apps/sft_v2. Currently this PR works with setup and implements train_step. I'm putting up this PR to land the trainer actor and unblock weight sync work. This is still missing RLLoss which is blocked on a update in ForgeEngine. After this lands and the ForgeEngine update, this will be integrated into apps/grpo (pending #58) and apps/rl will be removed.

### DIFF
--- a/apps/rl/main.py
+++ b/apps/rl/main.py
@@ -6,7 +6,7 @@
 """A working example showcasing a practical example of forge with RL.
 
 Run this with:
-    HF_HUB_DISABLE_XET=1 python -m apps.rl.main --config apps/rl/llama3_8b.yaml
+    python -m apps.rl.main --config apps/rl/llama3_8b.yaml
 
 """
 
@@ -14,7 +14,6 @@ import asyncio
 import logging
 import sys
 
-# from forge.actors import Policy, PolicyRouter, RLTrainer, ReplayBuffer
 from forge.actors import ReplayBuffer, RLTrainer
 
 from forge.cli.config import parse
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,10 +47,10 @@ oss = [
     "torchmonarch-nightly==2025.8.1",
     "torchstore",
 ]
-# cpu = [
-#     "torch==2.9.0.dev20250826",
-#     "monarch-no-torch==0.1.0.dev20250826",
-# ]
+cpu = [
+    "torch==2.9.0.dev20250826",
+    "monarch-no-torch==0.1.0.dev20250826",
+]
 
 
 [project.scripts]
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -54,7 +54,6 @@ class RLTrainer(ForgeActor, ForgeEngine):
     tokenizer: Tokenizer
     train_dataloader: Dataloader
     # val_dataloader: Dataloader
-    metric_logger: MetricLogger
     profiler: Profiler
     device: torch.device
     step: int
@@ -66,7 +65,6 @@ class RLTrainer(ForgeActor, ForgeEngine):
 
         self.current_step = 0
         self.num_training_steps = job_config.training.steps
-        self.metric_logger = None
         self.gradient_accumulation_steps = 1  # Example value, adjust as needed
         self._rank = current_rank().rank
         self._size = math.prod(current_size().values())
@@ -121,6 +119,13 @@ class RLTrainer(ForgeActor, ForgeEngine):
         # apply context parallelism if cp is enabled
         # ensure CP handles the separate freqs_cis buffer for each pp stage
         inputs = input_dict["tokens"]
+
+        if getattr(self.model_args, "use_flex_attn", False):
+            cp_mesh = (
+                parallel_dims.world_mesh["cp"] if parallel_dims.cp_enabled else None
+            )
+            init_attention_mask(inputs, self.tokenizer.base_tokenizer.eos_id, cp_mesh)
+
         optional_context_parallel_ctx = (
             dist_utils.create_context_parallel_ctx(
                 cp_mesh=parallel_dims.world_mesh["cp"],
@@ -176,7 +181,6 @@ class RLTrainer(ForgeActor, ForgeEngine):
         for k, v in batch.items():
             if isinstance(v, torch.Tensor):
                 batch[k] = v.to("cuda")  # TODO: hardcoded for now
-        self.train_step(batch)
 
         # TODO implement gradient accumulation
         # with GradientAccumulation(
@@ -213,8 +217,6 @@ class RLTrainer(ForgeActor, ForgeEngine):
         # self.pbar.close()
         if self.checkpointer:
             self.checkpointer.close()
-        if self.metric_logger:
-            self.metric_logger.close()
 
     def __repr__(self) -> str:
         return "Trainer"


src/forge/controller/proc_mesh.py
```diff
@@ -92,8 +103,24 @@ async def get_proc_mesh(process_config: ProcessConfig) -> ProcMesh:
...
-        p = await ProcMesh.from_alloc(alloc)
+        if env:
+
+            def setup():  # noqa: FB811
+                for k, v in env.items():
+                    os.environ[k] = v
+
+            p = await ProcMesh.from_alloc(alloc, setup=setup)
+        else:
+            p = await ProcMesh.from_alloc(alloc)
```
Jack-Khuu, line 114: nit: No strong pref

```suggestion
        def setup():  # noqa: FB811
            if env:
                for k, v in env.items():
                    os.environ[k] = v
        p = await ProcMesh.from_alloc(alloc, setup=setup)
```

src/forge/actors/policy.py
```diff
@@ -13,15 +13,7 @@
 from typing import Dict, List
 
 import torch
-from forge.controller import spawn_actors
-from forge.controller.service import ServiceConfig
-from forge.controller.spawn import spawn_service
```
Jack-Khuu, line 18: I blame my linter for missing this

Thanks

apps/rl/main.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+    print("Setup done")
+
+    print("shutting down...")
+    await asyncio.gather(*[a.mesh.stop() for a in [trainer]])
+
+
+@parse
+def recipe_main(cfg: DictConfig) -> None:
+    asyncio.run(run(cfg))
```
Jack-Khuu, line 59: Is the indirection to `run` because parse doesn't play well with async? vs @parse on `run`

apps/rl/main.py
```diff
@@ -0,0 +1,64 @@
...
+from omegaconf import DictConfig
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+async def run(cfg: DictConfig):
+    trainer, buffer = await asyncio.gather(
+        spawn_actors(
```
Jack-Khuu, line 30: Doesn't matter for the sake of the apps, but why not spawn_services?

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+    parallel_dims: ParallelDims
+    model: list[nn.Module]
+    loss_fn: LossFunction
+    optimizer: OptimizersContainer
+    lr_scheduler: LRSchedulersContainer
+    checkpointer: Checkpointer
+    tokenizer: Tokenizer
+    train_dataloader: Dataloader
+    # val_dataloader: Dataloader
+    metric_logger: MetricLogger
```
Jack-Khuu:
```suggestion
    metric_logger: MetricLogger | None
```

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+                    pred = model_parts[0](inputs)
+                    loss = self.loss_fn(pred, labels)
+                # need to free to before bwd to avoid peaking memory
+                del pred
+                loss.backward()
+
+        return loss
+
+    @endpoint
+    def train_step(self, batch) -> None:
```
Jack-Khuu, line 179: Not in this PR since things are still moving, but let's type batch

apps/rl/llama3_8b.yaml
```diff
@@ -0,0 +1,138 @@
...
+replay_buffer:
+  batch_size: 2
+  max_policy_age: 2
+  seed: None
+  processes:
+    scheduler: local # local | mast (not supported yet)
+    num_hosts: 1
+    num_procs: 1
+
+# policy:
```
ebsmothers, line 70: Remove from here on down?

apps/rl/main.py
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+"""A working example showcasing a practical example of forge with RL.
+
+Run this with:
+    HF_HUB_DISABLE_XET=1 python -m apps.rl.main --config apps/rl/llama3_8b.yaml
```
ebsmothers: Why do we need HF_HUB_DISABLE_XET here, we aren't downloading anything?

apps/rl/main.py
```diff
@@ -0,0 +1,64 @@
...
+
+"""
+
+import asyncio
+import logging
+import sys
+
+# from forge.actors import Policy, PolicyRouter, RLTrainer, ReplayBuffer
```
ebsmothers: remove

src/forge/actors/__init__.py
```diff
@@ -4,6 +4,25 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from .collector import Collector
+__all__ = ["Policy", "PolicyRouter", "RLTrainer", "ReplayBuffer"]
 
-__all__ = ["Collector"]
+
+def __getattr__(name):
```
ebsmothers, line 10: You trying to sneak this in or what? 👀

pyproject.toml
```diff
@@ -47,10 +47,10 @@ oss = [
     "torchmonarch-nightly==2025.8.1",
     "torchstore",
 ]
-cpu = [
-    "torch==2.9.0.dev20250826",
-    "monarch-no-torch==0.1.0.dev20250826",
-]
+# cpu = [
```
ebsmothers: What's this about?

apps/rl/llama3_8b.yaml
```diff
@@ -0,0 +1,138 @@
+# Config for GRPO finetuning using a Llama3.1 8B Instruct model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   export HF_HUB_DISABLE_XET=1
```
ebsmothers, line 5: super nit but for OSS I would leave this out. Don't subject others to the idiosyncracies of our infrastructure

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+        # Move tensors to the appropriate device
+        for k, v in batch.items():
+            if isinstance(v, torch.Tensor):
+                batch[k] = v.to("cuda")  # TODO: hardcoded for now
+        self.train_step(batch)
```
joecummings: Maybe I'm misreading my indents, but train_step calls train_step?

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+
+        # if self.current_step % self.train_config.val_every_n_steps == 0:
+        #     self.validate()
+        self.checkpointer.save(
+            curr_step=self.current_step,
+            last_step=self.current_step == self.num_training_steps,
+        )
+
+    @endpoint
+    def push_weights(self) -> None:
```
joecummings, line 212: nit: slightly more descriptive name here.

apps/sft_v2/main.py
```diff
@@ -71,7 +70,11 @@ class ForgeSFTRecipe(ForgeActor, ForgeEngine):
     device: torch.device
     step: int
 
-    def __init__(self, job_config: ForgeJobConfig):
+    def __init__(self, config: DictConfig):
+        job_config = ForgeJobConfig().to_dict()
+        # Hack to deal with literal types from titan
```
ebsmothers, line 75: nit but the hack is on the to_dict(), right? (Also fine to just remove the comment completely)

src/forge/controller/proc_mesh.py
```diff
@@ -40,20 +41,30 @@
...
-    mesh = await get_proc_mesh(processes)
-    actors = await mesh.spawn(name, actor_cls, cfg)
+    mesh = await get_proc_mesh(processes, set_address)
+    actors = await mesh.spawn(name, actor_cls, **cfg)
     actors.mesh = mesh
     return actors
 
 
-async def get_proc_mesh(process_config: ProcessConfig) -> ProcMesh:
+async def get_proc_mesh(process_config: ProcessConfig, set_address=False) -> ProcMesh:
```
joecummings, line 57: This does the automatic GPU alloc right? Is there a reference from Ray on how they do this?

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+        self.current_step = 0
+        self.num_training_steps = job_config.training.steps
+        self.metric_logger = None
+        self.gradient_accumulation_steps = 1  # Example value, adjust as needed
+        self._rank = current_rank().rank
+        self._size = math.prod(current_size().values())
+        self._init_dist()
+        super().__init__(job_config)
+
+    def _init_dist(self):
```
joecummings, line 74: Should probably go inside Monarch or a general helper function at some point

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+Dataloader = Any
+MetricLogger = Any
+Profiler = Any
+Tokenizer = Any
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class RLTrainer(ForgeActor, ForgeEngine):
```
ebsmothers, line 45: Is this the end state we anticipate? (Basically two different classes: one for the full SFT recipe run with Monarch and one with a very similar class for the RL trainer)

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+            "ROLE_WORLD_SIZE": str(self._size),
+            "ROLE_NAME": "rank",
+            "WORLD_SIZE": str(self._size),
+            "PYTORCH_CUDA_ALLOC_CONF": "expandable_segments:True",
+        }
+        os.environ.update(env)
+        logger.info("env: {}".format(env))
+
+    @endpoint
+    async def setup(self):
```
ebsmothers, line 100: Does this work right now? I thought you said we needed to factor out ForgeEngine's `__init__` into a deferred setup method for things to work properly

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def forward_backward(
+        self, input_dict: dict[str, torch.Tensor], labels: torch.Tensor
+    ) -> torch.Tensor:
+        model_parts = self.model_parts
+        parallel_dims = self.parallel_dims
+
+        # apply context parallelism if cp is enabled
+        # ensure CP handles the separate freqs_cis buffer for each pp stage
+        inputs = input_dict["tokens"]
```
ebsmothers, line 121: Fyi for any models with flex you will need to add the changes from #76

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+        #     self.data_parallel_size,
+        # ) as grad_acc:
+        # TODO: convert to GRPO Loss
+        labels = batch.pop("labels")
+        loss = self.forward_backward(batch, labels)
+        # self.pbar.update(1)
+        # self.pbar.set_description(f"{self.current_step}|Loss: {loss}")
+
+        self.optimizers.step()
+        self.optimizers.zero_grad()
```
ebsmothers, line 198: 👀

src/forge/actors/trainer.py
```diff
@@ -0,0 +1,220 @@
...
+                # need to free to before bwd to avoid peaking memory
+                del pred
+                loss.backward()
+
+        return loss
+
+    @endpoint
+    def train_step(self, batch) -> None:
+        # Move tensors to the appropriate device
+        for k, v in batch.items():
```
ebsmothers, line 181: nit: we have your utility [batch_to_device](https://github.com/meta-pytorch/forge/blob/main/src/forge/data/utils.py#L191) now

src/forge/actors/trainer.py
ebsmothers, line 212: nit: add a comment explaining what this will be used for

src/forge/controller/proc_mesh.py
```diff
@@ -40,20 +41,30 @@
 
 
 async def spawn_actors(
-    name: str, actor_cls: ForgeActor, cfg: DictConfig, processes: ProcessConfig
+    name: str,
+    actor_cls: ForgeActor,
+    cfg: DictConfig,
+    processes: ProcessConfig,
+    set_address: bool = False,
```
ebsmothers, line 48: Why/when do we actually need this? (I saw you use it in the spawn for ForgeSFTRecipe, but why there and nowhere else?)

src/forge/actors/__init__.py
joecummings: NO

src/forge/controller/proc_mesh.py
pbontrager: I wanted to pass None instead of a noop function in case it's faster, and the linter didn't let me conditionally define setup as None or a function.

apps/rl/main.py
pbontrager: We should probably not have run run but I left it this way for the reason you said

apps/rl/main.py
pbontrager: I had this before spawn_services and the next step is integration with grpo so it wasn't worth updating

src/forge/actors/trainer.py
pbontrager: Forgot to remove that line

src/forge/actors/trainer.py
pbontrager: This is brought over from sft_v2. There is a monarch helper that landed yesterday, but in general we're waiting for hostmesh support so we can handle this more robustly.

src/forge/controller/proc_mesh.py
pbontrager: This is just adding distributed variables that need to be shared across the proc_mesh. This should also be handled by the host_mesh setup in the future.

pyproject.toml
pbontrager: Removing, I forgot it there. It's broken with that addition but it needs to be fixed separate

apps/rl/llama3_8b.yaml
pbontrager: Currently leaving comments to make our internal use as fast as possible and will remove before oss

src/forge/actors/trainer.py
pbontrager: I don't think it's avoidable. They're very similar now but they'll diverge further before it's fully done.

src/forge/actors/trainer.py
pbontrager: It "works" it just isn't as fast as it should be since ForgeEngine is doing all it's work during spawn. When ForgeEngine is updated, I'll update this setup to call parent setup

src/forge/controller/proc_mesh.py
pbontrager: Changes will propagate, but you need a proper way to set a unique communication address per proc_mesh. This isn't at the actor level. This will be used for any mesh that uses distributed apis, including policy

PR #46, opened 2025-08-13
Add Structured Output Support to Policy

Jack-Khuu:
Add structured output (aka guided decoding) example to policy

---
` python src/forge/actors/policy.py` (With structured output toggled)

```
User: What is 3+5?
Assistant: Positive
```

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -9,6 +9,7 @@ import logging
 import os
 import sys
 from dataclasses import dataclass
+from typing import Dict
 
 import torch
 from monarch.actor import Actor, current_rank, endpoint, proc_mesh
@@ -18,7 +19,7 @@ from vllm.entrypoints.utils import _validate_truncation_size
 from vllm.executor.multiproc_worker_utils import set_multiprocessing_worker_envs
 from vllm.inputs import TextPrompt, TokensPrompt
 from vllm.lora.request import LoRARequest
-from vllm.sampling_params import RequestOutputKind, SamplingParams
+from vllm.sampling_params import GuidedDecodingParams, RequestOutputKind, SamplingParams
 from vllm.transformers_utils.tokenizer_group import init_tokenizer_from_configs
 from vllm.usage.usage_lib import UsageContext
 from vllm.utils import get_distributed_init_method, get_loopback_ip, get_open_port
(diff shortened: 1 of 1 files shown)


src/forge/actors/policy.py
```diff
@@ -320,7 +325,22 @@ async def _test(config):
     )
 
     policy_actor = await policy_mesh.spawn("policy", Policy, **config)
-    router = await router_mesh.spawn("policy_router", PolicyRouter, policy=policy_actor)
+
+    sampling_params = None
+    if guided_decoding:
+        # Add config for structured output
+        vllm_args = await policy_actor.get_vllm_args.choose()
```
ebsmothers, line 332: noob q: why are we running .choose here? It seems to me the more idiomatic thing to do here would be .call but maybe I'm missing the point (also I assume calling all actors is not a bottleneck here anyways as this just returns an instance variable)

pbontrager: You use choose when you just want one of the actors to do something. Basically like "if rank == 0: do"

src/forge/actors/policy.py
```diff
@@ -329,7 +349,7 @@ async def _test(config):
     router.run.call()
     print("Model running")
 
-    prompt = "Tell me a joke"
+    prompt = "What is 3+5?" if guided_decoding else "Tell me a joke"
```
pbontrager, line 352: Can you spin these things off as proper tests at some point? @ebsmothers is running something like this too big for our current test setup?

src/forge/actors/policy.py
```diff
@@ -320,7 +325,22 @@ async def _test(config):
     )
 
     policy_actor = await policy_mesh.spawn("policy", Policy, **config)
-    router = await router_mesh.spawn("policy_router", PolicyRouter, policy=policy_actor)
+
+    sampling_params = None
+    if guided_decoding:
```
pbontrager, line 330: We need to think of how we want to do this from the config in the future. Maybe some sampling param builders that the user can call?

src/forge/actors/policy.py
Jack-Khuu: From my understanding `call` also [triggers the entire Mesh](https://github.com/meta-pytorch/monarch/blob/8c62beccfb60da24538f58e8b3a27fa1f77014f6/python/monarch/_src/actor/endpoint.py#L134-L154), while like @pbontrager mentioned `choose` and `call_one` are more direct (or in the case of no arg it just picks a random)

src/forge/actors/policy.py
Jack-Khuu: We can throw them on the H100s and it should be fine once we get things more fleshed out

src/forge/actors/policy.py
Jack-Khuu: I'm kinda down to just lean into the vllm convention to start (Read from generation.json to populate, then let users override the kwargs)

This is probably something we ask for upstream

PR #51, opened 2025-08-13
Adds service implementation

allenwang28:
This PR adds a "service" - an orchestration system for managing multiple Proc/ActorMeshes. It adds automatic scaling, fault tolerance and load balancing.

There is a markdown file that describes service.md in detail.

### DIFF
--- a/src/forge/controller/service.py
+++ b/src/forge/controller/service.py
@@ -199,6 +199,9 @@ class ServiceMetrics:
 class AutoscalingConfig:
     """Configuration for autoscaling behavior."""
 
+    # Autoscaling control
+    enabled: bool = False  # Whether autoscaling is enabled (disabled by default)
+
     # Scale up thresholds
     scale_up_queue_depth_threshold: float = (
         5.0  # Average queue depth to trigger scale up
@@ -881,6 +884,10 @@ class Service:
         """Executes autoscaling decisions based on current metrics."""
         # Skip autoscaling if shutdown is requested
         if self._shutdown_requested:
+            return
+
+        # Skip autoscaling if disabled
+        if not self._cfg.autoscaling.enabled:
             return
 
         # Check scale up first (higher priority)
--- a/tests/test_service.py
+++ b/tests/test_service.py
@@ -205,6 +205,7 @@ async def test_replica_failure_and_recovery():
 async def test_autoscaling_scale_up():
     """Test automatic scale up under high load."""
     autoscaling_cfg = AutoscalingConfig(
+        enabled=True,
         scale_up_capacity_threshold=0.5,
         scale_up_queue_depth_threshold=2.0,
         scale_up_cooldown=0.5,
@@ -253,6 +254,7 @@ async def test_autoscaling_scale_up():
 async def test_autoscaling_scale_down():
     """Test automatic scale down when idle."""
     autoscaling_cfg = AutoscalingConfig(
+        enabled=True,
         scale_down_capacity_threshold=0.2,
         scale_down_queue_depth_threshold=0.5,
         scale_down_idle_time_threshold=3.0,
@@ -303,6 +305,7 @@ async def test_autoscaling_scale_down():
 async def test_autoscaling_limits():
     """Test that autoscaling respects min/max limits."""
     autoscaling_cfg = AutoscalingConfig(
+        enabled=True,
         scale_up_queue_depth_threshold=1.0,
         scale_down_capacity_threshold=0.9,  # High threshold to prevent scale down
     )


src/forge/controller/service.py
```diff
@@ -0,0 +1,1205 @@
...
+        ...         scale_down_capacity_threshold=0.3
+        ...     )
+        ... )
+        >>> service = Service(config, MyActorClass, model_path="/path/to/model")
+        >>> await service.__initialize__()
+
+        Session-based usage:
+
+        >>> async with service.session() as session:
+        ...     result1 = await service.my_endpoint(arg1, arg2)
```
joecummings, line 340: `session.my_endpoint(arg1, arg2)`?

src/forge/controller/recoverable_mesh.py
```diff
@@ -0,0 +1,283 @@
...
+        >>> elif mesh.failed:
+        ...     # Mesh needs recovery
+        ...     await mesh.spawn(setup_actor)  # Triggers recovery
+    """
+
+    def __init__(
+        self,
+        num_gpus: int,
+    ) -> None:
+        self.num_gpus = num_gpus
```
pbontrager: Should we change this to num_processes since it's not tied to GPUs?

tests/test_service.py
```diff
@@ -0,0 +1,519 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Tests for service.py
```
ebsmothers, line 8: i love unit tests

tests/test_service.py
```diff
@@ -0,0 +1,519 @@
...
+import pytest
+from forge.controller.service import AutoscalingConfig, ServiceConfig
+from forge.controller.spawn import spawn_service
+from monarch.actor import Actor, endpoint
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class Counter(Actor):
```
ebsmothers, line 23: doesn't matter much for now, but would be nice to have a simple actor like this made accessible more generally (e.g. in test utils) for all of our various actor-adjacent functionality

src/forge/controller/recoverable_mesh.py
```diff
@@ -0,0 +1,283 @@
...
+        Healthy: Mesh is operational and ready to handle requests
+        Recovering: Mesh is in the process of recovering from a failure
+        Unhealthy: Mesh has failed and needs recovery
+        Stopped: Mesh has been explicitly stopped and cannot be used
+    """
+
+    Healthy = 0
+    Recovering = 1
+    Unhealthy = 2
+    Stopped = 3
```
ebsmothers: nit: for enum use uppercase

src/forge/controller/recoverable_mesh.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+class RecoverableProcMesh(MeshTrait):
```
ebsmothers, line 72: High-level q: is there a world that we would ever put this in Monarch? Maybe it's not quite primitive enough, but feels a bit weird that we are wrapping a bunch of stuff like `proc_mesh._labels` (that at least I am not familiar with)

src/forge/controller/service.py
```diff
@@ -0,0 +1,1205 @@
...
+
+    def get_sessions_per_replica(self) -> float:
+        """Get average sessions per healthy replica."""
+        if self.healthy_replicas == 0:
+            return 0.0
+        return self.total_sessions / self.healthy_replicas
+
+
+@dataclass
+class AutoscalingConfig:
```
ebsmothers, line 199: OK admittedly I am an autoscaling noob, but how do we know that these are the right sets of fields to define our autoscaling? Iiuc it is just a bunch of heuristics, what's to say we don't want to use something like a PID controller instead? (Lmk if I'm totally missing the point here)

src/forge/controller/recoverable_mesh.py
allenwang28: yeah I expect this to be put into Monarch. Service may also become a Monarch level construct at some point!

src/forge/controller/recoverable_mesh.py
allenwang28: done

src/forge/controller/service.py
allenwang28: oops good catch, actually we're not supposed to have `as session` it currently just goes back through `service`

My question though is it more natural to be able to go through `as session`?

src/forge/controller/service.py
allenwang28: admittedly i'm not super well versed in autoscaling either, so a few of these fields (transparently generated by devmate) may be unnecessary

However my understanding is that min/max/default replicas are pretty standard, and that autoscaling is typically based on monitoring some type of metric (like queue depth) to make decisions to scale up or down.

If it helps we can simplify to just queue depth or something, but I was thinking we can start with a few knobs and change it as we learn more about the requirements during our prototyping

src/forge/controller/service.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Usage
+
+### Basic Service Setup
+
+```python
+from forge.controller.service import Service, ServiceConfig, AutoscalingConfig
+
+# Configure service parameters
+config = ServiceConfig(
```
pbontrager, line 67: Let's combine this with the ProcessConfig and try to unify all of our actors (Trainer, Policy, Envs, Rewards, Transforms) to be initiated as services. I want a consistent experience for assigning resources.

src/forge/controller/spawn.py
```diff
@@ -0,0 +1,42 @@
...
+
+from monarch.actor import Actor
+
+from forge.controller import Service, ServiceConfig
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+
+async def spawn_service(
```
pbontrager, line 19: I think we want (name:str, actor_cls: Actor, cfg: DictConfig, service_cfg: ServiceConfig)

src/forge/controller/service.py
```diff
@@ -0,0 +1,1205 @@
...
+logger.setLevel(logging.INFO)
+
+# Global context variable for session state
+_session_context: contextvars.ContextVar[dict | None] = contextvars.ContextVar(
+    "session_context", default=None
+)
+
+
+@dataclass
+class ReplicaMetrics:
```
pbontrager, line 65: TODO: tie this in to our Metric Logger to automatically log

src/forge/controller/service.py
```diff
@@ -0,0 +1,1205 @@
...
+    scale_down_step_size: int = 1  # How many replicas to remove at once
+
+    # Safety limits
+    max_queue_depth_emergency: float = 20.0  # Emergency scale up threshold
+    min_healthy_replicas_ratio: float = 0.5  # Minimum ratio of healthy replicas
+
+
+@dataclass
+class ServiceConfig:
+    gpus_per_replica: int
```
pbontrager: Let's change this to processes_per_replica and set defaults for min/max replicas to None so that autoscaling is off by default and doesn't need to be exposed to a new user.

src/forge/controller/service.md
allenwang28: leaving this as a follow up to merge with the main loop

src/forge/controller/recoverable_mesh.py
allenwang28: leaving this for when we integrate the custom proc_mesh API, I feel it could be more involved with MAST stuff

src/forge/controller/spawn.py
allenwang28: leaving for integration with apps

src/forge/controller/service.py
allenwang28: done

src/forge/controller/__init__.py
```diff
@@ -3,9 +3,21 @@
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
+from .actor import ForgeActor
+from .proc_mesh import get_proc_mesh, spawn_actors
+from .recoverable_mesh import RecoverableProcMesh
+from .service import AutoscalingConfig, Service, ServiceConfig
```
Jack-Khuu, line 9: nit: Why the switch to relative imports?

src/forge/controller/service.py
joecummings: hmmmm, if it is indeed boxed to session, it would make sense to call through there no?

However, if the APIs are the same on the service class, then that would be confusing.

Open to overall ideas on API strcuture?

src/forge/controller/service.md
```diff
@@ -0,0 +1,312 @@
...
+session_id = await service.start_session()
+result = await service.my_endpoint(session_id, arg1, arg2)
+await service.terminate_session(session_id)
+```
+
+### Stateless Calls
+
+```python
+# Direct calls without sessions (uses round-robin load balancing)
+result = await service.my_endpoint(arg1, arg2)
```
joecummings, line 104: This seems to contradict the above code. Can you do this without manually starting and terminating a session?

src/forge/controller/__init__.py
allenwang28: ah I noticed that all of the other folders were doing relative whereas controller was doing absolute. I'm okay either way

src/forge/controller/service.md
allenwang28: Yes, I'm basically trying to say that sessions are optional

PR #52, opened 2025-08-14
Adds initial CI suport

allenwang28:
This PR
- Cleans up unused tests
- Fixes the existing tests
- Adds in an initial CI yaml file (forked from Monarch)
- Since we added services, I am also removing stack because it keeps breaking with Monarch updates and is essentially obsolete...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,8 +12,6 @@ repos:
     -   id: check-merge-conflict
     -   id: no-commit-to-branch
         args: ['--branch=main']
-    # -   id: check-added-large-files
-    #     args: ['--maxkb=1000']
     -   id: end-of-file-fixer
         exclude: '^(.*\.svg)$'
 


src/forge/actors/__init__.py
```diff
@@ -5,6 +5,27 @@
...
-from .policy import Policy, PolicyRouter
 
-__all__ = ["Collector", "Policy", "PolicyRouter"]
+__all__ = ["Collector"]
+
+try:
+    from .policy import Policy, PolicyRouter
+
+    __all__.extend(["Policy", "PolicyRouter"])
+except ImportError as e:
```
joecummings: Devmate fix?

I'd prefer to iteratively expose things at the top level __init__ as they solidify, so we can just leave this blank for now and have people directly import Policy and PolicyRouter if needed assuming they have vllm installed.

allenwang28: done

scripts/common-setup.sh
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
joecummings: We shouldn't need this.

pyproject.toml
```diff
@@ -36,6 +37,7 @@ dev = [
     "pytest-cov",
     "tensorboard",
     "tomli>=1.1.0",
+    "tomli_w",
```
joecummings: I think this is a remnant of titan stuff?

.github/workflows/unit_test.yaml
```diff
@@ -0,0 +1,43 @@
...
+  run:
+    shell: bash -l -eo pipefail {0}
+
+jobs:
+  unit_tests:
+    if: github.repository_owner == 'pytorch'
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.9', '3.10', '3.11']
```
joecummings: 3.10 and over

.github/workflows/unit_test.yaml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          auto-update-conda: true
+          miniconda-version: "latest"
+          activate-environment: test
+          python-version: ${{ matrix.python-version }}
+      - name: Update pip
+        run: python -m pip install --upgrade pip
+      - name: Install dependencies
+        run: |
+          python -m pip install torch torchvision torchao
```
joecummings: Don't need torchvision

.github/workflows/unit_test.yaml
```diff
@@ -0,0 +1,42 @@
+name: Unit Test
```
ebsmothers, line 1: Does it run though (I don't see a signal in the CI yet)

pyproject.toml
```diff
@@ -13,6 +13,7 @@ dependencies = [
     # PyTorch
     "torchdata>=0.8.0",
     "torchtitan",
+    "torchao",
```
ebsmothers, line 16: Where are we using this? Also the way you set up this + unit_test.yaml I think we're installing it twice?

.github/workflows/unit_test.yaml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          auto-update-conda: true
+          miniconda-version: "latest"
+          activate-environment: test
+          python-version: ${{ matrix.python-version }}
+      - name: Update pip
+        run: python -m pip install --upgrade pip
+      - name: Install dependencies
+        run: |
+          python -m pip install torch torchao
```
ebsmothers: If we are gonna be on nightlies for titan, torch core, etc we may need to modify this at some point

.github/workflows/unit_test.yaml
allenwang28: works finally

pyproject.toml
allenwang28: removed

allenwang28:
> I'll come back to this and fix when higher priority. Right now I don't consider this a P0.
> 
> Blocking b/c there should be a signal in CI before landing, therefore something isn't working.

This PR now runs all of the tests on CPU! please take another look

pyproject.toml
allenwang28: actually - the torchao is because of vLLM

.pre-commit-config.yaml
```diff
@@ -12,8 +12,8 @@ repos:
     -   id: check-merge-conflict
     -   id: no-commit-to-branch
         args: ['--branch=main']
-    -   id: check-added-large-files
-        args: ['--maxkb=1000']
+    # -   id: check-added-large-files
```
joecummings: Just remove if not needed

assets/wheels/monarch_no_torch-0.1.0.dev20250815-py3-none-any.whl
joecummings, line 1: ?

assets/version.txt
joecummings, line 1: I don't think we need this.

assets/wheels/monarch_no_torch-0.1.0.dev20250815-py3-none-any.whl
allenwang28: doing this because `pip install torchnightly` doesn't work... I know the right thing is to fix that, but I hope you understand that will be difficult for a bit

assets/version.txt
allenwang28: this is mostly because the pyproject.toml uses it, similar to torchtitan does it

joecummings: But does it need it to run? IIUC, it should not and I'd prefer not to add until necessary.

assets/wheels/monarch_no_torch-0.1.0.dev20250815-py3-none-any.whl
joecummings: Got it - can you just refer to it in the pyproject.toml as e.g. `"mylib @ file:///${PROJECT_ROOT}/dist/mylib-0.1.0-py3-none-any.whl"`

PR #54, opened 2025-08-14
make dataset configurable and add validation loop

DNXie:
- Make dataset path and split configurable
- Add validation dataloader
- Add validation loop

Added `dataset_val` and `validation` sections in the config file:
```
validation:
  local_batch_size: 1
  freq: -1  # Change to a positive number to enable validation
  steps: 200  # Max steps to run validation. Validation disabled if negative.

dataset_val:
  path: yahma/alpaca-cleaned
  split: train[95%:]

```

**Test**:

```
uv run forge run --nproc_per_node 2 apps/sft/main.py --config apps/sft/llama3_8b.yaml
```

The validation freq (`freq`) is set to 2 (run validation every two steps):

```
1|Loss: 1.0664817094802856:   0%|          | 2/1000 [00:06<57:53,  3.48s/it]
Validation loss: 1.311207890510559
Running validation Loss: 1.3039: : 200it [00:24,  8.28it/s]
Validation loss: 1.311207890510559
3|Loss: 1.2771787643432617:   0%|          | 4/1000 [00:31<2:15:20,  8.15s/it]
Validation loss: 1.3083338737487793
Running validation Loss: 1.2985: : 200it [00:24,  8.25it/s]
Validation loss: 1.3083338737487793
5|Loss: 1.3110204935073853:   1%|          | 6/1000 [00:57<2:37:35,  9.51s/it]
Validation loss: 1.2877949476242065
Running validation Loss: 1.2748: : 199it [00:24,  8.32it/s]
Validation loss: 1.2877949476242065
7|Loss: 1.3947221040725708:   1%|          | 8/1000 [01:22<2:46:21, 10.06s/it]
Validation loss: 1.274566411972046
Running validation Loss: 1.2768: : 200it [00:24,  8.12it/s]
Validation loss: 1.274566411972046
9|Loss: 1.1953884363174438:   1%|          | 10/1000 [01:48<2:50:45, 10.35s/it]
Validation loss: 1.2020115852355957
Running validation Loss: 1.1951: : 200it [00:24,  8.17it/s]
Validation loss: 1.2020115852355957
11|Loss: 1.2392996549606323:   1%|          | 12/1000 [02:13<2:53:13, 10.52s/it]
Validation loss: 1.1668553352355957
Running validation Loss: 1.1603: : 200it [00:24,  8.12it/s]
Validation loss: 1.1668553352355957
13|Loss: 1.041933536529541:   1%|▏         | 14/1000 [02:39<2:54:35, 10.62s/it]]
Validation loss: 1.0708351135253906
Running validation Loss: 1.0622: : 200it [00:24,  8.22it/s]
Validation loss: 1.0708351135253906
15|Loss: 1.0107098817825317:   2%|▏         | 16/1000 [03:05<2:54:47, 10.66s/it]
```

Tested steps = 10 and 200 (exhaust the dataset).

### DIFF
--- a/apps/sft/main.py
+++ b/apps/sft/main.py
@@ -91,7 +91,7 @@ class ForgeSFTRecipe(ForgeEngine):
         # self.profiler = self.setup_profiler(self.train_config.profiler_config)
         # self.logger = self.setup_logger(self.train_config.logger_config)
 
-    def setup_data(self, dataset_config, batch_size, infinite=True):
+    def setup_data(self, dataset_config, batch_size):
         tokenizer = HuggingFaceModelTokenizer(
             tokenizer_json_path=os.path.join(
                 self.job_config.model.hf_assets_path, "tokenizer.json"
@@ -109,7 +109,6 @@ class ForgeSFTRecipe(ForgeEngine):
             message_transform=AlpacaToMessages(),
             path=dataset_config.path,
             split=dataset_config.split,
-            infinite=infinite,
         )
         packer = TextPacker(padding_idx=0)
         dataset = PackedDataset(
--- a/src/forge/data/datasets/hf_dataset.py
+++ b/src/forge/data/datasets/hf_dataset.py
@@ -75,7 +75,6 @@ class HfIterableDataset(InfiniteTuneIterableDataset):
         dataset_name: Optional[str] = None,
         filter_fn: Optional[Callable] = None,
         filter_kwargs: Optional[dict[str, Any]] = None,
-        infinite: bool = True,
         **load_dataset_kwargs,
     ):
         # Store configuration
@@ -114,7 +113,6 @@ class HfIterableDataset(InfiniteTuneIterableDataset):
         self._setup_hf_dataset(
             load_dataset_kwargs, num_shards_per_rank, filter_fn, filter_kwargs
         )
-        self.infinite = infinite
 
     @property
     def info(self) -> DatasetInfo:
@@ -254,9 +252,6 @@ class HfIterableDataset(InfiniteTuneIterableDataset):
 
                     samples_yielded += 1
                     yield sample
-
-                if not self.infinite:
-                    break
 
             except StopIteration:
                 # Expected when dataset is exhausted
--- a/src/forge/data/datasets/sft_dataset.py
+++ b/src/forge/data/datasets/sft_dataset.py
@@ -163,7 +163,6 @@ def sft_iterable_dataset(
     dataset_name: Optional[str] = None,
     filter_fn: Optional[Callable] = None,
     filter_kwargs: Optional[dict[str, Any]] = None,
-    infinite: bool = True,
     **load_dataset_kwargs: dict[str, Any],
 ) -> HfIterableDataset:
     """
@@ -208,6 +207,5 @@ def sft_iterable_dataset(
         dataset_name=dataset_name,
         filter_fn=filter_fn,
         filter_kwargs=filter_kwargs,
-        infinite=infinite,
         **load_dataset_kwargs,
     )


apps/sft/main.py
```diff
@@ -95,8 +104,8 @@ def setup_data(self):
         dataset = sft_iterable_dataset(
             model_transform=tokenizer,
             message_transform=AlpacaToMessages(),
-            path="yahma/alpaca-cleaned",
-            split="train",
+            path=dataset_config.path,
+            split=dataset_config.split,
```
ebsmothers, line 111: One thing we should think about is how to support additional args beyond those we've already hardcoded. E.g. in #50 we also need to pass `data_files`. (This is more of a config system question so it's OK to punt on it for now, but one path is to use something like instantiate for this, you can see [this section](https://docs.pytorch.org/torchtune/main/deep_dives/configs.html#configuring-components-using-instantiate) in the torchtune docs for an example)

apps/sft/main.py
```diff
@@ -214,13 +239,84 @@ def train(self) -> None:
...
+        total_val_tokens = torch.tensor(0.0, device=self.device)
+        with torch.no_grad():
+            val_pbar = tqdm(self.val_dataloader, desc="Validation", leave=False)
+            for batch_idx, batch in enumerate(val_pbar):
+                if batch_idx >= 100:  # TODO: remove this
+                    break
+                batch_to_device(batch, self.device)
+                current_num_tokens = batch[
+                    "labels"
+                ].numel()  # TODO: exclude ignore index
```
DNXie: @ebsmothers 
I try to do this like torchtune did
```
current_num_tokens = (
                    batch["labels"] != self._loss_fn.ignore_index
                ).sum()
```
but the current `self.loss_fn` doesn't have `ignore_index`

ebsmothers: Tbh I think this is something we kinda overparametrized in torchtune. You can just use the constant [CROSS_ENTROPY_IGNORE_IDX](https://github.com/meta-pytorch/forge/blob/d44bb20c47203bb9055e17692fa6fc7ea1996859/src/forge/data/utils.py#L10)

DNXie: Thanks! Fixed.

apps/sft/main.py
DNXie: I can support passing file paths. Which one (`data_files` or `path`) should it prioritize? For example, if user pass both `data_files` and `path`

apps/sft/main.py
```diff
@@ -214,13 +242,80 @@ def train(self) -> None:
...
+            batch_to_device(v, device)
+        elif isinstance(v, torch.Tensor):
+            batch[k] = v.to(device)
+        elif _SUPPORTS_FLEX_ATTENTION and isinstance(v, BlockMask):
+            batch[k] = v.to(device)
+        else:
+            raise ValueError(
+                f"""To use batch_to_device, all elements in the batch must be a dict or Tensor.
+Got key "{k}" with value of type {type(v)}"""
+            )
```
Jack-Khuu: nit: or BlockMask with flexattention enabled

DNXie: Thanks! Fixed.

apps/sft/main.py
```diff
@@ -214,13 +242,82 @@ def train(self) -> None:
...
+        print(f"\nValidation loss: {avg_val_loss}")
+
     def cleanup(self) -> None:
         if self.checkpointer:
             self.checkpointer.close()
         if self.metric_logger:
             self.metric_logger.close()
 
 
+def batch_to_device(batch: dict, device: torch.device) -> None:
```
ebsmothers: Let's move this out of main.py. Personally I would put in src/forge/utils.py for now, we can relocate later as needed.

apps/sft/main.py
```diff
@@ -38,6 +39,16 @@
...
 
 
+_SUPPORTS_FLEX_ATTENTION = (
+    torch.cuda.is_available() and torch.cuda.get_device_capability() >= (7, 5)
+)
+
+if _SUPPORTS_FLEX_ATTENTION:
+    from torch.nn.attention.flex_attention import BlockMask
+else:
+    BlockMask = torch.Tensor
```
ebsmothers: I think we can do without the conditional BlockMask logic for now. It has been in PyTorch stable for several releases now so we can just assume it's present (also I suspect we won't be running on non-CUDA devices for a bit either)

apps/sft/main.py
```diff
@@ -214,13 +242,82 @@ def train(self) -> None:
                 last_step=self.current_step == self.num_training_steps,
             )
 
+            if (
+                self._run_val_every_n_steps is not None
+                and self.current_step % self._run_val_every_n_steps == 0
+            ):
+                # pbar.refresh()
```
ebsmothers: Can remove this?

apps/sft/main.py
```diff
@@ -170,7 +197,8 @@ def forward_backward(
                     loss = self.loss_fn(pred, labels)
                 # need to free to before bwd to avoid peaking memory
                 del pred
-                loss.backward()
+                if do_backward:
```
ebsmothers, line 190: This won't work with pipeline parallel, right? Since there the backward happens inside of step(), I think we will need to handle that case differently

apps/sft/main.py
DNXie: removed.

apps/sft/main.py
DNXie: Updated. Pls check.

apps/sft/main.py
DNXie: Moved to utils.py

apps/sft/main.py
DNXie: Updated the code to use `eval` when not doing backward.

apps/sft/main.py
```diff
@@ -59,11 +62,23 @@ def __init__(self, job_config: ForgeJobConfig):
         self.current_step = 0
         self.num_training_steps = job_config.training.steps
         self.gradient_accumulation_steps = 1  # Example value, adjust as needed
+        self._run_val_every_n_steps = job_config.get("run_val_every_n_steps", None)
         super().__init__(job_config)
         self.metric_logger = None  # TODO: fix this
+        self.model = nn.ModuleList(self.model_parts)
```
ebsmothers: Why do we need to do this?

apps/sft/main.py
```diff
@@ -59,11 +62,23 @@ def __init__(self, job_config: ForgeJobConfig):
         self.current_step = 0
         self.num_training_steps = job_config.training.steps
         self.gradient_accumulation_steps = 1  # Example value, adjust as needed
+        self._run_val_every_n_steps = job_config.get("run_val_every_n_steps", None)
```
ebsmothers: This is fine as a one-off, but let's add a TODO to do this in a nicer way. One of the things I don't like about torchtune is that we've littered the training script with similar .get(value, default) statements. Titan configs come with built-in default values, we can consider landing these fields directly in ForgeJobConfig

apps/sft/main.py
DNXie: Removed `self.model`

apps/sft/main.py
DNXie: I have removed this.

PR #58, opened 2025-08-18
Skeleton of GRPO

joecummings:
### What this PR DOES do:

**Core GRPO Implementation:**

*   Complete **GRPO** training system for RL fine-tuning
*   Integration with **vLLM policy actors** for text generation during rollouts
*   **Multi-reward system** supporting math correctness and thinking tag rewards
*   **Replay buffer** with configurable batch sizes and policy version tracking
*   **Async training loops** with concurrent rollout generation and policy training

**Service Infrastructure:**

*   **Service-based architecture** using Monarch actors for distributed components
*   **Policy actors** with configurable sampling, device assignment, and tensor parallelism
*   **Trainer actors** with GRPO loss computation, KL regularization, and gradient clipping
*   **Reference model actors** for computing baseline log probabilities
*   **Advantage computation** using reward-to-go with normalization
*   Dataset integration with **GSM8K** math problems

### What this PR will NOT cover:

**Missing Features:**

*   **Weight synchronization** between trainer and policy (commented out: `# await trainer.update_weights(policy)`) @pradeepfn @pbontrager 
*   **Policy versioning** system (hardcoded to version 0) @joecummings @Jack-Khuu 
*   **Logging/monitoring** infrastructure (No wandb / tensorboard integration) @calvinpelletier @DNXie 
*   **Multi-turn conversations** or complex dialog handling 
*   **Production-ready error handling** and recovery mechanisms @allenwang28 

**Incomplete Work:**

*   **TODO:** Move policy processing initialization into setup (line 453 in [`**main.py**`](command:code-compose.open?%5B%22%2Fhome%2Fjrcummings%2Fprojects%2Fjoe-forge%2Fapps%2Fgrpo%2Fmain.py%22%2Cnull%5D "apps/grpo/main.py")) @Jack-Khuu 
*   **Manual device assignment** instead of automatic GPU allocation @pbontrager 
*   **Hardcoded hyperparameters** throughout the system

**Technical Limitations:**

*   Single model support (Qwen3-1.7B)
*   Fixed reward functions (no pluggable reward interface) @DNXie 
*   No distributed training across multiple machines 

This is a **working prototype** of GRPO with the core RL loop functional but requiring additional work for production deployment, proper weight updates, and comprehensive logging.

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -1,3 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
 import asyncio
 import time
 from dataclasses import dataclass
@@ -8,7 +14,6 @@ from datasets import load_dataset
 from forge.actors.policy import Policy, PolicyConfig, SamplingOverrides, WorkerConfig
 from forge.controller import ServiceConfig, spawn_service
 from forge.controller.actor import ForgeActor
-from forge.data.replay_buffer import ReplayBuffer
 from monarch.actor import endpoint
 from transformers import AutoModelForCausalLM, AutoTokenizer
 
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -13,9 +13,6 @@ from dataclasses import asdict, dataclass
 from typing import Dict, List
 
 import torch
-
-from forge.data.sharding import VLLMSharding
-from forge.interfaces import Policy as PolicyInterface
 from monarch.actor import Actor, current_rank, endpoint, proc_mesh
 from torchstore import MultiProcessStore
 from torchstore._state_dict_utils import DELIM
@@ -39,6 +36,9 @@ from vllm.v1.engine.processor import Processor
 from vllm.v1.request import Request
 from vllm.v1.structured_output import StructuredOutputManager
 from vllm.worker.worker_base import WorkerWrapperBase
+
+from forge.data.sharding import VLLMSharding
+from forge.interfaces import Policy as PolicyInterface
 
 
 logger = logging.getLogger(__name__)
--- a/src/forge/actors/replay_buffer.py
+++ b/src/forge/actors/replay_buffer.py
@@ -8,9 +8,9 @@ import random
 from dataclasses import dataclass
 from typing import Any
 
-from forge.controller import ForgeActor
-
 from monarch.actor import endpoint
+
+from forge.controller import ForgeActor
 
 
 @dataclass


apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+                wandb.log({"rollout_count": rollout_count, "avg_reward": avg_reward})
+                # print(
+                #     f"Generated {rollout_count} rollouts w/ average reward {avg_reward}"
+                # )
+
+    async def continuous_training():
+        training_step = 0
+        while True:
+            batch = await replay_buffer.sample.choose(curr_policy_version=0)
+            if batch is None:
```
pbontrager, line 506: I feel this should just be inside of the buffer.sample method. Then you'll just await until there's enough data for a sample.

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+    reward_actor = await spawn_service(
+        default_service_cfg,
+        RewardActor,
+        scoring_functions=[math_scoring_function, thinking_scoring_function],
+    )
+
+    print("All services initialized successfully!")
+
+    # Intialize logging
+    wandb.init(project="forge", name="grpo_test")
```
ebsmothers: Can merge with metric logging PR (#59) to get rid of this?

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+        sequence_log_probs = compute_sequence_logprobs(
+            self.model, input_ids, attention_mask, requires_grad=False
+        )
+
+        return (
+            sequence_log_probs.squeeze()
+        )  # Remove batch dimension for single response
+
+
+class DatasetActor(ForgeActor):
```
ebsmothers, line 359: Longer-term what is our plan here? (I.e. is our HfIterableDataset compatible with how we're setting things up here?)

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+    def _setup_dataset(self, *args, **kwargs):
+        def gsm8k_to_messages(sample):
+            question = sample["question"]
+            full_answer: str = sample["answer"]
+            answer = full_answer.split("#### ")[1]
+            return {"question": question, "answer": answer}
+
+        ds = load_dataset(*args, **kwargs)
+        ds = ds.map(gsm8k_to_messages)
+        ds = ds.shuffle()
```
ebsmothers, line 375: I notice no `split_dataset_by_node`. Is this by design?

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+            self.device = device
+
+        # Initialize model and tokenizer
+        self.model = AutoModelForCausalLM.from_pretrained(
+            model_name,
+            torch_dtype=torch.float16,
+            trust_remote_code=True,
+        ).to(self.device)
+
+        self.tokenizer = AutoTokenizer.from_pretrained(model_name)
```
ebsmothers: Why do we want tokenizer in the reference model?

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+            version = 0  # await policy.get_current_version.choose()
+            episode = Episode(
+                episode_id=rollout_count,
+                prompt=prompt,
+                target=target,
+                policy_version=version,
+            )
+
+            # async with policy.session(version=version):
+            #     actions: list[CompletionOutput] = await policy.generate(prompt)
```
ebsmothers: What's preventing us from doing this?

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+                group.advantage = advantage
+
+            await replay_buffer.add.choose(episode)
+
+            rollout_count += 1
+            if rollout_count % 10 == 0:
+                avg_reward = sum(group.reward for group in episode.groups) / len(
+                    episode.groups
+                )
+                wandb.log({"rollout_count": rollout_count, "avg_reward": avg_reward})
```
ebsmothers: Will this eventually be its own actor?

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+        self.model_name = model_name
+
+        # Set device
+        if device is None:
+            self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        else:
+            self.device = device
+
+        # Initialize model and tokenizer
+        self.model = AutoModelForCausalLM.from_pretrained(
```
ebsmothers, line 96: Have you tried hooking up to titan model definitions? Or is that out of scope for this PR

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+async def main():
+    """Main GRPO training loop with rollout and training processes."""
+    group_size = 5
+    model = "Qwen/Qwen3-1.7B"
+    # all_gpus = torch.cuda.device_count()
+
+    # ---- Setup services ---- #
+    default_service_cfg = ServiceConfig(
+        procs_per_replica=1,
+        num_replicas=1,
```
ebsmothers, line 394: Can we vary any of these yet? E.g. I see in some of the actors that you've hardcoded device info etc. Would be good to take that out asap

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+class ComputeAdvantages(ForgeActor):
+    """Compute advantages for GRPO using reward signals."""
+
+    def __init__(self, gamma: float = 0.99, lambda_: float = 0.95):
+        super().__init__()
+        self.gamma = gamma  # Discount factor
+        self.lambda_ = lambda_  # GAE lambda parameter
+
+    @endpoint
+    async def __call__(self, groups: list[Group]) -> list[float]:
```
ebsmothers, line 285: Is this based off of some specific reference implementation?

src/forge/data/replay_buffer.py
```diff
@@ -9,60 +9,55 @@
...
         self.batch_size = batch_size
         self.max_policy_age = max_policy_age
         if seed is not None:
             random.seed(seed)
         self.sampler = random.sample
 
     @endpoint
-    async def add(self, trajectory: Trajectory) -> None:
-        self.buffer.append(trajectory)
+    async def add(self, episode) -> None:
```
ebsmothers: So we are moving from Trajectories to Episodes in the replay buffer? Isn't this overly-tailored to GRPO?

apps/grpo/main.py
joecummings: Hmm I disagree. The contract here is just that the buffer will return a sample here when it computationally can, not when there's something inside. 

In addition, this would push possible errors a layer down if the replay buffer isn't getting filled for some reason. I'd rather have this be exposed logic to the user.

apps/grpo/main.py
joecummings: No, I'd rather keep things insanely minimal to start.

apps/grpo/main.py
joecummings: I think we need to re-build up our HfIterableDataset based on the patterns we observe when testing out this training loop. Right now, it's not super intuitive to use and it's also not clear what we actually *need* from there. 

My guess is that we'll keep 75% of the current implementation, but I don't want to assume before we actually start testing.

apps/grpo/main.py
joecummings: No, just missing it. Will add.

apps/grpo/main.py
joecummings: Shouldn't be needed actually b/c I believe we can get what we want from the tokens returned from vllm.

apps/grpo/main.py
joecummings: #70

apps/grpo/main.py
joecummings: I'm not sure it needs to be unless we're making so many requests from all over to log things and we want it going through a central source. 

Otherwise, we can just treat the logger as a normal component.

apps/grpo/main.py
joecummings: Waaaay out of scope.

apps/grpo/main.py
joecummings: I want to get everything working first and then update these.

apps/grpo/main.py
joecummings: My brain, the old impl of GRPO in torchtune, and good ol devmate.

src/forge/data/replay_buffer.py
joecummings: Everything in here is overly tailored to GRPO. I want to base our design decisions from real life, otherwise we can talk all day. This is definitely not the final state of our APIs

src/forge/actors/policy.py
```diff
@@ -359,6 +365,7 @@ def setup_worker(self):
             "distributed_init_method": distributed_init_method,
             "is_driver_worker": is_driver_worker,
         }
+        print(all_kwargs)
```
DNXie: Is this line for debugging? Should we remove this?

apps/grpo/main.py
ebsmothers: I'm not sure I trust that consensus..

apps/grpo/main.py
allenwang28: The ref logits should be from the base model, so we shouldn't be able to get it from vLLM? Unless we spin up vLLM actors specifically for getting the logits (I'm not opposed to this either)

cc @pbontrager

apps/grpo/main.py
allenwang28: would wandb complain about having multiple "sessions" created from the same job? If we can minimize the messages being passed in Monarch that generally seems like a good idea

apps/grpo/main.py
pbontrager: When does sample return None? Is it when number of usable rollouts < batch_size? I feel like this is going to surprise some people who try to use this and start running into Nones in their batch.

apps/grpo/main.py
pbontrager: Things like that can be changed by others after this code lands

apps/grpo/main.py
pbontrager: These are the tokens not the logits, that should be fine most of the time.

apps/grpo/main.py
pbontrager: I think we want it to be it's own actor so logs are called from within actors too. It's a lot of unnecessary data passing otherwise, especially if people log artifacts

apps/grpo/main.py
pbontrager: We're probably not going to want a Group type, but we put the type work on the back burner for now

src/forge/data/replay_buffer.py
pbontrager: Episode isn't GRPO specific, it's basically another word for Trajectory.

apps/grpo/main.py
```diff
@@ -0,0 +1,581 @@
...
+    async def continuous_rollouts():
+        rollout_count = 0
+        while True:
+            sample = await dataloader.__next__.choose()
+            if sample is None:
+                print("Dataloader is empty, exiting continuous rollout")
+                return
+            print(sample)
+            prompt, target = sample
+            version = 0  # await policy.get_current_version.choose()
```
DNXie, line 466: version is never updated. Is this correct?

apps/grpo/main.py
allenwang28: we should spin up a track fully dedicated to ensuring numerical correctness (which I think is the basis of @ebsmothers comment), but imo the right timing for this is after we have the working prototype

apps/grpo/main.py
allenwang28: hmm, I'm not sure if I understand that comment @pbontrager 

if the metrics logger is an actor, then all actors would need to pass messages over Monarch to the metrics actor right? If actors can just write directly to wandb that would minimize data passing?

apps/grpo/main.py
joecummings: Yeah if the replay buffer has nothing in it

apps/grpo/main.py
joecummings: Will fix later!

apps/grpo/main.py
```diff
@@ -0,0 +1,525 @@
+import asyncio
```
allenwang28, line 7: nit: add the copyright string thing
```suggestion
# Copyright (c) Meta Platforms, Inc. and affiliates.
# All rights reserved.
#
# This source code is licensed under the BSD-style license found in the
# LICENSE file in the root directory of this source tree.

"""To run:

python -m apps.grpo.main
"""
import asyncio
```

apps/grpo/main.py
```diff
@@ -0,0 +1,525 @@
...
+
+class Episode:
+    """Episode container for GRPO rollouts."""
+
+    def __init__(self, episode_id: int, prompt: str, target: str, policy_version: int):
+        self.episode_id = episode_id
+        self.prompt = prompt
+        self.target = target
+        self.policy_version = policy_version
+        self.groups: list[Group] = []
```
allenwang28, line 68: naming nit, but isn't what we're calling `Group` here actually an individual output, and what we're calling `self.groups` the actual `Group`?

apps/grpo/main.py
```diff
@@ -0,0 +1,525 @@
...
+
+        for episode in batch:
+            groups = episode.groups
+
+            # Collect all response texts and corresponding data
+            response_texts = []
+            ref_logprobs_list = []
+            advantages_list = []
+
+            for group in groups:
```
allenwang28, line 127: no action needed now, but imo it would be nifty for the episode structure itself to handle things like this

```
for episode in batch:
    tokenized = self.tokenizer(
        episode.as_response_group(),
        ...
    )
```

apps/grpo/main.py
allenwang28: I think exposing logic to the user makes sense, my only nit is that I might want it to have more of a queue-like semantic? Like we set a timeout, it raises an error if we hit the timeout etc.

src/forge/actors/policy.py
```diff
@@ -106,6 +103,11 @@ class Policy(PolicyInterface):
     @endpoint
     async def setup(self):
         # Set up policy_worker
+        self.available_devices = (
```
pbontrager, line 107: This should really come from the script and not the config. Ideally the script makes the master set of ranks and then passes them to the actors.

apps/grpo/main.py
joecummings: Fair lol

apps/grpo/main.py
joecummings: Does `choose` have a built-in timeout feature?

allenwang28: it doesn't, but the buffer itself can return an exception which will be propagated through choose

but the current service implementation will then mark the replica as unhealthy so we shouldn't add this yet

PR #59, opened 2025-08-19
Metric Logging

calvinpelletier:
Build metric logger: 
```python
metric_logger = get_metric_logger(**config.metrics)
````

Log a metric:
```python
metric_logger.log("name", value, step)
```

If you want to check if this is a log step before calculating the metric:
```python
if metric_logger.is_log_step(name, step):
    value = calculate_metric()
    metric_logger.log(name, value, step)
```

Open questions:
- multiple loggers? e.g. both wandb and stdout
- non-constant log frequencies? (e.g. for metrics that take a long time to calculate, I like to use a logarithmic frequency so that i get quick data points early in training and fewer points as training becomes more stable)  

Follow up PRs:
- put metric logger inside an actor


apps/sft/llama3_8b.yaml
```diff
@@ -2,10 +2,10 @@
 # profiling:
 #   enable_profiling: false
 
-# metrics:
-#   log_freq: 10
-#   enable_tensorboard: true
-#   save_tb_folder: "tb"
+metrics:
+  logger: "stdout"
```
ebsmothers: nit: don't need the quotes here (I think)

apps/sft_v2/main.py
```diff
@@ -74,7 +75,7 @@ class ForgeSFTRecipe(ForgeActor, ForgeEngine):
     def __init__(self, job_config: ForgeJobConfig):
         self.current_step = 0
         self.num_training_steps = job_config.training.steps
-        self.metric_logger = None  # TODO: fix this
+        self.metric_logger = get_metric_logger(**job_config.metrics)
```
ebsmothers: Will this work in the Monarch-style SFT script? (Worth testing cause I thought for that case we wanted to wrap in an Actor)

src/forge/util/logging.py
```diff
@@ -0,0 +1,147 @@
...
+    Args:
+        logger (logging.Logger): The logger.
+        msg (str): The warning message.
+        level (int): The logging level. See https://docs.python.org/3/library/logging.html#levels for values.
+            Defaults to ``logging.INFO``.
+    """
+    log_rank_zero(logger=logger, msg=msg, level=level)
+
+
+def deprecated(msg: str = "") -> Callable[[T], T]:
```
ebsmothers: Hopefully we don't need anything from here on down yet.. I would just remove it

src/forge/types.py
```diff
@@ -98,3 +101,6 @@ class ProcessConfig:
     oncall: str = "torchtune"
     identity: str = "pytorch_distributed"
     image: str = "forge_workspace:latest"
+
+
+Scalar = Union[Tensor, ndarray, int, float]
```
ebsmothers: nit but I wouldn't call it a Scalar (since a Tensor is kinda by definition not a scalar)

src/forge/interfaces.py
```diff
@@ -150,3 +150,110 @@ def tokenize_messages(
...
+    Args:
+        log_freq (Mapping[str, int]):
+            calls to `log` and `log_dict` will be ignored if `step % log_freq[metric_name] != 0`
+    """
+
+    def __init__(self, log_freq: Mapping[str, int]):
+        self._log_freq = log_freq
+        self._step = None
+
+    def set_step(self, step: int) -> None:
```
ebsmothers: Is there a specific reason we want to enable calls to log with optional step argument? Imo it makes the logic a bit more complicated, it would be nice to just guarantee this is being passed all the time.

src/forge/interfaces.py
```diff
@@ -150,3 +150,110 @@ def tokenize_messages(
...
+            step (int): current step. if not given, will use the one last provided via set_step()
+        """
+        if step is None:
+            assert (
+                self._step is not None
+            ), "`step` arg required if `set_step` has not been called."
+            step = self._step
+        return step % self._log_freq[name] == 0
+
+    def log(
```
ebsmothers: Personally I would lean towards keeping implementations out of the ABC as much as possible. Especially cause in this case (1) we are duplicating `log`/`_log` and `log_dict`/`_log_dict`, all pretty much as syntactic sugar to handle the logic of `step % log_freq == 0`.

src/forge/util/metric_logging.py
```diff
@@ -0,0 +1,203 @@
...
+
+# TODO: replace with direct instantiation via a path to the class in the config
+METRIC_LOGGER_STR_TO_CLS = {
+    "stdout": StdoutLogger,
+    "tensorboard": TensorBoardLogger,
+    "wandb": WandBLogger,
+}
+
+
+def _get_rank():
```
ebsmothers: We can just use [get_world_size_and_rank](https://github.com/pytorch/torchtune/blob/main/torchtune/training/_distributed.py#L306) and put in util/distributed.py or something

src/forge/interfaces.py
calvinpelletier: I personally like this style of metric logging because then you can just notify the logger of the step once in the main training loop and then all the rest of your metric code doesn't need to worry about passing around a step variable.

But I can change it to keep the logger as simple as possible

apps/sft_v2/main.py
calvinpelletier: I was originally going to do that in a follow-up PR. I wasn't able to get the monarch-style sft script working even without my changes

src/forge/types.py
calvinpelletier: yeah this came from torchtune code, I assume it was signifying that the tensor/ndarrays would only have one value in them. I'll remove it and enforce int or float

src/forge/util/metric_logging.py
```diff
@@ -0,0 +1,246 @@
...
+        organize_logs (bool): If `True`, this class will create a subdirectory within `log_dir` for the current
+            run. Having sub-directories allows you to compare logs across runs. When TensorBoard is
+            passed a logdir at startup, it recursively walks the directory tree rooted at logdir looking for
+            subdirectories that contain tfevents data. Every time it encounters such a subdirectory,
+            it loads it as a new run, and the frontend will organize the data accordingly.
+            Recommended value is `True`. Run `tensorboard --logdir my_log_dir` to view the logs.
+        **kwargs: additional arguments
+
+    Example:
+        >>> from torchtune.training.metric_logging import TensorBoardLogger
```
ebsmothers: Make sure to update the docstrings 😃 (I notice something similar in WandBLogger too)

apps/sft/llama3_8b.yaml
```diff
@@ -2,10 +2,11 @@
 # profiling:
 #   enable_profiling: false
 
-# metrics:
-#   log_freq: 10
-#   enable_tensorboard: true
-#   save_tb_folder: "tb"
+metrics:
```
ebsmothers, line 5: Probably the proper long-term thing to do is to define our own class [here](https://github.com/pytorch/torchtitan/blob/7d744b2318b279f9ae3afbec5a5f1f53dd740964/torchtitan/experiments/forge/job_config.py#L25) (it'll be helpful to have a single source of truth for our various config fields). But doesn't have to block this PR

apps/sft/main.py
```diff
@@ -60,7 +61,7 @@ def __init__(self, job_config: ForgeJobConfig):
         self.num_training_steps = job_config.training.steps
         self.gradient_accumulation_steps = 1  # Example value, adjust as needed
         super().__init__(job_config)
-        self.metric_logger = None  # TODO: fix this
+        self.metric_logger = get_metric_logger(**job_config.metrics)
```
ebsmothers, line 67: Related to my other comment -- once we have a proper `job_config.metrics` field we could drop the unpack here and just take `job_config.metrics` directly here. This would be more consistent with how APIs are called in ForgeEngine

apps/sft/llama3_8b.yaml
```diff
@@ -2,10 +2,11 @@
...
+  logger: wandb
+  project: dsafjkdsafjlskdjf
```
ebsmothers: Let's remove this (is project actually necessary for WandB? I thought they would create one for you if you don't pass it).

apps/sft/llama3_8b.yaml
```diff
@@ -2,10 +2,11 @@
 # profiling:
 #   enable_profiling: false
 
-# metrics:
-#   log_freq: 10
-#   enable_tensorboard: true
-#   save_tb_folder: "tb"
+metrics:
+  logger: wandb
```
ebsmothers: Do we want this to be the default logger? Personally I prefer it but (a) not sure it works on MAST (may need tensorboard there) and (b) we may want to add it as a dependency

calvinpelletier: yeah probably best for tensorboard to be the default unless we decide on adding wandb as a dep

calvinpelletier:
@ebsmothers addressed the comments. I'll leave the config class for the next PR after adding metric logging for sft_v2 and toy_rl

src/forge/util/logging.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Get a logger with a stream handler.
+
+    Args:
+        level (Optional[str]): The logging level. See https://docs.python.org/3/library/logging.html#levels for list of levels.
+
+    Example:
+        >>> logger = get_logger("INFO")
+        >>> logger.info("Hello world!")
+        INFO:torchtune.utils._logging:Hello world!
```
ebsmothers, line 26:
```suggestion
        INFO:forge.util.logging:Hello world!
```

src/forge/util/metric_logging.py
```diff
@@ -0,0 +1,285 @@
...
+
+class WandBLogger(MetricLogger):
+    """Logger for use w/ Weights and Biases application (https://wandb.ai/).
+    For more information about arguments expected by WandB, see https://docs.wandb.ai/ref/python/init.
+
+    Args:
+        freq (Mapping[str, int]):
+            calls to `log` and `log_dict` will be ignored if `step % freq[metric_name] != 0`
+        log_dir (Optional[str]): WandB log directory.
+        project (str): WandB project name. Default is `torchtune`.
```
ebsmothers, line 174:
```suggestion
        project (str): WandB project name.
```

src/forge/util/logging.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Args:
+        logger (logging.Logger): The logger.
+        msg (str): The warning message.
+        level (int): The logging level. See https://docs.python.org/3/library/logging.html#levels for values.
+            Defaults to ``logging.INFO``.
+    """
+    rank = dist.get_rank() if dist.is_available() and dist.is_initialized() else 0
+    if rank != 0:
+        return
```
DNXie, line 52: Shouldn't log rank be configurable like how [titan](https://github.com/pytorch/torchtitan/blob/ad06609c4e3ac119112b097b6b317f31c5c760f5/run_train.sh#L14) did? But I guess we can always improve later.

PR #60, opened 2025-08-19
Add Grouped Rollout to Policy

Jack-Khuu:
For Grouped Rollouts (n > 1), a `ParentRequest` is introduced to aggregate and track the fanned out children.

`policy_router.generate` will return a list of `CompletionOutputs` when all the children of the ParentRequest complete (this is maintained internally in vllm's [OutputProcessor](https://github.com/vllm-project/vllm/blob/21dce80ea96bcf033d159c0f952fb274567b315c/vllm/v1/engine/output_processor.py#L274))

---
Followups 
- Tests should be moved into CI tests
- Customization will be piped up to configs

image

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -390,7 +390,7 @@ if __name__ == "__main__":
         "enforce_eager": True,
         "resources": 2,
     }
-    asyncio.run(_test(config))
+    # asyncio.run(_test(config))
     # asyncio.run(_test(config, guided_decoding=True))
     # asyncio.run(_test(config, num_samples=2))
     # asyncio.run(_test(config, guided_decoding=True, num_samples=3))


src/forge/actors/policy.py
```diff
@@ -110,36 +115,49 @@ async def generate(self, prompt: str, priority: int = 0):
...
+                self.scheduler.add_request(child_request)
+            request_fut = asyncio.Future()
+            self.requests[request_id] = (parent_req, request_fut)
 
         return await request_fut
 
+    # Abstracted to match vllm
+    # https://github.com/vllm-project/vllm/blob/0e3bb543f064eb416bca4f6f3013efa3830b12f7/vllm/v1/engine/core.py#L419
+    def preprocess_add_request(self, request: EngineCoreRequest) -> tuple[Request, int]:
+        if request.mm_hashes is not None:
```
pbontrager, line 151: Maybe we should raise an NotImplemented error here?

src/forge/actors/policy.py
```diff
@@ -154,15 +172,26 @@ async def run(self):
...
                 iteration_stats=None,
             )
-            for output in processed_outputs.request_outputs:
-                if output.finished:
-                    fut = self.requests.pop(output.request_id)
-                    fut.set_result(output)
+            for request_output in processed_outputs.request_outputs:
+                if request_output.finished:
+                    if isinstance(
+                        payload := self.requests.get(request_output.request_id), tuple
```
pbontrager: Can we simplify this by making requests always a tuple and just always grabbing the parent request (even if there's no children) and get rid of this entire if/else block?

src/forge/actors/policy.py
```diff
@@ -110,36 +114,48 @@ async def generate(self, prompt: str, priority: int = 0):
...
-            #     output_processor.add_request(child_request, prompt_str,
-            #                                     parent_req, idx)
-            #     child_request = Request.from_engine_core_request(child_request)
-            #     self.scheduler.add_request(chile_request)
+            parent_req = ParentRequest(request_id, self.sampling_params)
+            for idx in range(num_samples):
+                # Note: `get_child_info` mutates ParentRequest to track the
+                # generated child request
+                child_request_id, params = parent_req.get_child_info(idx)
+                child_request = request if idx == num_samples - 1 else copy(request)
```
ebsmothers, line 134: What's the purpose of this line? Like why is the last sample being treated differently here?

Jack-Khuu: Micro-optimization mirrored from vllm's llm_engine

It's one less copy of the request (not attached to it)

PR #61, opened 2025-08-20
[Debugging] Push PolicyWorker into Router and leverage Services

Jack-Khuu:
Refactors 
- `PolicyRouter` -> `Policy`
- `Policy` -> `PolicyWorker`

Pushed Mesh and Worker set up into Policy (used to be Router)

Note: Will need follow up PR's to refactor to use @allenwang28 updated Service when ready

---
```
python src/forge/actors/policy.py
```

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -111,9 +111,6 @@ class Policy(PolicyInterface):
             log_stats=None,
         )
 
-    def should_spawn_workers(self) -> bool:
-        return True
-
     async def spawn_workers(self):
         self.worker_mesh = await proc_mesh(
             gpus=self.config.num_workers,
@@ -186,6 +183,7 @@ class Policy(PolicyInterface):
             request_fut = asyncio.Future()
             self.requests[request_id] = (parent_req, request_fut)
 
+        print("Awaiting in Generate")
         return await request_fut
 
     # Abstracted to match vllm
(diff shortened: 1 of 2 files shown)


src/forge/actors/policy.py
```diff
@@ -82,6 +86,34 @@ async def setup(self):
             log_stats=None,
         )
 
+    async def setupWorker(self, config, guided_decoding, num_samples):
```
joecummings: Bad Javascript developer

src/forge/actors/policy.py
```diff
@@ -82,6 +86,34 @@ async def setup(self):
...
+        self.worker_mesh = await proc_mesh(
+            gpus=config["resources"],
+            env={
+                "MASTER_ADDR": str(get_loopback_ip()),
+                "MASTER_PORT": str(get_open_port()),
+            },
+        )
```
allenwang28, line 121: hmm, we might want to keep the worker mesh's procs distinctly separate...

The reason is because downstream, we will want something that can monitor the proc health and re-spawn it if necessary: https://github.com/meta-pytorch/forge/blob/main/src/forge/controller/service.py#L948-L977

I think having vLLM router handling this will be unnecessarily complex

Jack-Khuu: Will wait to tackled on top of the planned Service change

Jack-Khuu:
History on this PR is borked

See https://github.com/meta-pytorch/forge/pull/70

PR #63, opened 2025-08-20
Modifies service.py to use get_proc_mesh, and unifies the ProcConfig with ServiceConfig

allenwang28:
Main changes:
- `ServiceConfig` now includes fields in `ProcConfig` and has conversion utility into a `ProcConfig`. I prefer this flat structure a bit more to avoid needless nested dataclasses
- Changes RecoverableProcMesh to use our `get_proc_mesh` capabilities


src/forge/types.py
```diff
@@ -98,3 +98,35 @@ class ProcessConfig:
...
+    identity: str = "pytorch_distributed"
+    image: str = "forge_workspace:latest"
+    # ServiceConfig-specific fields
+    health_poll_rate: float = 0.2
+    replica_max_concurrent_requests: int = 10
+    return_first_rank_result: bool = (
+        True  # Whether or not to auto-unwrap ValueMesh to first rank's result
+    )
+
+    def to_process_config(self) -> ProcessConfig:
```
Jack-Khuu, line 121: Are the args used to construct `ProcessConfig` needed at at the `ServiceConfig` level?

We can just save the ProcessConfig as a member in `ServiceConfig` otherwise

src/forge/controller/recoverable_mesh.py
```diff
@@ -121,9 +126,10 @@ class RecoverableProcMesh(MeshTrait):
...
         self,
-        num_procs: int,
+        proc_config: ProcessConfig,
     ) -> None:
-        self.num_procs = num_procs
+        self._proc_config: ProcessConfig = proc_config
+        self.num_procs = proc_config.num_procs
         self._proc_mesh: Optional[ProcMesh] = None
         self._recovery_task: Optional[asyncio.Task[None]] = None
         self.state: MeshState = MeshState.UNHEALTHY
```
Jack-Khuu, line 135: Any reason why `RecoverableProcMesh` can't be a dataclasses

src/forge/types.py
allenwang28: yeah they are, I specifically wanted to keep the structure flat because I personally find it annoying to have to do this all the time:

```
config = ServiceConfig(proc_config=ProcessConfig(...))
```

src/forge/controller/recoverable_mesh.py
allenwang28: good question.. I am thinking through how to handle the policy in service so I may very well turn it into a dataclass, coming up next

PR #65, opened 2025-08-22
[Service] Refactors service to align with EX545081

allenwang28:
Specifically, this PR:
- Introduces replica, which:
  - handles process lifecycle, async request queueing, fault recovery
  - modularizes it a bit more than the prior implementation
- Since replica pulls in functionality of `RecoverableProcMesh`, we remove it altogether
- Refactors service to use Replica
- Modifies test_service to be correct

Next PRs:
- add endpoints to service, to retain the ability to call() and choose()
- place replicas on their own procs
- place services on their own procs
- introduce sharded/distributed services? Or multiple services for envs

### DIFF
--- a/src/forge/controller/replica.py
+++ b/src/forge/controller/replica.py
@@ -208,7 +208,6 @@ class Replica:
             try:
                 logger.debug("Creating new proc_mesh for replica %d", self.idx)
                 await self.initialize()
-                self.state = ReplicaState.HEALTHY
                 logger.debug("Recovery completed successfully for replica %d", self.idx)
             except Exception as e:
                 logger.error("Recovery failed for replica %d: %s", self.idx, e)
@@ -216,7 +215,6 @@ class Replica:
                 raise
 
         logger.debug("Starting recovery for replica %d", self.idx)
-        self.state = ReplicaState.RECOVERING
         self._recovery_task = asyncio.create_task(_do_recovery())
         await self._recovery_task
 


src/forge/controller/replica.py
```diff
@@ -0,0 +1,505 @@
...
+            self.successful_requests += 1
+        else:
+            self.failed_requests += 1
+
+    def get_request_rate(self, window_seconds: float = 60.0) -> float:
+        """Gets requests per second over the last window_seconds."""
+        now = time.time()
+        cutoff = now - window_seconds
+        recent_requests = [t for t in self.request_times if t >= cutoff]
+        return len(recent_requests) / window_seconds if window_seconds > 0 else 0.0
```
Jack-Khuu: nit: if `window_seconds` is <= 0 we should error since it's probably a bug

src/forge/controller/replica.py
```diff
@@ -0,0 +1,505 @@
...
+
+        This method handles the complete actor spawning process including
+        recovery if the proc_mesh has failed.
+        """
+        # Ensure we have a healthy proc_mesh
+        await self._ensure_healthy_proc_mesh()
+
+        if not self.proc_mesh:
+            raise RuntimeError(
+                f"Replica {self.idx}: proc_mesh is None after recovery attempt"
```
Jack-Khuu, line 247: Can't the replica just have an uninitialized proc_mesh?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,505 @@
...
+
+            # Spawn the actor
+            self.actor = await self.proc_mesh.spawn(
+                actor_name,
+                actor_def,
+                *actor_args,
+                **actor_kwargs,
+            )
+
+            # Call setup if it exists
```
Jack-Khuu:
```suggestion
```

Replica setup always(?) exists, we can hide the actor check in `setup` impl

src/forge/controller/replica.py
```diff
@@ -0,0 +1,505 @@
...
+            return
+
+        if self.actor is None:
+            raise RuntimeError(f"Cannot setup replica {self.idx}: actor is None")
+
+        try:
+            # Call actor setup if it exists
+            if hasattr(self.actor, "setup"):
+                # TODO - should this be a standard in our Forge Actor(s)?
+                await self.actor.setup.call()
```
Jack-Khuu: my 2c: We should commit to the setup pattern, but this is a different discussion in a later PR (this looks good, don't change for now)

src/forge/controller/replica.py
```diff
@@ -0,0 +1,505 @@
...
+        # TODO - for policy replica, we would override this method to
+        # include multiple proc_meshes
+        if self.proc_mesh is not None:
+            logger.warning("Proc mesh already initialized for replica %d", self.idx)
+            return
+
+        logger.debug("Initializing proc_mesh for replica %d", self.idx)
+        try:
+            self.proc_mesh = await get_proc_mesh(process_config=self.proc_config)
+            logger.debug("Proc mesh initialized successfully for replica %d", self.idx)
```
Jack-Khuu: Do we wanna make initialized a state?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,505 @@
...
+
+    def start_processing(self):
+        """Start the replica's processing loop if not already running."""
+        if self._run_task is None or self._run_task.done():
+            self._run_task = asyncio.create_task(self.run())
+            logger.debug("Started processing loop for replica %d", self.idx)
+
+    async def enqueue_request(self, request: ServiceRequest):
+        """Enqueues a request for processing by this replica."""
+        if self.state == ReplicaState.STOPPED:
```
Jack-Khuu: Do we want to enqueue if the state is anything but `HEALTHY`?

src/forge/controller/replica.py
allenwang28: fixed

src/forge/controller/replica.py
allenwang28: I was looking through this part of the code and it was still a bit unorganized. Took a pass at making it clearer

src/forge/controller/replica.py
allenwang28: sg, I agree - let's bake this into a follow up

src/forge/controller/replica.py
allenwang28: IMHO, initialized but not healthy is not really a state we should expect, that would be UNHEALTHY. I hope the code re-organization makes this a bit clearer!

src/forge/controller/replica.py
allenwang28: Yeah good question - I think from the replica perspective, makes sense to allow enqueues when we expect the replica to come back up at some point

(the service as currently designed won't enqueue to not healthy replicas though)

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+            if not self.proc_mesh:
+                raise RuntimeError(
+                    f"Replica {self.idx}: proc_mesh is None after creation"
+                )
+
+            # Determine actor name
+            if "name" in actor_kwargs:
+                actor_name = actor_kwargs.pop("name")
+            else:
+                actor_name = actor_def.__name__
```
Jack-Khuu:
```suggestion
            actor_name = actor_kwargs.pop("name", actor_def.__name__)
```

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+                actor_def,
+                *actor_args,
+                **actor_kwargs,
+            )
+
+            # Call actor setup if it exists
+            if hasattr(self.actor, "setup") and self.actor is not None:
+                setup_method = getattr(self.actor, "setup", None)
+                if setup_method is not None:
+                    await setup_method.call()
```
Jack-Khuu:
```suggestion
            if setup_method := getattr(self.actor, "setup", None):
                await setup_method.call()
```

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+            self.start_processing()
+
+            logger.debug("Replica %d initialization complete", self.idx)
+
+        except Exception as e:
+            logger.error("Failed to initialize replica %d: %s", self.idx, e)
+            self.state = ReplicaState.UNHEALTHY
+            raise
+
+    async def spawn_actor(self, actor_def, *actor_args, **actor_kwargs):
```
Jack-Khuu, line 238: Duplicate code?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+            # Recovery already in progress, wait for it
+            await self._recovery_task
+            return
+
+        logger.debug("Starting recovery for replica %d", self.idx)
+        self.state = ReplicaState.RECOVERING
+
+        # Create the recovery task
+        self._recovery_task = asyncio.create_task(self._do_recovery())
+        await self._recovery_task
```
Jack-Khuu: It's been a while since I've played with distributed code, but this looks like an interesting set up.
Is the double await a common pattern?

Jack-Khuu:
```suggestion
        if self.state is not ReplicaState.RECOVERING:
          logger.debug("Starting recovery for replica %d", self.idx)
          self.state = ReplicaState.RECOVERING
          self._recovery_task = asyncio.create_task(self._do_recovery())
        await self._recovery_task
```

Does this work or are the checks needed?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+        self.actor = None
+
+        # Stop old proc_mesh if it exists
+        if old_proc_mesh is not None:
+            try:
+                await old_proc_mesh.stop()
+                logger.debug("Old proc_mesh stopped for replica %d", self.idx)
+            except Exception as e:
+                logger.warning(
+                    "Error stopping old proc_mesh for replica %d: %s", self.idx, e
```
Jack-Khuu: Is there a dangling proc_mesh issue if we set `self.proc_mesh` to None and fail to stop?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+        # Wait for processor to finish if it's running
+        if self._running:
+            # Give it a moment to finish current request and exit gracefully
+            for _ in range(50):  # Wait up to 5 seconds
+                if not self._running:
+                    break
+                await asyncio.sleep(0.1)
+
+            if self._running:
+                logger.warning("Replica %d processor didn't stop gracefully", self.idx)
```
Jack-Khuu: I like for-else, but I know it's controversial
```suggestion
            for _ in range(50):  # Wait up to 5 seconds
                if not self._running:
                    break
                await asyncio.sleep(0.1)
            else:
                logger.warning("Replica %d processor didn't stop gracefully", self.idx)
```

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+
+        # Fail any remaining requests in the queue
+        failed_requests = []
+        while not self.request_queue.empty():
+            try:
+                request = self.request_queue.get_nowait()
+                failed_requests.append(request)
+                self.request_queue.task_done()
+            except asyncio.QueueEmpty:
```
Jack-Khuu, line 455: Do we need the try catch?

Jack-Khuu:
Logic looks legit to me. A couple of questions, but we should be ready to go

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+                        request.future.set_exception(
+                            RuntimeError(f"Replica {self.idx} is still recovering")
+                        )
+                        self.request_queue.task_done()
+                        continue
+
+                    # Process the request
+                    asyncio.create_task(self._process_single_request(request))
+
+                except asyncio.TimeoutError:
```
joecummings, line 375: nit: couldn't we just set a larger timeout and/or parameterize it? Seems cleaner than a try except for timeout that allows to continue polling.

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+                    # Wait for a request with timeout to check health periodically
+                    request = await asyncio.wait_for(
+                        self.request_queue.get(), timeout=self._run_poll_rate_s
+                    )
+
+                    # Check if we have capacity - if we have too many ongoing,
+                    # we will put the request back and wait.
+                    if self.active_requests >= self.max_concurrent_requests:
+                        await self.request_queue.put(request)
+                        await asyncio.sleep(0.1)
```
joecummings: Why is this needed?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+    async def run(self):
+        """Runs the main processing loop for the replica.
+
+        Continuously processes requests from the queue while the replica is healthy.
+        Handles capacity management and graceful degradation on failures.
+        """
+        self._running = True
+
+        try:
+            while self.state in (ReplicaState.HEALTHY, ReplicaState.RECOVERING):
```
joecummings: Why do we want to try to run this for RECOVERING?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+
+                    # Check if we have capacity - if we have too many ongoing,
+                    # we will put the request back and wait.
+                    if self.active_requests >= self.max_concurrent_requests:
+                        await self.request_queue.put(request)
+                        await asyncio.sleep(0.1)
+                        continue
+
+                    # If we're recovering, reject the request
+                    if self.state == ReplicaState.RECOVERING:
```
joecummings: nit should we use the properties defined on the class for state checking?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+                    # Process the request
+                    asyncio.create_task(self._process_single_request(request))
+
+                except asyncio.TimeoutError:
+                    # No requests, just continue checking for new ones
+                    continue
+
+                except Exception as e:
+                    logger.error(
+                        "Error in replica %d processing loop: %s",
```
joecummings, line 381:
```python
adjective = "fun"
f"F strings are very {adjective}"
```

src/forge/controller/replica.py
joecummings: Hmmm it looks like this is the only place _ensure_healthy_proc_mesh is called, which means this is the only place that does recovery?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+    def healthy(self) -> bool:
+        return self.state == ReplicaState.HEALTHY
+
+    @property
+    def failed(self) -> bool:
+        """Check if the replica has failed and needs recovery."""
+        return self.state in (ReplicaState.RECOVERING, ReplicaState.UNHEALTHY)
+
+    def mark_failed(self):
+        """Mark the replica as failed, triggering recovery."""
```
joecummings, line 420: It doesn't actually trigger recovery though, right?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+        Fails any remaining requests in the queue.
+        """
+        logger.debug("Stopping replica %d", self.idx)
+
+        # Transition to stopped state to signal the run loop to exit
+        self.state = ReplicaState.STOPPED
+
+        # Wait for processor to finish if it's running
+        if self._running:
+            # Give it a moment to finish current request and exit gracefully
```
joecummings: This seems .... hacky, no?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,527 @@
...
+                await self.proc_mesh.stop()
+            except Exception as e:
+                logger.warning(
+                    "Error stopping proc_mesh for replica %d: %s", self.idx, e
+                )
+
+    # Metric-related getters
+
+    @property
+    def load(self) -> int:
```
joecummings: nit: load -> curr_load / curr_num_requests / idk something better than `load`

src/forge/controller/replica.py
allenwang28: embarrassing lol, fixed by using spawn_actor in initialize

src/forge/controller/replica.py
allenwang28: the proc_mesh will still close in this case, it'll just show an ugly error message once forge exits

src/forge/controller/replica.py
allenwang28: added a comment - a request could be grabbed during health loop and while stop, so this is needed

src/forge/controller/replica.py
allenwang28: yeah, it's essentially a concurrency check - i.e. in case multiple callers are calling `recover()` - we don't necessarily expect this to be the case now but this seems like a reasonable defensive optimization

src/forge/controller/replica.py
allenwang28: The timeout is parameterized already and we can't set that to infinity because then we won't be able to stop the replica

We shouldn't remove this except but we can set the variable to a high number, it just affects how long it takes to shutdown

src/forge/controller/replica.py
allenwang28: yeah good call, changed

src/forge/controller/replica.py
allenwang28: are f formatted better for logger? There was a whole thing at Google for awhile where I got nitted to death to use %s/%d instead of formatting in logger...

src/forge/controller/replica.py
allenwang28: yeah thanks for checking this. I refactored this to explicitly initialize() and recover(), and the service calls these respectively rather than re-using a single init function. It will actually trigger recover now

src/forge/controller/replica.py
allenwang28: refactored

src/forge/controller/replica.py
allenwang28: done

src/forge/controller/replica.py
allenwang28: Good catch, I changed it so that it only runs for healthy. If it ever becomes not healthy, the loop stops (can't accept anymore requests) and re-spawns at recovery.

src/forge/controller/replica.py
allenwang28: it's to avoid a busy-wait (like put back in the queue and check again) which would take 100% of the cycles

But this is a bad approach, I added a semaphore which is a cleaner approach

src/forge/controller/replica.py
joecummings: F strings are technically faster, that's all I know.

src/forge/controller/replica.py
```diff
@@ -0,0 +1,514 @@
...
+    _run_task: Optional[asyncio.Task] = None
+
+    # Metrics tracking
+    metrics: ReplicaMetrics = field(default_factory=ReplicaMetrics)
+
+    def __post_init__(self):
+        # This semaphore is used to enforce max_concurrent_requests
+        # Once it is acquired max_concurrent_requests times, future
+        # requests are blocked until standing requests complete.
+        self._capacity_semaphore = asyncio.Semaphore(self.max_concurrent_requests)
```
joecummings, line 143: kewl, I didn't know asyncio had this

src/forge/controller/replica.py
```diff
@@ -0,0 +1,514 @@
...
+        - Transitions to healthy state
+        - Starts the processing loop
+        """
+        assert self.proc_mesh is None, "Proc mesh should not be set yet"
+        try:
+            # Create proc_mesh
+            await self.create_proc_mesh()
+
+            # Ensure we have a healthy proc_mesh
+            if not self.proc_mesh:
```
joecummings, line 164: Dumb question, how would we get here? If the proc mesh creation errors, wouldn't it go to the exception immediately? And isn't that the only wait to have a none for self.proc_mesh?

src/forge/controller/replica.py
allenwang28: using f-strings now

src/forge/controller/replica.py
```diff
@@ -0,0 +1,514 @@
...
+            )
+            # Transition to healthy state and start processing
+            self.state = ReplicaState.HEALTHY
+            self.start_processing()
+
+            logger.debug(f"Replica {self.idx} initialization complete")
+
+        except Exception as e:
+            logger.error(f"Failed to initialize replica {self.idx}: {e}")
+            self.state = ReplicaState.UNHEALTHY
```
joecummings, line 183: Would this be unhealthy or would it just be uninitialized?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,514 @@
...
+                    logger.debug("Old proc_mesh stopped for replica %d", self.idx)
+                except Exception as e:
+                    logger.warning(
+                        "Error stopping old proc_mesh for replica %d: %s", self.idx, e
+                    )
+
+            try:
+                logger.debug("Creating new proc_mesh for replica %d", self.idx)
+                await self.initialize()
+                self.state = ReplicaState.HEALTHY
```
joecummings: Initialize with already set this state.

src/forge/controller/replica.py
```diff
@@ -0,0 +1,514 @@
...
+                logger.debug("Recovery completed successfully for replica %d", self.idx)
+            except Exception as e:
```
joecummings, line 212: Won't this be handled in `initialize`?

src/forge/controller/replica.py
```diff
@@ -0,0 +1,514 @@
...
+
+            # Spawn the actor
+            self.actor = await self.proc_mesh.spawn(
+                actor_name,
+                actor_def,
+                *actor_args,
+                **actor_kwargs,
+            )
+            # Call actor setup if it exists
+            if setup_method := getattr(self.actor, "setup", None):
```
joecummings, line 262: Can't wait for this hidden gem to trip some people up haha

src/forge/controller/replica.py
```diff
@@ -0,0 +1,514 @@
...
+
+    def start_processing(self):
+        """Start the replica's processing loop if not already running."""
+        if self._run_task is None or self._run_task.done():
+            self._run_task = asyncio.create_task(self.run())
+            logger.debug("Started processing loop for replica %d", self.idx)
+
+    async def enqueue_request(self, request: ServiceRequest):
+        """Enqueues a request for processing by this replica."""
+        if self.stopped:
```
joecummings, line 282: What's the raw error if you let this through?

src/forge/controller/replica.py
allenwang28: not a dumb question - create_proc_mesh would raise the exception but this check is for if a downstream replica implementation (like Policy) doesn't do this correctly - this is another layer of guardrails. But with this current implementation, we will never hit it

src/forge/controller/replica.py
allenwang28: hmm good question  - uninitialized is like "the service just started and i haven't tried to init" vs unhealthy is like "the service tried to init but couldn't for some reason"

so in this context, unhealthy basically means "this thing is never going to initialize, you should look at your code", but right now I'm not doing anything. We can follow up when this inevitably becomes a problem

src/forge/controller/replica.py
allenwang28: yeah this basically just adds another layer of logs saying "oh it failed during recovery, because it failed to initialize"

src/forge/controller/replica.py
allenwang28: the service waits forever for a request that will never complete

PR #69, opened 2025-08-25
[Service] Turns Service into an Actor and splits service into its own files

allenwang28:
This diff makes a few changes:
- Changes Service into an actor, allowing it to run on its own process, allowing it to scale out better.
  - This turns `_call`, `__initialize__`, `_call_all`(enqueue on all healthy replicas) , `start_session`, `get_metrics`, `get_metrics_summary`, `terminate_session`, `stop`, into endpoints
- Introduces a ServiceInterface. This retains the existing session APIs, but now creates endpoints over the actor_def endpoints. It specifically introduces:
  - `choose()`: run this endpoint on a replica
  - `call()`: run this endpoint on all replicas
- Fixes a bug where `sess_id` is an arg. Turns it into a `kwarg` cc @joecummings  
- Refactors service.py into
  -  interface.py (including session and ServiceInterface), 
  - metrics.py (including ServiceMetrics),
  - service.py (the base Service actor),
- adds new tests

### DIFF
--- a/src/forge/controller/interface.py
+++ b/src/forge/controller/interface.py
@@ -117,6 +117,10 @@ class ServiceInterface:
 
     The ServiceInterface acts as the handle that is returned to end clients,
     providing a simple interface that makes actual calls to the Service Actor.
+
+    This is also needed to simplify serializing a handle to the service, in case
+    we want to pass this to other actors in the future.
+
     """
 
     def __init__(self, _proc_mesh, _service, actor_def):


src/forge/controller/interface.py
```diff
@@ -0,0 +1,184 @@
...
+
+    async def call(self, *args: P.args, **kwargs: P.kwargs) -> List[R]:
+        """Broadcasts a request to all healthy replicas and returns the results as a list."""
+        result = await self.actor_mesh._call_all.call_one(
+            self.endpoint_name, *args, **kwargs
+        )
+        return result
+
+
+class ServiceInterface:
```
joecummings, line 110: Why is this needed?

src/forge/controller/service.py
```diff
@@ -414,10 +345,6 @@ async def start_session(self) -> str:
 
         return sess_id
 
-    def session(self) -> SessionContext:
```
allenwang28, line 417: moved to ServiceInterface

src/forge/controller/interface.py
allenwang28: added a comment in the docstring, but a few reasons:
- primarily so you don't have to interact with the Service actor, which IMO can be annoying
- pairs the proc_mesh with the Service actor
- later we might want to pass a reference to the service to other actors, without moving where it's placed. This doubles as a handle so we can make calls to it

PR #70, opened 2025-08-25
Leverage Services for Policy + Rename PolicyRouter

Jack-Khuu:
_Recreation of https://github.com/meta-pytorch/forge/pull/61, since commit history got borked_

---

Leverage `spawn_services` for **Policy** and performs associated refactors 
- `PolicyRouter` -> `Policy`
  - Pushed Mesh and Worker set up into Policy (used to be Router).
- `Policy` -> `PolicyWorker`
 
---
```
python src/forge/actors/policy.py
```

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -9,7 +9,7 @@ import logging
 import os
 import sys
 from copy import copy
-from dataclasses import dataclass
+from dataclasses import asdict, dataclass
 from typing import Dict, List
 
 import torch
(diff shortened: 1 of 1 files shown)


Jack-Khuu:
Code-wise this is pretty close to what we will be using

Debugging an issue where `Policy.run_processing` doesn't get called the same with `spawn_service` as `spawn_actor`

src/forge/actors/policy.py
```diff
@@ -39,24 +45,43 @@
...
-
 logger = logging.getLogger(__name__)
 
 
 @dataclass
-class PolicyRouter(Actor):
+class Policy(PolicyInterface):
     # TODO: Add dp support
-    policy: Actor
+    config: DictConfig
```
joecummings: I don't like individual components having knowledge of our frontend config system. Can we parameterize this fully or pass in a policy specific config class?

src/forge/actors/policy.py
```diff
@@ -387,67 +428,49 @@ def get_default_sampling_params(vllm_config, overrides=None) -> SamplingParams:
...
     for batch, response in enumerate(responses):
         print("~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~")
         print(f"Batch {batch}:")
         print(f"User: {prompt}\nAssistant: {response.text}")
         print("\n~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~")
 
-    await router.shutdown.call()
-
 
 if __name__ == "__main__":
```
allenwang28, line 503: Can we move the policy main into its own app? apps/vllm.py or something 

I think having a standalone vLLM application will be useful for Stanford too

Jack-Khuu: Totally let's do it in a. separate PR to unblock this one

allenwang28:
lgtm, as long as it's running!

PR #73, opened 2025-08-26
Switch to lazy module imports

pbontrager:
Use lazy imports for modules so that you only import the the classes you actually use. For each submodule like actors, controller, data, etc you define the public apis at the top and then define a __getattr__ method to import them only when requested. This will make imports faster and avoid, avoid circular dependencies, and make optional dependencies easier to manage. At the app level, users would be expected to import

`from forge.actor import Policy` or `from forge.data import collate_packed`.

### DIFF
--- a/src/forge/actors/__init__.py
+++ b/src/forge/actors/__init__.py
@@ -4,6 +4,16 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from .collector import Collector
+__all__ = ["Policy", "PolicyRouter"]
 
-__all__ = ["Collector"]
+
+def __getattr__(name):
+    if name == "Policy":
+        from .policy import Policy
+
+        return Policy
+    if name == "PolicyRouter":
+        from .policy import PolicyRouter
+
+        return PolicyRouter
+    raise AttributeError("fmodule {__name__} has no attribute {name}")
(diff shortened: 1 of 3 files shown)


src/forge/actors/__init__.py
```diff
@@ -4,6 +4,16 @@
...
+def __getattr__(name):
+    if name == "Policy":
+        from .policy import Policy
+
+        return Policy
+    if name == "PolicyRouter":
+        from .policy import PolicyRouter
+
+        return PolicyRouter
+    raise AttributeError("fmodule {__name__} has no attribute {name}")
```
ebsmothers, line 19:
```suggestion
    raise AttributeError(f"module {__name__} has no attribute {name}")
```

pbontrager:
> While I definitely appreciate the motivation, I worry that this is introducing a maintenance nightmare down the line

How is this harder to maintain than what we had before? It's just more verbose, not more complex. I can implement the getattr method to grab the imports from a dictionary so future updates are only one line, but you said it looked too complex.

ebsmothers:
Oh if we're choosing between the two options I would prefer the other one. With this approach every new API added needs a corresponding else block in an `__init__` file.. seems like a very obvious footgun for contributors.

> > While I definitely appreciate the motivation, I worry that this is introducing a maintenance nightmare down the line
> 
> How is this harder to maintain than what we had before? It's just more verbose, not more complex. I can implement the getattr method to grab the imports from a dictionary so future updates are only one line, but you said it looked too complex.

joecummings:
See #83

PR #77, opened 2025-08-26
[RFC] StoreInterface

casteryh:
Summary:
This is a proposed interface that will be the backend of the replay buffer. I have included methods that would be useful from the point of view of making a replay buffer.

This also doubles as a proposal for the actual torchstore api.

Test Plan:
n/a

### DIFF
--- a/src/forge/interfaces.py
+++ b/src/forge/interfaces.py
@@ -5,13 +5,13 @@
 # LICENSE file in the root directory of this source tree.
 
 from abc import ABC, abstractmethod
-from typing import Any, Mapping
-
-from monarch.actor import endpoint
+from typing import Any, List, Mapping
 
 from forge.controller import ForgeActor
 
 from forge.types import Action, Message, Observation, Scalar, State
+
+from monarch.actor import endpoint
 
 
 class Transform(ABC):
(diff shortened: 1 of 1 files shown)


src/forge/interfaces.py
```diff
@@ -88,6 +91,99 @@ async def update_weights(self):
...
+    def __len__(self) -> int:
+        """Return the number of key-value pairs in the buffer.
+
+        Returns:
+            int: The number of items in the buffer.
+        """
+        pass
+
+    @abstractmethod
+    def __getitem__(self, key: K) -> V:
```
casteryh: Do we want async for this method (and the rest)?

src/forge/interfaces.py
```diff
@@ -88,6 +91,99 @@ async def update_weights(self):
...
+            Iterable[K]: An iterable containing all keys in the buffer.
+        """
+        pass
+
+
+class RawBuffer(BufferView[K, V], ABC):
+    """Abstract interface for the underlying storage backend (raw buffer) of a ReplayBuffer."""
+
+    @abstractmethod
+    def add(self, key: K, val: V) -> None:
```
casteryh: same question: async?

casteryh:
@LucasLLC

casteryh:
Also @kaiyuan-li

src/forge/interfaces.py
```diff
@@ -90,6 +90,200 @@ async def update_weights(self):
...
+    @abstractmethod
+    async def exists(self, key: str) -> bool:
+        """
+        Check if a key exists in the store.
+        """
+        pass
+
+    # TODO(yuxuanh): add this to torchstore.
+    @abstractmethod
+    async def release(self, key: str) -> None:
```
joecummings, line 171: Interesting idea: what's the inspiration here?

casteryh: Say the trainer is at step 10 and will no longer need stuff from step 5. A reasonable thing to do would be simply mark all keys starting with `replay_buffer.step_10` as released and move on, instead of waiting it to be actually deleted.

While from the torchstore side it's probably easier to implement this as instant deletion right now, it would be nice to have this semantics, for if and when we hit a scale where this matters.

casteryh: However, on the other hand, since everything is implemented in Python. It's probably fast enough to just delete instantly since we don't deallocate memory when deleting. Indeed, currently all keys are held by a single process Controller actor in torchstore right now - so it makes less sense to reinvent GC ourself.

Things do get complicated if we need to shard the controller. And it's much easier to just not make any promises.

In this regard, we should probably remove the `delete[_all]` methods all together, as it would be a nightmare to do it correctly in a distributed setting.

src/forge/interfaces.py
```diff
@@ -90,6 +90,200 @@ async def update_weights(self):
...
+            key (K): The key to release
+
+        Returns:
+            None
+        """
+        pass
+
+    # TODO(yuxuanh): add this to torchstore.
+    @abstractmethod
+    async def release_all(self, prefix: str) -> None:
```
DNXie, line 188: what's the difference between release and delete?

LucasLLC: And delete?

src/forge/interfaces.py
LucasLLC: There could be some perf gains here where we notify controller of delete and then let storage volumes garbage collect later.

LucasLLC:
I like the ideas here but in general I think as a practice I think we should only add what we need as it's needed and used. Great work!

src/forge/interfaces.py
DNXie: I meant the difference between two functions `release` and `delete`

PR #85, opened 2025-08-28
Integrate metric logger into grpo training loop

DNXie:
Integrate metric logger into grpo training loop with some basic logging

image


apps/grpo/main.py
```diff
@@ -511,7 +521,9 @@ async def continuous_training():
                 if training_step % 10 == 0:
                     print(f"Completed {training_step} training steps")
                     if training_result:
-                        print(f"Latest loss: {training_result.get('loss', 'N/A')}")
+                        loss_value = training_result.get("loss", 0.0)
+                        print(f"Latest loss: {loss_value}")
+                        logger.log("loss/training_step", loss_value, training_step)
```
felipemello1, line 526: You probably want to make sure that the input is a dictionary, like here: https://github.com/pytorch/torchtune/blob/67ab86b94de9e7ac7dd9850113ebe69e2bbd307c/recipes/full_finetune_distributed.py#L909

I think that we will have an abundance of metrics coming from dataset and reliability metrics. This is how i envisioned it being used: https://fb.workplace.com/groups/1189731669410969/permalink/1279384097112392/

I understand that its a a simple PR just to get logging started. Just sharing where I think we should land after a few iterations.

joecummings: Yeah great callout, but lets land this one for now and iterate to that point.

DNXie: @felipemello1 Thanks for the suggestions! This is a simple PR to simply integrate the logger. In my next PR, I will add logging for various metrics using log_dict. 
@joecummings Thanks for approval. I will go ahead and land this one for now.

PR #87, opened 2025-08-28
Integrate reward functions into grpo main

DNXie:
Integrate reward functions into grpo main


DNXie:
> You can and should delete the impl I put in grpo.main before merging

@joecummings Removed. Good catch! I will land this one.

PR #91, opened 2025-08-28
Enables dynamic GPU allocation for local workloads

allenwang28:
Context for this PR - I want to be able to demo a service that spawns N replicas of vLLM. For that to work, we need dynamic GPU allocation. Our current approach hard codes GPU assignment.

So this PR does a few things:
- Introduces `GpuResourceManager`, a singleton actor (or "controller") responsible for tracking and releasing GPUs. This only works for local proc mesh for now, will expand to multi-host once/if needed. Since it's a singleton, there's no concern about race conditions etc.
- Updates proc and service types to now track `num_gpus` too, which is the signal for receiving GPU ids
- Adds stubs for the new `hostmesh` APIs - currently broken for some NYI stuff in Monarch. Probably need to work through a re-design of procmesh management/allocation
- Updates applications (sft_v2, rl, grpo_main, vllm) accordingly

Next up:
- Create a policy specific Replica which will use `get_proc_mesh()` correctly

### DIFF
--- a/src/forge/controller/proc_mesh.py
+++ b/src/forge/controller/proc_mesh.py
@@ -75,8 +75,8 @@ async def get_proc_mesh(process_config: ProcessConfig) -> ProcMesh:
         if process_config.num_hosts != 1:
             raise ValueError("Local scheduler only supports 1 host")
 
-        if process_config.num_gpus > 0:
-            gpu_ids = await get_gpu_ids(process_config.num_gpus)
+        if process_config.with_gpus:
+            gpu_ids = await get_gpu_ids(process_config.num_procs)
             env["CUDA_VISIBLE_DEVICES"] = ",".join(map(str, gpu_ids))
 
         # TODO - update to use this_host() whenever it supports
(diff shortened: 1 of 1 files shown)


apps/grpo/main.py
```diff
@@ -393,21 +386,20 @@ async def main():
...
-        default_service_cfg,
+        ServiceConfig(procs_per_replica=1, num_replicas=1),
         ComputeAdvantages,
         gamma=0.99,
         lambda_=0.95,
     )
 
     ref_model = await spawn_service(
-        default_service_cfg,
+        ServiceConfig(procs_per_replica=1, num_replicas=1),
```
joecummings: This probably needs a GPU

src/forge/controller/__init__.py
joecummings, line 1: Not multi-submodule surfacing please. If something is in service, we should import from service.

apps/grpo/main.py
```diff
@@ -351,40 +351,33 @@ async def main():
...
             num_workers=1,
             worker_params=WorkerConfig(model=model),
             sampling_params=SamplingOverrides(num_samples=group_size, max_tokens=16),
-            available_devices="3",
         ),
     )
 
     trainer = await spawn_service(
-        default_service_cfg,
+        ServiceConfig(procs_per_replica=1, gpus_per_replica=1, num_replicas=1),
```
pbontrager: What would happen if procs_per_replica and gpus_per_replica didn't match? Would it make more sense to just have a boolean "gpus_procs"?

apps/grpo/main.py
allenwang28: fixed

apps/grpo/main.py
allenwang28: yeah good call, I just named it `with_gpus` for now

PR #92, opened 2025-08-28
RL Trainer GRPO Updates

pbontrager:
- Made RLTrainer's config setup no longer trash
- Removed ForgeEngine inheritance + model creation happens in setup now
- [TODO] match train_step to apps/grpo -> Followup PR


Jack-Khuu:
Stamp to unblock

Jack-Khuu:
Can you list which titan hash you're on btw?

PR #93, opened 2025-08-28
Support dp_size in replay buffer

DNXie:
Added a `dp_size` dimension in replay buffer sampling to enable data parallel.

Updated `GRPO/main.py` accordingly 

**Test:**
1. `pytest tests/unit_tests/test_replay_buffer.py`

```
 7 passed, 7 warnings in 23.35s
```

2.
`python -m apps.grpo.main --config apps/grpo/qwen3_1_7b.yaml`

```
All services initialized successfully!
Starting GRPO training loops...
Generated 10 rollouts w/ average reward 0.275
Completed 10 training steps
Latest loss: 0.01575610041618347
Generated 20 rollouts w/ average reward 0.05
Completed 20 training steps
Latest loss: 0.0045821815729141235
```

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -150,6 +150,7 @@ class Trainer(ForgeActor):
     beta: float = 0.1
     epsilon: float = 0.1
     device: torch.device | None = None
+    dp_rank: int = 0  # TODO: support data parallelism, hard code it for now
 
     @endpoint
     def setup(self):
@@ -178,6 +179,7 @@ class Trainer(ForgeActor):
 
     @endpoint
     async def train_step(self, batch: list[Episode]):
+        batch = batch[self.dp_rank]
         pad_id = batch[0].pad_id
 
         # prepare batch
@@ -447,7 +449,6 @@ async def main(cfg: DictConfig):
             if batch is None:
                 await asyncio.sleep(0.1)
             else:
-                batch = batch[0]  # Hard coded because we are not doing data parallel
                 training_result = await trainer.train_step.choose(batch)
                 training_step += 1
                 if training_step % 10 == 0:


src/forge/actors/replay_buffer.py
```diff
@@ -43,23 +44,32 @@ async def sample(self, curr_policy_version: int, batch_size: int | None = None):
...
+        idx_to_sample = self.sampler(range(len(self.buffer)), k=total_samples)
         sorted_idxs = sorted(
             idx_to_sample, reverse=True
         )  # Sort in desc order to avoid shifting idxs
         sampled_episodes = [self.buffer.pop(i) for i in sorted_idxs]
-        return sampled_episodes
+
+        # Reshape to (dp_size, bsz, ...)
+        reshaped_episodes = []
+        for dp_idx in range(self.dp_size):
```
pbontrager: We don't want to return a sorted sample here as that reduces variability in the sample. You need to get the index of the sorted array and then probably do this as a nested for loop to be easier to read.

```
batch = []
for rank in self.dp_size:
	local_batch = []
	for i in bsz:
		e = sampled_episodes[sort_order[rank*i]]
		local_batch.append(e)
	batch.append(local_batch)
```

DNXie: Thanks for pointing out this issue. I have updated this part. Please review.

tests/unit_tests/test_replay_buffer.py
```diff
@@ -105,6 +105,11 @@ async def test_sample_with_evictions(self, replay_buffer) -> None:
...
         )
         assert samples is not None
-        assert len(samples) == 1
-        assert samples[0] == trajectory_1
+        assert len(samples[0]) == 1
+        assert samples[0][0] == trajectory_1
         replay_buffer.clear.call_one().get()
+
+
+if __name__ == "__main__":
```
pbontrager: Instead of adding this, do `pytest tests/unit_tests/test_replay_buffer.py`

tests/unit_tests/test_replay_buffer.py
```diff
@@ -76,15 +76,15 @@ async def test_sample(self, replay_buffer) -> None:
         # Test a simple sampling w/ no evictions
```
pbontrager, line 76: Can you also test that len(samples) == dp_size?

src/forge/actors/replay_buffer.py
```diff
@@ -43,23 +44,31 @@ async def sample(self, curr_policy_version: int, batch_size: int | None = None):
...
-        sorted_idxs = sorted(
-            idx_to_sample, reverse=True
-        )  # Sort in desc order to avoid shifting idxs
-        sampled_episodes = [self.buffer.pop(i) for i in sorted_idxs]
-        return sampled_episodes
+        idx_to_sample = self.sampler(range(len(self.buffer)), k=total_samples)
+        sampled_episodes = [self.buffer[i] for i in idx_to_sample]
+
+        # Evict sampled episodes (descending order so pops are safe)
+        for i in sorted(idx_to_sample, reverse=True):
```
pbontrager: This is cleaner but is it moving the data twice? It's probably fine.

apps/grpo/main.py
```diff
@@ -447,6 +447,7 @@ async def continuous_training():
             if batch is None:
                 await asyncio.sleep(0.1)
             else:
+                batch = batch[0]  # Hard coded because we are not doing data parallel
```
pbontrager: Update the trainer instead to accept the batch and then take batch = batch[self.dp_rank]

tests/unit_tests/test_replay_buffer.py
DNXie: Added a test case.

src/forge/actors/replay_buffer.py
DNXie: I have updated the code to make it more efficient.

apps/grpo/main.py
DNXie: Updated.

PR #94, opened 2025-08-29
Adds TitanRefModel in place of HF based Reference Model

Jack-Khuu:
Replaces the existing HF based RefModel with a torchtitan backed `TitanRefModel`

Note: This PR persists the old implementation using `AutoModelForCausalLM` until GRPO is ready for full migration to torchtitan

---

`python apps/grpo/main.py`

Example Output
```
Generated 60 rollouts w/ average reward 0.0
Generated 70 rollouts w/ average reward 0.1
Generated 80 rollouts w/ average reward 0.1
Completed 20 training steps
Latest loss: 75.34173269197345
Generated 90 rollouts w/ average reward 0.1
Generated 100 rollouts w/ average reward 0.0
Generated 110 rollouts w/ average reward 0.1
Generated 120 rollouts w/ average reward 0.1
Completed 30 training steps
Latest loss: 32.164904564619064
Generated 130 rollouts w/ average reward 0.0
Generated 140 rollouts w/ average reward 0.1
Generated 150 rollouts w/ average reward 0.1
Generated 160 rollouts w/ average reward 0.1
Completed 40 training steps
Latest loss: 18.617477774620056
Generated 170 rollouts w/ average reward 0.1
Generated 180 rollouts w/ average reward 0.1
Generated 190 rollouts w/ average reward 0.1
Generated 200 rollouts w/ average reward 0.1
Completed 50 training steps
Latest loss: 9.065430700778961

```

### DIFF
--- a/apps/rl/llama3_8b.yaml
+++ b/apps/rl/llama3_8b.yaml
@@ -18,7 +18,7 @@ trainer:
   processes:
     scheduler: local # local | mast (not supported yet)
     num_hosts: 1
-    num_gpus: 4
+    with_gpus: True
     num_procs: 4
 
   optimizer:
@@ -68,7 +68,7 @@ replay_buffer:
   processes:
     scheduler: local # local | mast (not supported yet)
     num_hosts: 1
-    num_gpus: 0
+    with_gpus: False
     num_procs: 1
 
 # policy:
--- a/src/forge/controller/service/__init__.py
+++ b/src/forge/controller/service/__init__.py
@@ -7,7 +7,7 @@
 from .interface import ServiceInterface, Session, SessionContext
 from .metrics import ServiceMetrics
 from .replica import Replica, ReplicaMetrics
-from .service import Service, ServiceConfig
+from .service import Service, ServiceActor, ServiceConfig
 from .spawn import shutdown_service, spawn_service
 
 __all__ = [
@@ -19,6 +19,7 @@ __all__ = [
     "ServiceMetrics",
     "Session",
     "SessionContext",
+    "ServiceActor",
     "spawn_service",
     "shutdown_service",
 ]
--- a/src/forge/controller/service/interface.py
+++ b/src/forge/controller/service/interface.py
@@ -87,6 +87,35 @@ class ServiceEndpoint(Generic[P, R]):
 
     """
 
+    def __init__(self, service, endpoint_name: str):
+        self.service = service
+        self.endpoint_name = endpoint_name
+
+    async def choose(self, *args: P.args, **kwargs: P.kwargs) -> R:
+        """Chooses a replica to call based on context and load balancing strategy."""
+        # Extract sess_id from kwargs if present
+        sess_id = kwargs.pop("sess_id", None)
+        return await self.service._call(sess_id, self.endpoint_name, *args, **kwargs)
+
+    async def call(self, *args: P.args, **kwargs: P.kwargs) -> List[R]:
+        """Broadcasts a request to all healthy replicas and returns the results as a list."""
+        result = await self.service.call_all(self.endpoint_name, *args, **kwargs)
+        return result
+
+
+class ServiceEndpointV2(Generic[P, R]):
+    """An endpoint object specific to services.
+
+    This loosely mimics the Endpoint APIs exposed in Monarch, with
+    a few key differences:
+    - Only choose and call are retained (dropping stream and call_one)
+    - Call returns a list directly rather than a ValueMesh.
+
+    These changes are made with Forge use cases in mind, but can
+    certainly be expanded/adapted in the future.
+
+    """
+
     def __init__(self, actor_mesh, endpoint_name: str):
         self.actor_mesh = actor_mesh
         self.endpoint_name = endpoint_name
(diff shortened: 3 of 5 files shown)


src/forge/actors/reference_actor.py
```diff
@@ -0,0 +1,204 @@
...
+from transformers import AutoModelForCausalLM
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+@dataclass
+class ReferenceActor(ForgeActor):
+    model: Model = field(default_factory=Model)
```
pbontrager: You shouldn't have ForgeActor here. Just all the titan data class types you need, then in setup you can build ForgeJobConfig and have that as the only thing you use to spawn the reference worker. The engine should be in the worker with no HF imports.

src/forge/actors/reference_actor.py
```diff
@@ -0,0 +1,309 @@
...
+        if parallel_dims.pp_enabled:
+            raise NotImplementedError("PP not implemented yet")
+        else:
+            # Non-PP forward / backward
+            with self.engine.train_context(optional_context_parallel_ctx):
+                assert len(model_parts) == 1
+                with self.engine.maybe_enable_amp:
+                    pred = model_parts[0](input_ids)
+
+        # TODO: Update compute_sequence_logprobs to convert probs (logits) to logprobs
```
pbontrager: When you get to this, use [this](https://github.com/pytorch/torchtune/blob/67ab86b94de9e7ac7dd9850113ebe69e2bbd307c/recipes/dev/grpo_full_finetune_distributed.py#L627) as the correct reference (with an understanding that you're using Titan's model instead of tune's and there are no utils).

src/forge/actors/reference_actor.py
```diff
@@ -0,0 +1,309 @@
...
+        }
+        os.environ.update(env)
+
+    @endpoint
+    async def setup(self):
+        engine_config = {f.name: getattr(self, f.name) for f in fields(self)}
+        self.engine = ForgeEngine(ForgeJobConfig(**engine_config))
+
+    @endpoint
+    async def forward(self, token_ids: list[int]) -> torch.Tensor:
```
pbontrager: We can keep things bs=1 for now, but we're going to need to add our own batching solution where we take n requests off the queue at a time and process them as a batch.

Jack-Khuu: The batching itself is an easy add using the queue in ReferenceActor (the one in the PR that's unused) or we can surface it as a common API in Replica (Just `processes_single_request` atm)

apps/grpo/main.py
```diff
@@ -401,10 +336,16 @@ async def main():
...
+        #     RefModel,
+        #     model_name=model,
+        # ),
+        # GOAL: Swap this in and everything should just "work"
         spawn_service(
             ServiceConfig(procs_per_replica=1, num_replicas=1, with_gpus=True),
-            RefModel,
-            model_name=model,
+            TitanRefModel,
+            # model=TitanJobModelConfig(name=model),
```
Jack-Khuu: Commentary: Getting this to work is the goal

src/forge/actors/reference_actor.py
```diff
@@ -0,0 +1,382 @@
...
+        if parallel_dims.pp_enabled:
+            raise NotImplementedError("PP not implemented yet")
+        else:
+            # Non-PP forward
+            # (jackkhuu) I don't think iether context are needed for inference here
+            with self.engine.train_context(optional_context_parallel_ctx):
+                assert len(model_parts) == 1
+                with self.engine.maybe_enable_amp:
+                    # Titan Tranformer
+                    pred = model_parts[0](input_ids)
```
Jack-Khuu: Commentary: This is where we enter titan's transformer and the inference fails 

See P1926047471 for snippet of the 
```
 /mnt/code/pytorch/aten/src/ATen/native/cuda/IndexKernelUtils.cu:16: vectorized_gather_kernel: block: [8,0,0], thread: [32,0,0] Assertion `ind >=0 && ind < ind_dim_size && "vectorized gather kernel index out of bounds"` failed.
``` 
Error

src/forge/actors/policy.py
```diff
@@ -310,7 +310,7 @@ async def run(self):
             for request_output in processed_outputs.request_outputs:
                 if request_output.finished:
                     _, fut = self.requests.pop(request_output.request_id)
-                    fut.set_result(request_output.outputs)
+                    fut.set_result(request_output)
```
Jack-Khuu, line 313: Adopted from https://github.com/meta-pytorch/forge/pull/97

src/forge/actors/__init__.py
```diff
@@ -24,5 +24,9 @@ def __getattr__(name):
         from .replay_buffer import ReplayBuffer
 
         return ReplayBuffer
+    elif name == "TitanRefModel":
```
joecummings, line 27: 🙃

src/forge/actors/policy.py
joecummings: Why this instead of raw outputs?

Jack-Khuu: Pragmatically: Less merge conflict with Philip's PR

I don't have strong preference, but it does make the output self contained which is nice when we need to pass the results around

src/forge/actors/reference_actor.py
```diff
@@ -0,0 +1,351 @@
...
+from torchtitan.experiments.forge.job_config import ForgeJobConfig
+from transformers import AutoModelForCausalLM
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+@dataclass
+class TitanRefModel(ForgeActor):
```
pbontrager, line 39: On naming: I would just call this ReferenceModel, we don't use Titan anywhere else and we havn't generalized it yet

src/forge/actors/reference_actor.py
```diff
@@ -0,0 +1,351 @@
...
+    """
+    Represents a reference actor leveraging a torchtitan model for execution
+
+    Intended for generating reference_logprobs - for example in KL Divergence
+    """
+
+    # Refer to titan JobConfig for enabling more ForgeEngine configuration
+    model: Model = field(default_factory=Model)
+    parallelism: Parallelism = field(default_factory=Parallelism)
```
pbontrager, line 48: I feel like we would need Checkpoint and Compile too. Or is checkpoint not needed for loading?

src/forge/actors/reference_actor.py
```diff
@@ -0,0 +1,351 @@
...
+                cp_rotate_method=self.job_config.parallelism.context_parallel_rotate_method,
+            )
+            if parallel_dims.cp_enabled
+            else None
+        )
+
+        if parallel_dims.pp_enabled:
+            raise NotImplementedError("PP not implemented yet")
+        else:
+            # (jackkhuu) Not sure if either context are needed for inference here
```
pbontrager, line 123: Can you separate out the request call from forward? Also look at the updated RefActor in main to make sure they're in sync

src/forge/actors/reference_actor.py
```diff
@@ -0,0 +1,351 @@
...
+        ).squeeze(-1)
+
+        # Sum log probabilities across sequence (masked by attention)
+        shifted_attention_mask = attention_mask[:, 1:]
+        sequence_log_probs = (token_log_probs * shifted_attention_mask).sum(dim=-1)
+
+        return sequence_log_probs
+
+
+"""
```
pbontrager, line 233: Can we comment out everything below this?

src/forge/actors/reference_actor.py
```diff
@@ -0,0 +1,351 @@
...
+        sequence_log_probs = compute_sequence_logprobs(
+            self.model, input_ids, attention_mask, requires_grad=False
+        )
+
+        return (
+            sequence_log_probs.squeeze()
+        )  # Remove batch dimension for single response
+
+
+def compute_sequence_logprobs(
```
pbontrager, line 202: You should update this to match the one from main. This should probably go in the trainer.py file

src/forge/actors/reference_actor.py
Jack-Khuu: Not needed by default, [Checkpointing](https://github.com/pytorch/torchtitan/blob/52a4c17f4423ba60f48177a15b29d5046d6d3876/torchtitan/config/job_config.py#L397) is used when we want to write out or load from a checkpoint

Former doesn't happen with the reference model, the latter does unlock custom ref models (or using checkpoints as ref models)

src/forge/actors/reference_actor.py
Jack-Khuu: This was copy paste from main (at the time Joe's)

`compute_logprobs` is the file is the implementation from https://github.com/meta-pytorch/forge/pull/97

Jack-Khuu:
Follow up changes in https://github.com/meta-pytorch/forge/pull/118

PR #95, opened 2025-08-29
skeleton code for torchstore integration of forge learner.

pradeepfn:
1. For initial feedback around.
2. TODO items listed in the order of priority.
3. Will add test script.

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
+import asyncio
 import logging
 import math
 import os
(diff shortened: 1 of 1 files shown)


src/forge/actors/trainer.py
```diff
@@ -68,6 +71,10 @@ def __init__(self, config: DictConfig):
         self.gradient_accumulation_steps = 1  # Example value, adjust as needed
         self._rank = current_rank().rank
         self._size = math.prod(current_size().values())
+
+        # init torchstore
+        self._tstore = asyncio.run(MultiProcessStore.create_store())
```
joecummings, line 76: Do you need the asyncio.run?

casteryh: > Do you need the asyncio.run?

This is because `__init__` is not async?

PR #96, opened 2025-08-29
Adds support for custom replicas, updates policy to spawn correctly within `spawn_service`

allenwang28:
This PR does a few things:
- Moves `Replica`'s `create_proc_mesh`, `spawn_actors`, and `stop` functionality into `ForgeActor` as `@classmethod` for `launch` and `shutdown`. Why?
  - This couples definition of how an actor should be launched with the actor def itself, rather than being in a separate object.
  - This gives flexibility for more complex actors, like `Policy`, which spawns multiple proc meshes and actor types  
- Uses `ForgeActor` for everything we expect to be spawned as a service
  - GPUManager becomes a regular actor to avoid circular imports (it doesn't really need to be a ForgeActor either)
- Modifies `Policy` to pick up these changes 
- GRPO example mods:
  - Dataset adds in `**kwargs`, `launch()` doesn't play well with `*args` and I couldn't figure out the exact right way to make args work correctly. Therefore `spawn_service` really only accepts `kwargs` at the moment. 
  - Spawns and shutdowns all services at the same time to reduce the initialization time
- vLLM example changes:
  - Use the `with policy.session` context manager and generally QoL updates

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -94,187 +94,13 @@ class PolicyConfig:
 
 
 @dataclass
-class PolicyWorker(ForgeActor):
-    model: str
-    tensor_parallel_size: int = 1
-    pipeline_parallel_size: int = 1
-    enforce_eager: bool = False
-    vllm_args: EngineArgs = None
-    state_dict_key: str = "model_state_dict"
-
-    def __post_init__(self):
-        """Build vLLM Arguments
-
-        vLLM specific TODOS
-        - output format
-        - check_health
-        - _aggregate workers output
-        - register_failure_callback
-
-        Testing
-        - all LLM generate methods, verify against LLM inputs
-        - all executor methods verify no changes
-        """
-        if self.vllm_args is None:
-            # Use default vllm EngineArgs
-            self.vllm_args = EngineArgs(
-                model=self.model,
-                tensor_parallel_size=self.tensor_parallel_size,
-                pipeline_parallel_size=self.pipeline_parallel_size,
-                enforce_eager=self.enforce_eager,
-            )
-            # Original method returns False when not run in the main thread
-            self.vllm_args._is_v1_supported_oracle = lambda *_: True
-        else:
-            # Check that provided args match Policy args
-            cfg = [
-                "model",
-                "tensor_parallel_size",
-                "pipeline_parallel_size",
-                "data_parallel_size",
-            ]
-            for key in cfg:
-                value = getattr(self, key) if key != "data_parallel_size" else 1
-                if getattr(self.vllm_args, key) != value:
-                    logger.warning(
-                        f"{key} args don't match value in EngineArgs, overriding with {value}"
-                    )
-                    setattr(self.vllm_args, key, value)
-        # Build Config
-        self.vllm_args = self.vllm_args.create_engine_config(UsageContext.LLM_CLASS)
-
-    @endpoint
-    async def setup(self, store: MultiProcessStore = None):
-        self.torchstore = store
-        # TODO: remove ["gpus"] when monarch implements a flat rank
-        self.rank = current_rank()["gpus"]
-        self.worker = self.setup_worker()
-
-    @endpoint
-    async def execute_model(self, schedule: SchedulerOutput):
-        return self.worker.execute_model(schedule)
-
-    async def _load_tensor_parallel_state_dict(self, current_state_dict: dict):
-        """
-        Load full state dict from torchstore into tensor parallel model with deterministic sharding.
-        """
-
-        updated_count = 0
-        # setting explictly to llama3 for now as its our only use case
-        sharding = VLLMSharding(self.tensor_parallel_size, self.rank)
-
-        for param_name in current_state_dict.keys():
-            current_tensor = current_state_dict[param_name]
-
-            # Load the full tensor from torchstore
-            # TODO: only get the part of the tensor that is needed
-            stored_tensor = await self.torchstore.get(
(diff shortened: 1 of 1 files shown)


apps/vllm/main.py
```diff
@@ -72,35 +74,34 @@ def get_configs(args: Namespace) -> (PolicyConfig, ServiceConfig):
...
+    )
+    service_config = ServiceConfig(
+        procs_per_replica=worker_size, num_replicas=1, with_gpus=True
     )
-    service_config = ServiceConfig(procs_per_replica=1, num_replicas=1)
 
     return policy_config, service_config
 
 
 async def run_vllm(service_config: ServiceConfig, config: PolicyConfig, prompt: str):
```
joecummings, line 86: Can't be delete this now that we have the vllm app?

apps/grpo/main.py
joecummings, line 1: Is it out of scope to move the vllm processing loop within the policy instead of having the start that in main.py?

src/forge/controller/system_controllers/gpu_manager.py
```diff
@@ -8,15 +8,13 @@
...
-
-from forge.controller import ForgeActor
+from monarch.actor import Actor, ActorError, endpoint, get_or_spawn_controller
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
 
-class GpuManager(ForgeActor):
+class GpuManager(Actor):
```
joecummings, line 17: Why don't we want this to be a ForgeActor?

src/forge/controller/service/spawn.py
```diff
@@ -8,37 +8,51 @@
...
 from forge.controller.service.interface import ServiceInterface
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.DEBUG)
+logger.setLevel(logging.INFO)
 
 
 async def spawn_service(
-    service_cfg: ServiceConfig, actor_def: Type[Actor], *actor_args, **actor_kwargs
+    service_cfg: ServiceConfig, actor_def: Type[ForgeActor], **actor_kwargs
```
joecummings, line 23: Hmmm there will be things that likely need *args, not just **kwargs - is there a long term plan to make that possible?

src/forge/actors/policy.py
```diff
@@ -454,6 +267,234 @@ def setup_worker(self):
...
 
+@dataclass
+class Policy(PolicyInterface):
+    config: PolicyConfig
+    # Gets set up by setup
+    sampling_params: SamplingParams | None = None
+    lora_request: LoRARequest | None = None
+    tokenization_kwargs: dict = field(default_factory=dict)
+
+    _run_task: asyncio.Task | None = None
```
pbontrager: nit: the fields in the dataclass are basically the user parameters and show the user what they can change. Any extra variables, especially internal variables should be added when they're call in setup or in __post_init__

apps/grpo/main.py
allenwang28: ? yeah it should have been moved in this PR

src/forge/controller/system_controllers/gpu_manager.py
allenwang28: Circular import since `proc_mesh.py` wants to use `GpuManager`, and `ForgeActor` uses `proc_mesh` :/ can be re-arranged later

src/forge/actors/policy.py
```diff
@@ -454,6 +267,234 @@ def setup_worker(self):
         return worker
 
 
+@dataclass
+class Policy(PolicyInterface):
```
Jack-Khuu, line 97: nit: tad more difficult to see deltas since doesn't track when PolicyWorker/Policy switched spots

Anything to call out explicitly?

apps/vllm/main.py
allenwang28: hmm I'm not sure I'm following, we still need a ServiceConfig for spawning a service here regardless?

src/forge/actors/policy.py
allenwang28: good call, fixed

src/forge/actors/policy.py
```diff
@@ -95,24 +97,73 @@ class PolicyConfig:
...
-    policy_worker: Actor = None
+    sampling_params: SamplingParams | None = None
+    lora_request: LoRARequest | None = None
+    tokenization_kwargs: dict = field(default_factory=dict)
+    policy_worker: "PolicyWorker" = None
+
+    def __post_init__(self):
+        self._run_task: asyncio.Task | None = None
+        self._policy_proc: ProcMesh | None = None
+        self._worker_procs: ProcMesh | None = None
```
Jack-Khuu, line 108: nice

apps/grpo/main.py
```diff
@@ -351,66 +355,69 @@ async def main():
...
-            sampling_params=SamplingOverrides(num_samples=group_size, max_tokens=16),
+    (
+        dataloader,
+        policy,
+        trainer,
+        replay_buffer,
+        compute_advantages,
+        ref_model,
+        reward_actor,
+    ) = await asyncio.gather(
```
Jack-Khuu, line 366: Nice optimization

Will note that this is harder to read and more error prone if the order gets changed or services list mutated. Not sure if there's a way to get our cake and eat it too

allenwang28: hmm we could do something like:

```
dataloader_task = spawn_service(...) # don't await yet
policy_task = spawn_service(...)
```

then do the bulk await at the end:
```
dataloader, policy = await asyncio.gather(...)
```

but I'm not sure if it fully solves the problem

PR #98, opened 2025-08-29
Adds scripts and instructions for installing Forge from source.

allenwang28:
Took commands that worked and created two workflows:
1. Building wheels
2. Installing wheels

Note that vLLM wheel is too big (~680MB) to fit on GitHub, so I created a release in Forge to support it...

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -132,7 +132,7 @@ main() {
     pip install "$WHEEL_DIR"/*.whl
 
     log_info "Installing Forge from source..."
-    pip install .
+    pip install -e .
 
     # Set up environment
     log_info "Setting up environment..."


scripts/install.sh
```diff
@@ -0,0 +1,206 @@
...
+    download_vllm_wheel
+
+    log_info "Installing PyTorch nightly..."
+    pip install torch==$PYTORCH_VERSION --index-url https://download.pytorch.org/whl/nightly/cu129
+
+    log_info "Installing all wheels (local + downloaded)..."
+    pip install "$WHEEL_DIR"/*.whl
+
+    log_info "Installing Forge from source..."
+    pip install .
```
pbontrager: This should be "pip install -e .[dev]?

pradeepfn: The script works as it is. However, we need dev mode pip install, to iterative development. Good to have that as the default mode of installation.

scripts/build_wheels.sh
```diff
@@ -0,0 +1,347 @@
...
+    # Save to file for persistence
+    cat > ~/.forge_cuda_env << 'EOF'
+export CUDA_VERSION=12.9
+export NVCC=/usr/local/cuda-${CUDA_VERSION}/bin/nvcc
+export CUDA_NVCC_EXECUTABLE=/usr/local/cuda-${CUDA_VERSION}/bin/nvcc
+export CUDA_HOME=/usr/local/cuda-${CUDA_VERSION}
+export PATH="${CUDA_HOME}/bin:$PATH"
+export CUDA_INCLUDE_DIRS=$CUDA_HOME/include
+export CUDA_CUDART_LIBRARY=$CUDA_HOME/lib64/libcudart.so
+export LD_LIBRARY_PATH=/usr/local/cuda-12.9/compat:${LD_LIBRARY_PATH:-}
```
pradeepfn, line 139: Lets put the conda installed libs in the ld path; otherwise it complains for dynamic libs including python-3.10 libs.

export LD_LIBRARY_PATH=${CONDA_PREFIX}/lib:/usr/local/cuda-12.9/compat:${LD_LIBRARY_PATH:-}

PR #100, opened 2025-08-30
Fix infinite recursion when _service is not set

casteryh:
Summary:
To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service, I must find _service. To find _service...
Stack overflow. Maybe I should look somewhere else.

### DIFF
--- a/src/forge/controller/service/interface.py
+++ b/src/forge/controller/service/interface.py
@@ -179,7 +179,6 @@ class ServiceInterface:
 
     def __getattr__(self, name: str):
         """Forward all other attribute access to the underlying Service Actor."""
-        # No infinite recursion please
         try:
             _service = object.__getattribute__(self, "_service")
         except AttributeError:


src/forge/controller/service/interface.py
```diff
@@ -179,10 +179,16 @@ def _get_internal_state(self):
 
     def __getattr__(self, name: str):
         """Forward all other attribute access to the underlying Service Actor."""
+        # No infinite recursion please
```
allenwang28: we can remove this line

allenwang28:
```suggestion
```

casteryh:
> thanks for the fix, but how was this triggered?

I don't know exactly but I think it was triggered during some sort of error recovery of an actor, when apparently there were uncaught errors thrown inside the endpoint of the actor.

allenwang28:
> I think it was triggered during some sort of error recovery of an actor, when apparently there were uncaught errors thrown inside the endpoint of the actor.

hmm ok, if you see that again please let me know!

PR #102, opened 2025-08-31
Check in prototype notebook

allenwang28:
Checks in a notebook for our initial prototype!

### DIFF
--- a/demo.ipynb
+++ b/demo.ipynb
@@ -536,7 +536,7 @@
     "Let's build a step-by-step synchronous training loop to see how these services work together. The basic RL cycle is:\n",
     "\n",
     "1. Collect Experience: Get a prompt, generate a response, evaluate the reward\n",
-    "2. Compute Advantages: Calculate how much better/worse each action was than expected\n",
+    "2. Compute Rewards: Calculate how much better/worse each action was than expected\n",
     "3. Store Experience: Add the episode to our replay buffer\n",
     "4. Sample & Train: Get a batch of experiences and update the policy\n",
     "5. Repeat: Continue this cycle to improve the policy\n"
@@ -668,7 +668,8 @@
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3"
+   "pygments_lexer": "ipython3",
+   "version": "3.10.18"
   }
  },
  "nbformat": 4,


demo.ipynb
```diff
@@ -0,0 +1,1193 @@
...
+    "procs: ProcMesh = this_host().spawn_procs(per_host={\"gpus\": 8})\n",
+    "counters: Counter = procs.spawn(\"counters\", Counter, 0)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "202a8126-2e45-4074-a2ba-87e12d2f06dc",
+   "metadata": {},
+   "source": [
+    "### Broadcasting Messages\n",
```
pbontrager, line 143: Is this too in the weeds for what we want to show users? I don't think we want to expose all of Monarch to people do we? We should be simplifying abstractions as we go.

demo.ipynb
```diff
@@ -0,0 +1,1193 @@
...
+    "\n",
+    "Our mission is to accelerate innovation in reinforcement learning by empowering researchers and developers to explore new RL algorithms and infrastructure techniques. Whether you're designing novel training methods or optimizing distributed systems, Forge provides a foundation to build upon."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "24de9912-ed10-4729-9616-2f85bbf64e43",
+   "metadata": {},
+   "source": [
+    "## Brief Intro to Monarch\n",
```
pbontrager, line 29: I'm uncertain about this section. It covers actors well but then we're basically asking users to use services in place of actors. My biggest concern with this doc right now is that it covers so many infra concepts that people won't make the connection to why this matters for RL. But maybe hearing the last part presented would help.

demo.ipynb
```diff
@@ -0,0 +1,1193 @@
...
+    "## Building the RL Training Loop\n",
+    "### The Role of RL in Post-Training\n",
+    "One way to interpret the role of RL in post-training is to align a base pre-trained model towards hard-to-define targets. The goal is \"sampling\" the right data that we think will best align the model.\n",
+    "\n",
+    "This is the role of \"rollouts\" - creating the dataset used to update our policy. Rather than training on a static dataset, RL dynamically generates training data by having the current policy interact with the environment.\n",
+    "\n",
+    "Let's build a step-by-step synchronous training loop to see how these services work together. The basic RL cycle is:\n",
+    "\n",
+    "1. Collect Experience: Get a prompt, generate a response, evaluate the reward\n",
+    "2. Compute Advantages: Calculate how much better/worse each action was than expected\n",
```
pbontrager: This should be compute rewards as advantages are algorithm specific

demo.ipynb
```diff
@@ -0,0 +1,1193 @@
...
+   "source": [
+    "The response quality isn't great (it's only a 1.7B model!), but the service infrastructure is working perfectly.\n",
+    "\n",
+    "## Building the RL Training Loop\n",
+    "### The Role of RL in Post-Training\n",
+    "One way to interpret the role of RL in post-training is to align a base pre-trained model towards hard-to-define targets. The goal is \"sampling\" the right data that we think will best align the model.\n",
+    "\n",
+    "This is the role of \"rollouts\" - creating the dataset used to update our policy. Rather than training on a static dataset, RL dynamically generates training data by having the current policy interact with the environment.\n",
+    "\n",
+    "Let's build a step-by-step synchronous training loop to see how these services work together. The basic RL cycle is:\n",
```
pbontrager, line 536: This loop is too generic I think, it doesn't make it obvious why some random off the shelf services, like found in other libraries couldn't work instead.

1. Collect Experience: run tasks, tools, and mutli-step workflows
2. Compute Rewards: run verifiers and judges to calculate rewards
3. Store Experience: Add the episode to our replay buffer
4. Sample & Train: Sample a batch and update the policy in parallel
5. Broadcast Policy: Overlapped policy broadcast for services to update from
6. Repeat: Continue this cycle to improve the policy

This maybe goes too far into incomplete features, but maybe something in between

demo.ipynb
allenwang28: discussed offline, agreed on it for public release but for immediate milestones it should be fine

demo.ipynb
allenwang28: fixed

demo.ipynb
allenwang28: will change by the public release

PR #107, opened 2025-09-01
Add torchstore to policy

joecummings:
```python
spawn_service(
            ServiceConfig(procs_per_replica=1, with_gpus=True, num_replicas=1),
            Policy,
            config=PolicyConfig(
                worker_params=WorkerConfig(model=model),
                sampling_params=SamplingOverrides(
                    num_samples=group_size, max_tokens=16
                ),
            ),
            store=await MultiProcessStore.create_store(),
        ),
```

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -326,8 +326,9 @@ class Policy(PolicyInterface):
         futures = [fut for _, fut in self.requests.values()]
         if futures:
             await asyncio.gather(*futures)
-        await self.policy_worker.update.call()
-        self.weights_version += 1
+        new_version = self.weights_version + 1
+        await self.policy_worker.update.call(version=new_version)
+        self.weights_version = new_version
         return self.weights_version
 
     @endpoint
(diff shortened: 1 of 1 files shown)


src/forge/actors/policy.py
```diff
@@ -313,9 +320,20 @@ async def run(self):
                     fut.set_result(request_output.outputs)
 
     @endpoint
-    async def update_weights(self):
+    async def update_weights(self) -> int:
         """Update the policy weights."""
-        pass
+        # Wait for all current requests to finish, then publish model weights
```
pbontrager, line 325: Should there be a check that the new version exists on the store?

src/forge/actors/policy.py
```diff
@@ -101,18 +101,21 @@ class Policy(PolicyInterface):
...
         self._worker_procs: ProcMesh | None = None
+        self.weights_version: int = 0
 
     @classmethod
     async def launch(  # pyright: ignore[reportIncompatibleMethodOverride]
         cls: type["Policy"],
         *,
         process_config: ProcessConfig,
         config: PolicyConfig,
+        store: MultiProcessStore | None = None,
```
pbontrager, line 118: Does spawn services need to know to pass this in?

joecummings:
> I'm just not sure from this how store is getting passed in but otherwise looks good

Yeah, the launch logic IMO is really confusing - took me awhile to figure out why store was not being passed into the policy worker. This is a consequence of the `launch` vs init params of the service. Worth a discussion later as to which is preferred. Having both right now is just confusing.

src/forge/actors/policy.py
joecummings: Yes, it has to be passed in from the top level so you can use the same for Trainer.

src/forge/actors/policy.py
joecummings: I was just thinking that it would fail when trying to lookup the key and that would be obvious enough for now, but I could make it more explicit.

PR #115, opened 2025-09-02
VLLM + torchstore integration test upgrade to match the new Policy APIs.

pradeepfn:
Policy engine introduce a Policy service. This broke the existing vllm + torchstore example script.
This diff fixes it. 

1. Only the single worker case is fixed here.
2 Will fix the multi-worker case in followup PR.

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -332,9 +332,9 @@ class Policy(PolicyInterface):
         return self.weights_version
 
     @endpoint
-    async def get_model_params(self) -> Dict[str, torch.Tensor]:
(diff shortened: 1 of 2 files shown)


src/forge/actors/policy.py
```diff
@@ -331,6 +331,12 @@ async def update_weights(self) -> int:
         self.weights_version = new_version
         return self.weights_version
 
+    @endpoint
+    async def get_model_params(self) -> Dict[str, torch.Tensor]:
```
joecummings: If it's only for testing purposes, can you underscore it please?

PR #116, opened 2025-09-02
Add YAML-based configuration support for vLLM main

DNXie:
- Support vllm/main to take configs from yaml file
- Support grpo/main to take configs from yaml file (Removed from this PR, in https://github.com/meta-pytorch/forge/pull/141)
- Remove `PolicyConfig`
- Make `WorkerConfig` inherit `EngineArgs` and added `from_dict` to `WorkerConfig`.
- Rename `WorkerConfig` to `EngineConfig`
- Rename `SamplingOverrides` to `SamplingConfig`
- Updated Policy `__post_init__` to use `from_dict` when a dict is passed.
- Added unit tests for config file reading.  

## Test Run vllm/main:
```
export HF_HUB_DISABLE_XET=1
python -m apps.vllm.main --config apps/vllm/config.yaml
```
```
Requesting generation...

Generation Results:
================================================================================
Sample 1:
User: Tell me a joke
Assistant: . I need a laugh.
Here's one: A man walked into a library
--------------------------------------------------------------------------------
Sample 2:
User: Tell me a joke
Assistant:
I'll try to come up with one. Why did the scarecrow win
--------------------------------------------------------------------------------

Shutting down..
```

## Test Run grpo/main:
```
python -m apps.grpo.main --config apps/grpo/llama3_8b.yaml
```
```
...
Generated 10 rollouts w/ average reward 0.0
Generated 20 rollouts w/ average reward 0.0
Generated 30 rollouts w/ average reward 0.1
Generated 40 rollouts w/ average reward 0.1
Generated 50 rollouts w/ average reward 0.0
Completed 10 training steps
Latest loss: 114.34945392608643
Generated 60 rollouts w/ average reward 0.0
Generated 70 rollouts w/ average reward 0.0
Generated 80 rollouts w/ average reward 0.0
Generated 90 rollouts w/ average reward 0.0
...
```

## Unit test

```
python forge/tests/unit_tests/test_policy_config.py
```
```
Monkey patched Triton's _build! See /home/dxie/.fbpkg_conda_envs/forge-af45115/lib/python3.10/site-packages/patch_triton.py
Monkey patched Triton's nvsmi! See /home/dxie/.fbpkg_conda_envs/forge-af45115/lib/python3.10/site-packages/patch_triton.py
INFO 09-02 20:38:19 [__init__.py:235] Automatically detected platform cuda.
....
----------------------------------------------------------------------
Ran 4 tests in 0.002s

OK
```

## Test `vllm_args`
I tested with `vllm_args`
- string (see test case test_invalid_worker_config_from_dict)
- null (see the config file)
- with parameters values (see test case test_policy_yaml_config_loading)

### DIFF
--- a/tests/unit_tests/test_policy_config.py
+++ b/tests/unit_tests/test_policy_config.py
@@ -7,16 +7,31 @@
 import tempfile
 import unittest
 
+import pytest
 import yaml
 
-from forge.actors.policy import EngineConfig, Policy, SamplingConfig
+
+def _import_error():
+    """Check if there are import errors that would cause CI failures."""
+    try:
+        import forge.actors.policy  # noqa: F401
+
+        return False
+    except ImportError:
+        return True
 
 
 class TestPolicyConfig(unittest.TestCase):
     """Test suite for Policy configuration handling after PolicyConfig removal."""
 
+    @pytest.mark.skipif(
+        _import_error(),
+        reason="Import error, likely due to missing dependencies on CI.",
+    )
     def test_policy_default_initialization(self):
         """Policy initializes with default values."""
+        from forge.actors.policy import EngineConfig, Policy, SamplingConfig
+
         policy = Policy()
 
         # Default factories
@@ -35,8 +50,14 @@ class TestPolicyConfig(unittest.TestCase):
         self.assertFalse(policy.sampling_config.guided_decoding)
         self.assertEqual(policy.sampling_config.max_tokens, 512)
 
+    @pytest.mark.skipif(
+        _import_error(),
+        reason="Import error, likely due to missing dependencies on CI.",
+    )
     def test_policy_with_dict_configs(self):
         """Policy accepts dicts for engine_config and sampling_config, including nested dicts."""
+        from forge.actors.policy import EngineConfig, Policy, SamplingConfig
+
         # Test with nested dict structure
         engine_dict = {
             "model": "test-model-6789",
@@ -86,8 +107,14 @@ class TestPolicyConfig(unittest.TestCase):
         )
         self.assertEqual(engine_dict["nested_config"]["custom_settings"]["top_p"], 0.9)
 
+    @pytest.mark.skipif(
+        _import_error(),
+        reason="Import error, likely due to missing dependencies on CI.",
+    )
     def test_policy_yaml_config_loading(self):
         """Policy can be constructed from a YAML config file."""
+        from forge.actors.policy import Policy
+
         yaml_content = """
         engine_config:
           model: "yaml-test-model-9876"
@@ -124,8 +151,14 @@ class TestPolicyConfig(unittest.TestCase):
 
             self.assertEqual(policy.available_devices, "yaml-test-device-xyz")
 
+    @pytest.mark.skipif(
+        _import_error(),
+        reason="Import error, likely due to missing dependencies on CI.",
+    )
     def test_engineconfig_ignores_invalid_keys(self):
         """EngineConfig.from_dict ignores unexpected keys."""
+        from forge.actors.policy import EngineConfig
+
         engine_config = {
             "model": "custom-model",
             "tensor_parallel_size": 2,
--- a/tests/integration_tests/test_policy_update.py
+++ b/tests/integration_tests/test_policy_update.py
@@ -11,7 +11,7 @@ import pytest_asyncio
 
 import torch
 
-from forge.actors.policy import EngineCnofig, Policy, SamplingConfig
+from forge.actors.policy import EngineConfig, Policy, SamplingConfig
 from forge.controller.service import ServiceConfig, spawn_service
 from forge.data.sharding import VLLMSharding
 from torchstore import MultiProcessStore
@@ -168,7 +168,7 @@ def validate_loaded_tensors_equals_original(
 
 def get_configs(worker_size: int, model_name: str) -> Tuple[Dict, ServiceConfig]:
 
-    engine_config = EngineCnofig(
+    engine_config = EngineConfig(
         model=model_name,
         tensor_parallel_size=worker_size,
         pipeline_parallel_size=1,


apps/vllm/config.yaml
```diff
@@ -0,0 +1,19 @@
...
+    tensor_parallel_size: 2
+    pipeline_parallel_size: 1
+    enforce_eager: true
+    vllm_args: null
+  sampling_params:
+    num_samples: 2
+    guided_decoding: false
+  available_devices: null
+
+service_config:
```
pbontrager: This should be:
```
prompt:

policy:
  worker_params
  sampling_params
  available_devices
  service_config
```

apps/vllm/main.py
```diff
@@ -6,81 +6,34 @@
...
-from forge.actors.policy import Policy, PolicyConfig, SamplingOverrides, WorkerConfig
+from forge.actors.policy import Policy, PolicyConfig
 from forge.controller.service import ServiceConfig, shutdown_service, spawn_service
-from vllm.outputs import CompletionOutput
 
 
-async def main():
-    """Main application for running vLLM policy inference."""
-    args = parse_args()
+def load_yaml_config(path: str) -> dict:
```
pbontrager: look at the rl script. you can just use the "@parse" function

apps/vllm/main.py
```diff
@@ -6,81 +6,34 @@
...
+    with open(path, "r") as f:
+        return yaml.safe_load(f)
 
-    # Create configuration objects
-    policy_config, service_config = get_configs(args)
 
-    # Resolve the Prompts
-    if args.prompt is None:
-        prompt = "What is 3+5?" if args.guided_decoding else "Tell me a joke"
+def get_configs(cfg: dict) -> tuple[PolicyConfig, ServiceConfig, str]:
```
pbontrager: This shouldn't exist. I should be able to do:

```
spain_service(
	ServiceConfig(**config.policy.service.pop())
	Policy
	**config.policy
)
```

src/forge/actors/policy.py
```diff
@@ -79,19 +79,35 @@ class WorkerConfig:
...
-    model: str
+    model: str = "meta-llama/Llama-3.1-8B-Instruct"
     tensor_parallel_size: int = 1
     pipeline_parallel_size: int = 1
     enforce_eager: bool = False
-    vllm_args: EngineArgs = None
+    vllm_args: EngineArgs = field(default_factory=EngineArgs)
+
+    @classmethod
+    def from_dict(cls, d: dict):
```
pbontrager: You can use this, but you still need defaults for our class methods because the user might not specify every parameter which should then use the field default.

src/forge/actors/policy.py
```diff
@@ -79,19 +79,35 @@ class WorkerConfig:
...
+        d = dict(d)  # copy
+        if "vllm_args" in d and isinstance(d["vllm_args"], dict):
+            d["vllm_args"] = EngineArgs(**d["vllm_args"])
+        return cls(**d)
 
 
 @dataclass
 class PolicyConfig:
```
pbontrager: PolicyConfig shouldn't exist so this wouldn't be a problem

src/forge/actors/policy.py
```diff
@@ -108,6 +124,8 @@ def __post_init__(self):
         self._policy_proc: ProcMesh | None = None
         self._worker_procs: ProcMesh | None = None
         self.weights_version: int = 0
+        if isinstance(self.config, dict):
```
pbontrager: This check isn't safe because you need to check if each field is a diff. Maybe check if all of the fields are dicts or all are the correct types, and otherwise through an error.

apps/vllm/main.py
```diff
@@ -89,11 +42,11 @@ async def run_vllm(service_config: ServiceConfig, config: PolicyConfig, prompt:
 
     async with policy.session():
         print("Requesting generation...")
-        responses: List[CompletionOutput] = await policy.generate.choose(prompt=prompt)
+        response_output = await policy.generate.choose(prompt=prompt)
```
Jack-Khuu: Thank for the fix; Can we add the typehint?

apps/vllm/main.py
```diff
@@ -104,5 +57,19 @@ async def run_vllm(service_config: ServiceConfig, config: PolicyConfig, prompt:
...
 
 
+def main():
+    import argparse
+
+    parser = argparse.ArgumentParser(description="vLLM Policy Inference Application")
+    parser.add_argument(
+        "--config", type=str, required=True, help="Path to YAML config file"
+    )
+    args = parser.parse_args()
```
Jack-Khuu: Ditto above, won't need if you use `@parse`

apps/vllm/config.yaml
DNXie: Updated.

apps/vllm/main.py
DNXie: Fixed.

apps/vllm/main.py
DNXie: Updated the main.py with `@parse` . Please check.

src/forge/actors/policy.py
DNXie: Removed PolicyConfig

src/forge/actors/policy.py
DNXie: I tested with `vllm_args` is a string (see test case `test_invalid_worker_config_from_dict`), `null` (see the config file), and with parameters (see test case `test_policy_yaml_config_loading`)

apps/vllm/main.py
DNXie: Added.

apps/vllm/main.py
DNXie: Fixed.

src/forge/actors/policy.py
DNXie: It should use default values when everything is empty. See the test case `test_policy_default_initialization`

apps/vllm/main.py
```diff
@@ -6,94 +6,40 @@
...
-
     async with policy.session():
         print("Requesting generation...")
-        responses: List[CompletionOutput] = await policy.generate.choose(prompt=prompt)
+        response_output: RequestOutput = await policy.generate.choose(prompt=prompt)
 
         print("\nGeneration Results:")
         print("=" * 80)
-        for batch, response in enumerate(responses):
+        for batch, response in enumerate(response_output.outputs):
```
Jack-Khuu: fyi I spun this into https://github.com/meta-pytorch/forge/pull/123, to fastforward the fix to main

DNXie:
Merged https://github.com/meta-pytorch/forge/pull/122 to make the unit test pass. Thanks @casteryh

.github/workflows/unit_test.yaml
```diff
@@ -21,6 +21,15 @@ jobs:
           miniconda-version: "latest"
           activate-environment: test
           python-version: ${{ matrix.python-version }}
+      - name: Install torchstore
```
pbontrager: This shouldn't be in this PR

apps/grpo/llama3_8b.yaml
```diff
@@ -0,0 +1,64 @@
+# GRPO training configuration
```
pbontrager: These should be organized by service name:

```
trainer:
	model:
	dataset:
	...
	service: # for service config
	
policy:
	...
	service:

replay_buffer:
	...
	service

...
```

But also I'd leave grpo for a followup PR

apps/vllm/llama3_8b.yaml
```diff
@@ -0,0 +1,19 @@
...
+    tensor_parallel_size: 2
+    pipeline_parallel_size: 1
+    enforce_eager: true
+  sampling_overrides:
+    num_samples: 2
+    guided_decoding: false
+    max_tokens: 512
+  available_devices: null
+
+service_config:
```
pbontrager: Put this at the same level as worker_params and maybe just call it "service"

src/forge/actors/policy.py
```diff
@@ -61,41 +63,36 @@ class SamplingOverrides:
...
 
-    num_samples: int
+    num_samples: int = 1
     guided_decoding: bool = False
     max_tokens: int = 512
 
 
 @dataclass
-class WorkerConfig:
+class WorkerConfig(EngineArgs):
```
pbontrager: Maybe let's just name this EngineConfig?

src/forge/actors/policy.py
```diff
@@ -61,41 +63,36 @@ class SamplingOverrides:
         guided_decoding: Whether to use guided decoding.
     """
 
-    num_samples: int
+    num_samples: int = 1
```
pbontrager: You need to pull changes from main here, maybe this can inherit from GuidedDecoding in vLLM too

src/forge/actors/policy.py
pbontrager: @Jack-Khuu Is there any problem with inheriting from this directly?

src/forge/actors/policy.py
```diff
@@ -108,13 +105,19 @@ def __post_init__(self):
         self._policy_proc: ProcMesh | None = None
         self._worker_procs: ProcMesh | None = None
         self.weights_version: int = 0
+        if isinstance(self.worker_params, Mapping):
+            self.worker_params = WorkerConfig.from_dict(self.worker_params)
+        if isinstance(self.sampling_overrides, dict):
```
pbontrager: This should follow the above patter exactly (Mapping, from_dict)

src/forge/actors/policy.py
```diff
@@ -108,13 +105,19 @@ def __post_init__(self):
...
+            self.sampling_overrides = SamplingOverrides(**self.sampling_overrides)
 
     @classmethod
     async def launch(  # pyright: ignore[reportIncompatibleMethodOverride]
         cls: type["Policy"],
         *,
         process_config: ProcessConfig,
-        config: PolicyConfig,
+        worker_params: WorkerConfig | dict = WorkerConfig(),
```
pbontrager: Use Mapping

src/forge/actors/policy.py
```diff
@@ -128,16 +131,25 @@ async def launch(  # pyright: ignore[reportIncompatibleMethodOverride]
         policy_proc_config.with_gpus = False
 
         policy_proc = await get_proc_mesh(process_config=policy_proc_config)
+
+        if isinstance(worker_params, (dict, DictConfig)):
```
pbontrager: Use Mapping

src/forge/actors/policy.py
```diff
@@ -174,7 +186,7 @@ async def setup(self):
         self.vllm_args = await self.policy_worker.get_vllm_args.choose()
 
         # Setup sampling params
-        sampling_overrides = self.config.sampling_params
+        sampling_overrides = self.sampling_overrides
```
pbontrager: Update this to match current main

.github/workflows/unit_test.yaml
DNXie: Removed

apps/grpo/llama3_8b.yaml
DNXie: Removed for now. Will open a followup PR

src/forge/actors/policy.py
DNXie: Renamed it to `EngineConfig` and all `worker_params` to `engine_params`

src/forge/actors/policy.py
DNXie: Updated

src/forge/actors/policy.py
DNXie: Updated

apps/vllm/llama3_8b.yaml
DNXie: Updated

src/forge/actors/policy.py
DNXie: Rebased and updated.

apps/grpo/main.py
```diff
@@ -362,12 +362,8 @@ async def main():
...
             ServiceConfig(procs_per_replica=1, with_gpus=True, num_replicas=1),
             Policy,
-            config=PolicyConfig(
-                worker_params=WorkerConfig(model=model),
-                sampling_params=SamplingOverrides(
-                    n=group_size, max_tokens=max_res_tokens
-                ),
-            ),
+            engine_params=EngineConfig(model=model),
+            sampling_overrides=SamplingOverrides(n=group_size, max_tokens=16),
```
Jack-Khuu:
```suggestion
            sampling_overrides=SamplingOverrides(n=group_size, max_tokens=max_res_tokens),
```

src/forge/actors/policy.py
DNXie: Updated.

src/forge/actors/policy.py
DNXie: Rebased. Will leave the inheritance to future PR.

src/forge/actors/policy.py
```diff
@@ -72,37 +73,39 @@ def __post_init__(self):
...
-        pipeline_parallel_size: Number of pipeline parallel workers.
-        enforce_eager: Whether to enforce eager mode.
-        vllm_args: vLLM engine args.
+    EngineConfig extends EngineArgs with worker-specific fields.
+    Overlapping keys in input dict will override EngineArgs defaults.
     """
 
-    model: str
+    model: str = "meta-llama/Llama-3.1-8B-Instruct"
     tensor_parallel_size: int = 1
```
DNXie, line 92: I read it [here](https://github.com/meta-pytorch/forge/blob/main/src/forge/actors/policy.py#L299) that `tensor_parallel_size` is under `EngineConfig.parallel_config.tensor_parallel_size`. If so, Is this implementation correct?  Should the user pass the value like this instead:
```
policy:
  engine_params:
     parallel_config:
        tensor_parallel_size = 1
```

@pbontrager

apps/vllm/main.py
```diff
@@ -6,95 +6,34 @@
...
-    args = parse_args()
 
-    # Create configuration objects
-    policy_config, service_config = get_configs(args)
+async def run(cfg: DictConfig):
 
-    # Resolve the Prompts
-    if args.prompt is None:
-        prompt = "What is 3+5?" if args.guided_decoding else "Tell me a joke"
+    if "prompt" in cfg and cfg["prompt"] is not None:
```
Jack-Khuu:
```suggestion
    if (prompt := cfg.get("prompt")) is None:
        gd = cfg.policy.get("sampling_overrides", {}).get("guided_decoding", False)
        prompt = "What is 3+5?" if gd else "Tell me a joke"
```

Jack-Khuu: This can replace the if and else (line selector didn't like my selection)

apps/vllm/main.py
```diff
@@ -6,95 +6,34 @@
...
-    )
+    print("Spawning service...")
 
-    policy_config = PolicyConfig(
-        worker_params=worker_params, sampling_params=sampling_params
-    )
-    service_config = ServiceConfig(
-        procs_per_replica=worker_size, num_replicas=1, with_gpus=True
+    policy = await spawn_service(
+        ServiceConfig(**cfg.policy.service), Policy, **cfg.policy
```
Jack-Khuu, line 32: This seems funky 

Does Policy need the service configs args?

apps/vllm/llama3_8b.yaml
```diff
@@ -0,0 +1,19 @@
...
+    enforce_eager: true
+  sampling_overrides:
+    n: 2
+    guided_decoding: false
+    max_tokens: 512
+  available_devices: null
+  service:
+    procs_per_replica: 2
+    num_replicas: 1
+    with_gpus: true
```
Jack-Khuu, line 15: Out of scope for this PR, but we should think about the "service in yaml pattern" when we have some breathing room

We're gonna have a pattern of excluding this field when passings args around (since `X.service` is not a common Agent Arg)

src/forge/actors/policy.py
Jack-Khuu: Will respond on updated PR

src/forge/actors/policy.py
Jack-Khuu: I comment on this below, what we have is fine since parallel_config doesn't actually exist until `create_engine_config` is called

src/forge/actors/policy.py
```diff
@@ -348,11 +368,7 @@ async def stop(self):
 
 @dataclass
 class PolicyWorker(ForgeActor):
-    model: str
-    tensor_parallel_size: int = 1
-    pipeline_parallel_size: int = 1
-    enforce_eager: bool = False
-    vllm_args: EngineArgs = None
+    vllm_args: EngineConfig | dict = EngineConfig()
```
Jack-Khuu: So there is an typing/design quirk that is throughout the repo, that I'm not a huge fan of, but it's out of scope for this PR

The types used during construction isn't necessarily the same type as the field after construction. In this case vllm_args as input can be a `dict | EngineArgs` but `dict` gets converted to `EngineArgs` (to make calling easier with the yaml?), and in [`post__init`](https://github.com/meta-pytorch/forge/pull/116/files#diff-831fe9863b096d39b6f902985aaee6e61a43223e546268648df082d801f495e1L397) it further gets transformed into a [`vLLMConfig`](https://github.com/vllm-project/vllm/blob/main/vllm/config/__init__.py#L3316)

src/forge/actors/policy.py
```diff
@@ -72,37 +73,39 @@ def __post_init__(self):
...
-    Config args used for setting up the policy worker.
-
-    Args:
-        model: Model name.
-        tensor_parallel_size: Number of tensor parallel workers.
-        pipeline_parallel_size: Number of pipeline parallel workers.
-        enforce_eager: Whether to enforce eager mode.
-        vllm_args: vLLM engine args.
+    EngineConfig extends EngineArgs with worker-specific fields.
+    Overlapping keys in input dict will override EngineArgs defaults.
```
Jack-Khuu, line 88:
```suggestion
    EngineConfig extends EngineArgs surfacing worker-specific fields.
    Args of this class override EngineArgs defaults.
```

src/forge/actors/policy.py
```diff
@@ -72,37 +73,39 @@ def __post_init__(self):
...
+    def from_dict(cls, d: Mapping):
+        d = dict(d)
+        all_fields = set(cls.__dataclass_fields__.keys())
+        valid_args = {k: v for k, v in d.items() if k in all_fields}
+        return cls(**valid_args)
+
 
 @dataclass
-class WorkerConfig:
+class EngineConfig(EngineArgs):
```
Jack-Khuu, line 85: Following up from the other comment: 

Extending should work fine. I like this more since the current construction of PolicyWorker was convoluted, with the awkward duplication

Jack-Khuu: Since this class is effectively EngineArgs overrides I would call it that `EngineArgOverrides`

src/forge/actors/policy.py
```diff
@@ -368,31 +384,14 @@ def __post_init__(self):
...
-                model=self.model,
-                tensor_parallel_size=self.tensor_parallel_size,
-                pipeline_parallel_size=self.pipeline_parallel_size,
-                enforce_eager=self.enforce_eager,
+        if isinstance(self.vllm_args, dict):
+            self.vllm_args = EngineConfig.from_dict(self.vllm_args)
+        elif not isinstance(self.vllm_args, EngineConfig):
+            raise TypeError(
+                f"vllm_args must be a EngineConfig or dict, got {type(self.vllm_args)}"
             )
```
Jack-Khuu: nit: Do we need this check, since we specify the typehint?

No strong pref if you wanna keep

src/forge/actors/policy.py
```diff
@@ -115,13 +118,21 @@ def __post_init__(self):
         self._policy_proc: ProcMesh | None = None
         self._worker_procs: ProcMesh | None = None
         self.weights_version: int = 0
+        if isinstance(self.engine_params, Mapping):
+            self.engine_params = EngineConfig.from_dict(self.engine_params)
+        if isinstance(self.sampling_overrides, Mapping):
+            self.sampling_overrides = SamplingOverrides.from_dict(
+                self.sampling_overrides
+            )
```
Jack-Khuu: Are we allowing Mapping as an input type just to work around the yaml?

src/forge/actors/policy.py
```diff
@@ -72,37 +73,39 @@ def __post_init__(self):
...
+        all_fields = set(cls.__dataclass_fields__.keys())
+        valid_args = {k: v for k, v in d.items() if k in all_fields}
+        return cls(**valid_args)
 
 
 @dataclass
 class Policy(PolicyInterface):
-    config: PolicyConfig
+    engine_params: EngineConfig | Mapping = field(default_factory=EngineConfig)
+    sampling_overrides: SamplingOverrides = field(default_factory=SamplingOverrides)
```
Jack-Khuu:
```suggestion
    sampling_overrides: SamplingOverrides | Mapping = field(default_factory=SamplingOverrides)
```

tests/unit_tests/test_policy_config.py
```diff
@@ -0,0 +1,143 @@
...
+        self.assertEqual(engine_dict["nested_config"]["gpu_memory_utilization"], 0.9)
+        self.assertEqual(engine_dict["nested_config"]["max_model_len"], 4096)
+        self.assertEqual(
+            engine_dict["nested_config"]["custom_settings"]["temperature"], 0.7
+        )
+        self.assertEqual(engine_dict["nested_config"]["custom_settings"]["top_p"], 0.9)
+
+    def test_policy_yaml_config_loading(self):
+        """Policy can be constructed from a YAML config file."""
+        yaml_content = """
```
Jack-Khuu, line 118: nit: Can we make this a physical test artifact file instead of the tempfile if we're testing loading?

maybe `unit_tests/resources/test_policy.yaml`

apps/vllm/llama3_8b.yaml
DNXie: Could you be more clear with the suggestions?

src/forge/actors/policy.py
DNXie: Yes. Any suggestions?

src/forge/actors/policy.py
DNXie: Good suggestion! Just renamed.

src/forge/actors/policy.py
DNXie: I think you are right. This seems redundant and almost will never triggered. Removed

apps/vllm/llama3_8b.yaml
pbontrager: I personally think we should have spawn_service handle this to make it less awkward but we can do that later.

Something like
```
await spawn_service(Policy, **cfg.policy)
```

where `spawn_service(actor: Actor, service_config: ServiceConfig | Mapping, **kwargs)`

apps/vllm/main.py
pbontrager: They all do, every service needs it's own config for the resources it'll get. See previous comment for how this can be made smoother

src/forge/actors/policy.py
```diff
@@ -72,37 +73,43 @@ def __post_init__(self):
...
+    def from_dict(cls, d: Mapping):
+        d = dict(d)
+        all_fields = set(cls.__dataclass_fields__.keys())
+        valid_args = {k: v for k, v in d.items() if k in all_fields}
+        return cls(**valid_args)
+
 
 @dataclass
-class WorkerConfig:
+class EngineArgOverrides(EngineArgs):
```
pbontrager: I don't like the idea of calling these overrides. Technically we don't even need to define these dataclasses ourselves, we're just taking them from vllm and adding a from_dict method to them. For now, to let it land we can just have EngineConfig and SamplingConfig and then we'll probably switch to using the direct vllm ones in the future.

src/forge/actors/policy.py
```diff
@@ -72,37 +73,43 @@ def __post_init__(self):
...
 
-@dataclass
-class PolicyConfig:
-    worker_params: WorkerConfig
-    sampling_params: SamplingOverrides
-    available_devices: str = None
+    @classmethod
+    def from_dict(cls, d: Mapping):
+        d = dict(d)
+        all_fields = set(cls.__dataclass_fields__.keys())
```
pbontrager: Safer to use public apis like `[f.name for f in fields(cls)]`

apps/vllm/main.py
Jack-Khuu: I might be missing something, but where does Policy use `cfg.policy.service`

apps/vllm/llama3_8b.yaml
Jack-Khuu: Agreed, no action required here

Seconding the API too

src/forge/actors/policy.py
Jack-Khuu: No need to change anything here, but worth us thinking about down the line if we should shim this out across the repo (abstration that handles all the class constructions, actors can act on pure python) so that the actor logic is simpler

src/forge/actors/policy.py
Jack-Khuu: > Technically we don't even need to define these dataclasses
> switch to using the direct vllm ones in the future.

Seconding 

No strong pref on naming since it'll be purged later

apps/vllm/main.py
```diff
@@ -22,7 +22,9 @@
 
 async def run(cfg: DictConfig):
 
-    if "prompt" in cfg and cfg["prompt"] is not None:
+    if (prompt := cfg.get("prompt")) is None:
```
Jack-Khuu, line 25: Looks like there's duplicate logic here

src/forge/actors/policy.py
DNXie: Sounds reasonable. I agree. Let's not include it in this PR for now.

src/forge/actors/policy.py
DNXie: Thanks for the discussion. Just renamed them to `EngineConfig` and `SamplingConfig`

src/forge/actors/policy.py
DNXie: Updated.

apps/grpo/llama3_8b.yaml
DNXie: Submitted the followup PR here https://github.com/meta-pytorch/forge/pull/141

PR #125, opened 2025-09-04
Update reference_actor.py

casteryh:
fix a bug in calculating ref log probs

### DIFF
--- a/src/forge/actors/reference_actor.py
+++ b/src/forge/actors/reference_actor.py
@@ -17,8 +17,6 @@ from dataclasses import dataclass, field, fields
 from typing import Any
 
 import torch
-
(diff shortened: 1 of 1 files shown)


Jack-Khuu:
Can you rebase jic and merge if no issues?

casteryh:
unit test in CI failing b/c torchtitan is not installed, will fix later.

allenwang28:
> unit test in CI failing b/c torchtitan is not installed, will fix later.

For the future, we should not be pushing changes that break CI.

casteryh:
> > unit test in CI failing b/c torchtitan is not installed, will fix later.
> 
> For the future, we should not be pushing changes that break CI.

ack. fixed in #132

Issue #126, opened 2025-09-04
Installation issue (Sudo required - only Fedora distribution)


AlirezaShamsoshoara:
The current installation script has dependencies to SUDO access. Not all users have sudo access on their compute machines. 
Another issue is dependencies to the Fedora distribution:

```bash
sudo dnf install -y libibverbs rdma-core libmlx5 libibverbs-devel rdma-core-devel
sudo dnf install gh"
```

It would be nice to make it compatible with both Conda and other OS distributions.

AlirezaShamsoshoara:
Closing this one as this [PR](https://github.com/meta-pytorch/forge/pull/127) addressed this.

PR #127, opened 2025-09-04
update installation script for conda/ubuntu users without sudo access

AlirezaShamsoshoara:
The changes mainly addresses the installation [issue](https://github.com/meta-pytorch/forge/issues/126) for users without SUDO access (Conda users) or Ubuntu users.

Changes:

- add log warning
- Don't exit if user does not have sudo
- Consider Ubuntu packages in case user wants to install them with sudo
- Consider conda-forge as a channel for conda installation
- gh will be installed at the conda level for users without sudo access

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -36,15 +36,6 @@ check_conda_env() {
     log_info "Installing in conda environment: $CONDA_DEFAULT_ENV"
 }
 
-# Check sudo access only
-check_sudo_only() {
-    if ! sudo -n true 2>/dev/null; then
-        log_error "This script requires passwordless sudo access for system packages"
-        log_info "Run 'sudo -v' first, or configure passwordless sudo"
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -34,22 +36,67 @@ check_conda_env() {
     log_info "Installing in conda environment: $CONDA_DEFAULT_ENV"
 }
 
-# Check sudo access
-check_sudo() {
+# Check sudo access only
+check_sudo_only() {
```
allenwang28: could you remove this and install_systems_packages_sudo_only?

AlirezaShamsoshoara: Thanks @allenwang28 for quickly review that. Just removed those two _only functions from the script.

allenwang28:
looks great, thanks!

PR #130, opened 2025-09-05
[parked] Use gpu runner in CI

casteryh:
Need to activate conda env manually every time because setup-miniconda doesn't work well with self-hosted runner, which is a known issue (https://github.com/conda-incubator/setup-miniconda/issues/269#issue-1508678633).

### DIFF
--- a/.github/workflows/unit_test.yaml
+++ b/.github/workflows/unit_test.yaml
@@ -47,6 +47,8 @@ jobs:
           export LD_LIBRARY_PATH=$CUDA_HOME/lib64:$LD_LIBRARY_PATH
           export LIBRARY_PATH=$CUDA_HOME/lib64:$LIBRARY_PATH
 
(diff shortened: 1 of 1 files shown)


allenwang28:
hey @casteryh  I am excited about this PR! Can you please let me know what support you need?

casteryh:
> hey @casteryh I am excited about this PR! Can you please let me know what support you need?

yeah that would be great. specifically do you know what runner we can use? the one I am trying with seems to be ancient, and I don't know where to find an up-to-date list of available runners.

allenwang28:
https://github.com/pytorch/torchtitan/blob/main/.github/workflows/integration_test_8gpu_h100.yaml I think would be a reasonable reference point, please lmk if you run into issues!

casteryh:
The current issue is the monarch wheel requires glibc 2.29.
But even running yum update on the base docker image pytorch/almalinux-builder won't give you glibc>=2.29.

Jack-Khuu:
cc: @ebsmothers in case it's relevant/useful

PR #132, opened 2025-09-05
Skip tests when there are import issues on CI worker

casteryh:


allenwang28:
Maybe we can use `pytest.importorskip` instead? from https://stackoverflow.com/a/42520506

casteryh:
> Maybe we can use `pytest.importorskip` instead? from https://stackoverflow.com/a/42520506

I think we should explicitly mark these tests as not tested by CI and try to fix CI and migrate these no-CI tests after, instead of burying it.

Adding an automatic pytest skip further complicates it as the test may or may not have been run, and the only way to know is to dig in the logs. For example, if we do introduce a genuine import error in the future, then whole test will be skipped (even locally) , which we probably don't want.

allenwang28:
sure we can explicitly mark these tests as skipped. IMO the right way to do this is with pytest with like `pytest.mark.skip` or something rather than having a separate directory for more tests

casteryh:
> sure we can explicitly mark these tests as skipped. IMO the right way to do this is with pytest with like `pytest.mark.skip` or something rather than having a separate directory for more tests

If you say so. Done.

PR #136, opened 2025-09-08
Make sudo installation optional and add new cases for different OS dist.

AlirezaShamsoshoara:
This PR addresses makes the installation optional for sudo user:
- The default for the installation would be on CONDA channel unless user specifies with an input argument (`--use-sudo`).
- Also, the installation now has different cases for different OS distributions that can be added to the script.
- The README file is updated accordingly.
- The script is tested on Nvidia DevGPU & CoreWeave node for three and eight GPUs.


allenwang28:
Note - error is unrelated to this PR / due to TorchStore which @LucasLLC will look into later

PR #137, opened 2025-09-08
[Icebox] Disperse GRPO app function/classes into respective files

Jack-Khuu:
**There are no functional change**

This PR does some sanitization and moves some the function/classes in `app/grpo` into more appropriate files. Specifically:
- `reference_agent.py` => `reference_model.py`
- `reference_agent.TitanRefModel` => `reference_model.ReferenceModel`
- `grpo.RefModel` => `reference_model.HFReferenceModel`
- `compute_logits` + `Episode` => `trainer.py`

Spinning this up as a separate PR to make reviewing functional changes to RefModel clearer down the line
- Also deletes unused code in reference_model.py
---

```
 python apps/grpo/main.py
```


apps/grpo/main.py
```diff
@@ -28,21 +29,6 @@
 logger.setLevel(logging.DEBUG)
 
 
-def compute_logprobs(
```
Jack-Khuu, line 31: Note that the implementation here is functionally equivalent to the implementation previously in `reference_actor.py` (which was moved to `trainer.py`)

src/forge/actors/trainer.py
```diff
@@ -30,12 +34,85 @@
...
+    token_logprobs = torch.log_softmax(logits / temperature, dim=-1)
+    # (bsz, completion_len, 1)
+    logprobs = torch.gather(token_logprobs, 2, input_ids.unsqueeze(-1))
+    # (bsz, completion_len)
+    logprobs = logprobs.squeeze(-1)
+
+    return logprobs
+
+
+@dataclass
```
allenwang28, line 81: IMO keep this top level in grpo/main, at the very least don't put it in the trainer file

pbontrager:
I don't actually think we should do this yet. I think we should just delete these as we have importable versions. I don't want the HF version to disperse into the codebase and I don't want us to abstract over different backends. We should be extremely lazy about moving things out of main.

src/forge/actors/trainer.py
Jack-Khuu: Thoughts on where to store Episode down the line? 
I'm fine using "Episode" before then

Jack-Khuu:
Quick responses!

Noted, iceboxing this

PR #138, opened 2025-09-08
Publishing weights in to torchstore from RLTrainer and getting them from policy engine.

pradeepfn:
Working integration test that works with Llama8B. 
1. TorchTitan publishes weights to torchstore (TS), 
2. vLLM service consumes weights from TS. 

State-dict conversion to HF format happens during weight-save/publish in to TS using sd_adaptor (mostly).  Custome param concatanations that are not supported by the current sd_adaptor impl of the model was implemented directly within the weights publish function.

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -16,9 +16,6 @@ from typing import Callable
 import torch
 import torchstore as ts
 
-from forge.controller import ForgeActor
-from forge.data.utils import batch_to_device
-
 from monarch.actor import current_rank, current_size, endpoint
 from torch import Tensor
 from torch.distributed.checkpoint._nested_dict import flatten_state_dict
@@ -38,6 +35,9 @@ from torchtitan.config.job_config import (
 from torchtitan.distributed import utils as dist_utils
 from torchtitan.experiments.forge.engine import ForgeEngine
 from torchtitan.experiments.forge.job_config import ForgeJobConfig
+
+from forge.controller import ForgeActor
+from forge.data.utils import batch_to_device
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)


LucasLLC:
@pradeepfn please clarify, does put hang or throw?

tests/integration_tests/test_policy_update.py
```diff
@@ -268,6 +279,8 @@ async def test_llama3_policy_update_single(llama3_torchstore_setup):
     print("Starting Llama 3 8B torchstore test (single GPU)...")
 
     store, original_state_dict = llama3_torchstore_setup
+    # store = MultiProcessStore.create_store()
+    await run_rl_trainer(store, worker_size=1)
```
LucasLLC: I would recommend pulling and using singleton API's. Let me know if that works

pradeepfn: Of course. Done!.

pradeepfn:
> @pradeepfn please clarify, does put hang or throw?

No torchstore issues.

src/forge/actors/trainer.py
```diff
@@ -91,7 +93,9 @@ def __post_init__(self):
     @endpoint
     async def setup(self):
         # TODO: update ForgeEngine to not use ForgeJobConfig
-        engine_config = {f.name: getattr(self, f.name) for f in fields(self)}
+        engine_config = {
+            f.name: getattr(self, f.name) for f in fields(self) if f.name != "store"
```
pbontrager: store isn't a field, why is this needed?

src/forge/actors/trainer.py
```diff
@@ -68,7 +70,7 @@ def __post_init__(self):
                     f"{f.name} should be a {f.type} type or a dict like object"
                 )
 
-        self.current_step = 0
+        self.current_step = 1  # fragile contract.
```
pbontrager, line 80: Why are we starting at 1? Also, we probably want a todo to update this from the checkpoint

src/forge/actors/trainer.py
```diff
@@ -261,8 +265,49 @@ def train_step(self, batch) -> None:
...
+        # 1. Use the torch.titan adaptor's 'to_hf' routines to convert the state dict.
+        # 2. Missing conversions ( QKV, MLP fusion) is done using custom code. Probably
+        #    we should move that code to 'to_hf' function.
+
+        assert (
+            self.engine.checkpointer.sd_adapter is not None
+        ), "trying to save checkpoint in HF safetensors format, but sd_adapter is not provided."
+        hf_state_dict = self.engine.checkpointer.sd_adapter.to_hf(flattened_state_dict)
+
+        for i in range(32):  # improve this using regex similar to to_hf function.
```
pbontrager: Is this the hf_to_vllm logic? Can we pull that out as a function in the bottom of the file (like llama3_to_vllm) or something so it's easy to see which part of the code we want to get rid of in the future, and so we can support Qwen in the meantime as well?

src/forge/actors/trainer.py
```diff
@@ -261,8 +265,49 @@ def train_step(self, batch) -> None:
...
+        # save to torchstore. Hacking in to the Checkpointer's prepped state-dict for now.
+        # TODOs:
+        # 1. Checkpoint invokes state-dict flattening during dcp_save for [MODEL].
+        #    May need to replicate the same in this code path.
+        # 2. Unify CheckpointManager and TorchStore weights save control path.
+        print(f"Getting keys from checkpointer state and pushing to TS ...")
+        assert (
+            "model" in self.engine.checkpointer.states
+        ), "Model state not found in checkpointer state"
+        sd = self.engine.checkpointer.states["model"].state_dict()
```
pbontrager, line 215: Where is this coming from? When you call this, does it create the sd right then or did it have to be saved in the train step earlier? Does it return the sd on GPU or CPU? Also does it handle blocking the trainer from updating the weights while it's getting them?

tests/integration_tests/test_policy_update.py
```diff
@@ -248,55 +210,74 @@ async def llama3_torchstore_setup():
...
     )
+    results = await policy._get_model_params.call()
+    assert len(results) == 1
+    print("Successfully got model state dict after update")
+    return results[0]
+
+
+@pytest.mark.asyncio
+@requires_cuda
+async def test_llama3_policy_update_single(setup_test):
```
pbontrager: Can we have a 2 gpu test two? Maybe the trainer with dp=2 and vllm with tp=2?

Also, I'm not sure if these tests would catch if the true weights are coming through as we don't change the state dict.

src/forge/actors/trainer.py
pradeepfn: Good catch. This is missing refactor. TS client went to singleton mode, and we removed store as a init param. Missed this reference.

src/forge/actors/trainer.py
pradeepfn: Because policy engine starting at 1. Lets keep this fragile contract as it is. The true version has to come from a config or external book-keeping entity.

src/forge/actors/trainer.py
pradeepfn: I'm accessing the module state-dict prepped by torch.titan as part of checkpoint save. 

1. This is a in-memory state-dict. ( Tensor/DTensor).
2. It returns tensors with original storage. Means GPU/UVM backed tensors.

>> Also does it handle blocking the trainer from updating the weights while it's getting them?

Hmm.. it does not block the trainer. However, ForgeEngine drive the trainer using train_step. Therefore, there is no race-conditions with current code.

There is improvements to be made to this code. In the ideal case;
1. the state-dict get prepped for weight-exchange and checkpoint save purposes. 
2. Once the initial state-dict prep we can cache the prepped state-dict for later iterations of the training steps for efficiency reasons ( if there is opportunity). 
3. We move all the model weights and optimizer state to torchstore.
4. Policy engine  (only) lookup the model-weights from torchstore
5. Async checkpointing upload lookups model-weights and optimizer states for uploading in to remote persistent storage.


We don't have all the piece right now. But tapping in to checkpoint state-dict is the right thing to do as the first step.

tests/integration_tests/test_policy_update.py
pradeepfn: Of course. However, I want to land the simplest thing first in this PR. Let me create a issue for that.
https://github.com/meta-pytorch/forge/issues/143

src/forge/actors/trainer.py
pradeepfn: Great question. I picked up this set of changes (torch.cat) from Ankita's code. I will verify if this is llama3_to_hf or hf_to_vllm.
if it is former, I will but this in 'to_hf' functions in llama3 model.
if it is latter, I can go your suggested path.

pradeepfn: @ankitageorge @tianyu-l  can you help me figure out, where does QKV fusion and MLP fusion fits in ?

pradeepfn: Confirmed with Ankita, this is hf->vllm code.

tianyu-l: @pradeepfn 
I saw the function `llama3_hf_to_vllm` which looks awkward. I think it's better to have `to_vllm` in torchtitan sd_adapter, instead of calling `to_hf`.

src/forge/actors/trainer.py
```diff
@@ -261,10 +264,60 @@ def train_step(self, batch) -> None:
...
     @endpoint
-    def push_weights(self) -> None:
-        pass
+    async def push_weights(self) -> None:
+        # save to torchstore. Hacking in to the Checkpointer's prepped state-dict for now.
+        # TODOs:
+        # 1. Checkpoint invokes state-dict flattening during dcp_save for [MODEL].
+        #    May need to replicate the same in this code path.
+        # 2. Unify CheckpointManager and TorchStore weights save control path.
+        print(f"Getting keys from checkpointer state and pushing to TS ...")
```
allenwang28: nit
```suggestion
        logger.info(f"Getting keys from checkpointer state and pushing to TS ...")
```

src/forge/actors/trainer.py
```diff
@@ -261,10 +264,60 @@ def train_step(self, batch) -> None:
...
+        assert (
+            "model" in self.engine.checkpointer.states
+        ), "Model state not found in checkpointer state"
```
allenwang28: When would `"model"` not be in the `self.engine.checkpointer.states`? In other words, can we update the assertion error to be more informative? Does this fail if the user didn't initialize the trainer properly/what do they need to do to make it work??

tests/integration_tests/test_policy_update.py
```diff
@@ -145,22 +140,25 @@ def validate_loaded_tensors_equals_original(
...
                 tensor_to_compare = expected_tensor.cpu().float()
 
+            # Training trainer emitted and loaded tensors are of type bfloat16,
+            # where as a HF model loaded(expected) tensor has type float16.
             if not torch.allclose(
                 loaded_tensor.float(),
                 tensor_to_compare,
-                rtol=1e-5,
-                atol=1e-8,
+                rtol=1e-2,
```
allenwang28, line 148: is this new `rtol`/`atol` expected? question for @pbontrager

tests/integration_tests/test_policy_update.py
```diff
@@ -244,55 +208,74 @@ async def llama3_torchstore_setup():
...
+    """
+    Spawn the RL trainer
+    Args:
+        worker_size: Number of workers/procs.
+    """
+    cfg: DictConfig = OmegaConf.load("apps/rl/llama3_8b.yaml")
+    rl_trainer = await spawn_service(
+        ServiceConfig(procs_per_replica=worker_size, with_gpus=True, num_replicas=1),
+        RLTrainer,
+        **cfg.trainer,
```
allenwang28: maybe we should hardcode the trainer config here rather than load from `apps/rl/llama3_8b.yaml`

src/forge/actors/trainer.py
pbontrager: I don't think we can change this without risking breaking checkpoint expectations from titan side. I'd rather just use a separate variable in the trainer for "checkpoint name" (can be a property that's just current_step + 1 for now). This could also be passed in from the controller which would be better.

src/forge/actors/trainer.py
pbontrager: I guess you're right that it should be mostly safe since we control the update from the controller. But since they're async calls they could be overlapped so we'll have to be careful for now.

tests/integration_tests/test_policy_update.py
pbontrager: This is never great, but given the bf16/fp16 comments I could see that. This is also an allclose and not a comparison of the mean so we should be safe here. If we can load the hf side with bf16 instead of fp16 we might be able to regain the tighter tolerance.

pradeepfn:
FYI: @joecummings  is working on merging this diff. This is because of recent API changes that was pushed by him. @joecummings  can add more insights. thanks.

src/forge/actors/trainer.py
pradeepfn: Essentially, this only happens if the checkpoint_manager of torchtitan is not initialized prior to calling push_weights routine. I can update the error message (followup PR) with that.

src/forge/actors/trainer.py
```diff
@@ -197,8 +204,30 @@ def train_step(
...
+        if self.engine.checkpointer.sd_adapter is None:
+            raise RuntimeError(
+                "Trying to save checkpoint in HF safetensors format, but sd_adapter is not provided."
+            )
+        hf_state_dict = self.engine.checkpointer.sd_adapter.to_hf(flattened_state_dict)
+        # TODO: Figure out how to gracefully handle which model to-vLLM conversion is needed
+        vllm_ready_hf_sd = _qwen3_hf_to_vllm(sd=hf_state_dict, num_layers=28)
+        key = f"{self.state_dict_key}{DELIM}{policy_version}"
+        start_time = time.time()
+        await ts.put_state_dict(state_dict=vllm_ready_hf_sd, key=key)
```
Ritesh1905, line 226: Question: are there benfits to doing this at the state dict level and not at the key level where we could parallelize the individual put operation per key?

joecummings: Right now, the benefit is simplicity. 

Eventually I imagine we will want to do this on a per-slice level.

cc @LucasLLC

src/forge/actors/trainer.py
```diff
@@ -197,8 +204,30 @@ def train_step(
...
+            raise RuntimeError("Model state not found in checkpointer state")
+        sd = self.engine.checkpointer.states["model"].state_dict()
+        flattened_state_dict, _ = flatten_state_dict(sd)
+        if self.engine.checkpointer.sd_adapter is None:
+            raise RuntimeError(
+                "Trying to save checkpoint in HF safetensors format, but sd_adapter is not provided."
+            )
+        hf_state_dict = self.engine.checkpointer.sd_adapter.to_hf(flattened_state_dict)
+        # TODO: Figure out how to gracefully handle which model to-vLLM conversion is needed
+        vllm_ready_hf_sd = _qwen3_hf_to_vllm(sd=hf_state_dict, num_layers=28)
```
Ritesh1905, line 223: Curious: what is the reasoning for doing this at the learner and not at the generator? The trainer just pushes it weights and the generator can based on it's implementation (vLLM, sglang etc.) can modify the sd.

pradeepfn: This is an temp thing I did. It will be moved to generator sd loading + it will be moved out of the trainer/generator critical path based on efficiency numbers.

pbontrager: To add to Pradeep's answer, vLLM already handles it's own hf -> vllm mapping. The only reason we've recreated it is so we can add a shaded loading solution which we want to eventually upstream. It will be on the generator side like he said.

PR #139, opened 2025-09-08
Update RefModel to use Episode

Jack-Khuu:
Mini update to (titan) `ReferenceModel` signature to match RefModel of `apps/GRPO/main` by using `Episode` as input

Git refuses to pick up the file rename (`reference_actor.py` => `reference_model.py`), so I'll callout explicitly what was changed in `reference_model.py`
- Deleted code under the experimental section
- Rename `TitanRefModel` => `ReferenceModel`
- Updated forward to use Episode (which wraps req/resp)
 
```python
async def forward(self, episode: 'Episode') -> torch.Tensor:
async def forward(self, request: list[int], response: list[int]) -> torch.Tensor:
```

---
`python apps/grpo/main.py`


Jack-Khuu:
>  import the logprobs computation from the one in trainer when it lands.

trainer.py doesn't have a logprobs calc, though the one in `rl/main.py` Is functionally equivalent to the one here. I'll have a separate PR combining the 3 (rl/grpo/ref)

PR #140, opened 2025-09-08
Off-by-1 GRPO

joecummings:
https://wandb.ai/jcummings/grpo-training/workspace?nw=nwuserjcummings

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -196,9 +196,9 @@ class Trainer(ForgeActor):
 
         mask = response != pad_id
         loss = self.loss(logprobs, ref_logprobs, advantages, mask)
-        self.optimizer.zero_grad()
         loss.backward()
         self.optimizer.step()
+        self.optimizer.zero_grad(set_to_none=True)
 
         return loss.item()
 
@@ -447,7 +447,7 @@ async def main(cfg: DictConfig):
             if batch is None:
                 await asyncio.sleep(0.1)
             else:
-                loss = sum(await trainer.train_step.call(batch))
+                loss = await trainer.train_step.choose(batch)
                 training_step += 1
                 mlogger.log("loss/training_step", loss, training_step)
                 await trainer.push_weights.call(policy_version)
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -243,9 +243,6 @@ class Policy(PolicyInterface):
         Returns:
             RequestOutput: vLLM class with the generated response.
         """
-        return await self._generate(prompt, priority)
-
-    async def _generate(self, prompt: str, priority: int = 0) -> RequestOutput:
         self.request_id += 1 % sys.maxsize
         request_id = str(self.request_id)  # implement from a counter
 


casteryh:
Just fyi the link gives me 404

joecummings:
> Just fyi the link gives me 404

https://wandb.ai/jcummings/grpo-training/workspace?nw=nwuserjcummings

Try now?

apps/grpo/main.py
```diff
@@ -208,11 +209,19 @@ async def train_step(self, batch: list[Episode]):
...
 
     @endpoint
-    async def push_weights(self):
-        pass
+    async def push_weights(self, version: int):
+        """Update policy model weights with trainer's current weights."""
+        start_time = time.time()
+        assert self.store is not None, "Store must be provided to save weights"
+        key = f"{self.state_dict_key}{DELIM}{version}"  # Use version as unique id
+        await push_state_dict(self.store, self.model.state_dict(), key)
```
pbontrager: You should call convert_state_dict here so you can use the policy weight update like here https://github.com/meta-pytorch/forge/blob/a3d2466637a8348c3bbbcb022f3a3be993035ff7/tests/integration_tests/test_policy_update.py#L253

apps/grpo/main.py
```diff
@@ -239,15 +251,8 @@ async def compute(self, group: Group) -> list[float]:
...
-
-        # if std is nan, return 0s. Remove this before shipping
-        if std.isnan().any():
-            advantages = torch.zeros_like(rewards)
-        else:
-            advantages = (rewards - mean) / (std + 1e-4)
-
-        x = advantages.squeeze(0).tolist()
-        return x
+        advantages = (rewards - mean) / (std + 1e-4)
```
pbontrager, line 242: Should we throw an error if g == 1?

apps/grpo/main.py
```diff
@@ -409,7 +418,13 @@ async def continuous_rollouts():
                 print("Dataloader is empty, exiting continuous rollout")
                 return
             prompt, target = sample["request"], sample["target"]
-            version = 0  # await policy.get_current_version.choose()
+            responses = await policy.generate.choose(prompt)
```
pbontrager, line 402: We'll throw away a lot of data this way for fully on policy

apps/grpo/main.py
```diff
@@ -436,30 +449,33 @@ async def continuous_rollouts():
...
-                    print(f"Completed {training_step} training steps")
-                    if training_result:
-                        loss_value = training_result.get("loss", 0.0)
-                        print(f"Latest loss: {loss_value}")
-                        logger.log("loss/training_step", loss_value, training_step)
-                # await trainer.update_weights(policy)
+                logger.log("loss/training_step", loss, training_step)
+                await trainer.push_weights.choose(policy_version)
+                policy_version += 1
+                await policy.update_weights.choose()
```
pbontrager: Shouldn't this be a call? We want all replicas updated

apps/grpo/main.py
```diff
@@ -436,30 +449,33 @@ async def continuous_rollouts():
...
                 training_step += 1
-                if training_step % 10 == 0:
-                    print(f"Completed {training_step} training steps")
-                    if training_result:
-                        loss_value = training_result.get("loss", 0.0)
-                        print(f"Latest loss: {loss_value}")
-                        logger.log("loss/training_step", loss_value, training_step)
-                # await trainer.update_weights(policy)
+                logger.log("loss/training_step", loss, training_step)
+                await trainer.push_weights.choose(policy_version)
```
pbontrager: This should also be call technically even though choose works since replicas=1

apps/grpo/main.py
```diff
@@ -436,30 +449,33 @@ async def continuous_rollouts():
...
         while True:
-            batch = await replay_buffer.sample.choose(curr_policy_version=0)
+            batch = await replay_buffer.sample.choose(
+                curr_policy_version=policy_version
+            )
             if batch is None:
                 await asyncio.sleep(0.1)
             else:
-                training_result = await trainer.train_step.choose(batch)
+                loss = await trainer.train_step.choose(batch)
```
pbontrager, line 450: This should also be a call

src/forge/actors/policy.py
```diff
@@ -313,25 +329,89 @@ async def run(self):
...
-        """Update the policy weights."""
-        # Wait for all current requests to finish, then publish model weights
-        futures = [fut for _, fut in self.requests.values()]
-        if futures:
-            await asyncio.gather(*futures)
-        new_version = self.weights_version + 1
-        await self.policy_worker.update.call(version=new_version)
-        self.weights_version = new_version
-        return self.weights_version
+    async def update_weights(self):
```
pbontrager, line 344: I think we want to keep this as simple as possible in regards to queueing logic. I think we can just add a wait loop to the original function that waits for the scheduler to be empty. In the generate function we can have a similar loop that waits if the request version is newer than the current version.

```
futures = [fut for _, fut in self.requests.values()]
if futures:
	await asyncio.gather(*futures)

while sum(self.scheduler.get_request_counts()) > 0:
	asyncio.sleep(.1)

new_version = self.weights_version + 1
await self.policy_worker.update.call(version=new_version)
self.weights_version = new_version
return new_version
```

Scheduler api [here](https://github.com/vllm-project/vllm/blob/1c63a16b653d1f3f4260e862f01eaffed283c4c8/vllm/v1/core/sched/scheduler.py#L1072)

casteryh:
> > Just fyi the link gives me 404
> 
> https://wandb.ai/jcummings/grpo-training/workspace?nw=nwuserjcummings
> 
> Try now?

Saw the graphs 📈 nice!! it's working!

src/forge/actors/policy.py
```diff
@@ -313,25 +329,89 @@ async def run(self):
...
+    async def _process_queued_requests(self):
+        """Process all queued requests after weight update completes."""
+        queued_requests = self._request_queue.copy()
+        self._request_queue.clear()
+
+        for i, (prompt, priority, future) in enumerate(queued_requests):
+            try:
+                # Use the internal method directly to avoid the updating weights check
+                result = await self._generate(prompt, priority)
+                future.set_result(result)
```
Jack-Khuu: Is skipping the update check safe?

apps/grpo/main.py
joecummings: For short responses yeah definitely, if you look at the WandB logs (buffer_size/rollout), you can see that we build up a buffer of about 100 episodes and then evict the majority of them back and forth during weight updates. 

When we start allowing much longer generations and our models are much bigger, this won't be as big of an issue.

apps/grpo/main.py
```diff
@@ -340,6 +345,8 @@ async def main():
         project="grpo-training",
     )
 
+    store = await MultiProcessStore.create_store()
```
joecummings: @LucasLLC  Is this still the recommended way of doing things?

apps/grpo/main.py
joecummings: It really should never be 1 b/c then there's no advantages to be calculation and the whole algo falls apart. I'd lean towards assuming our users are slightly more intelligent than that.

apps/grpo/main.py
joecummings: In the end tho this conversion logic will be owned by Titan, yes?

apps/grpo/main.py
```diff
@@ -436,30 +449,33 @@ async def continuous_rollouts():
                 episode.advantage = advantage
                 await replay_buffer.add.choose(episode)
 
+            avg_response_len = (
+                sum(len(e.response_tokens) for e in group.episodes) / group_size
+            )
+            logger.log("avg_response_len/rollout", avg_response_len, rollout_count)
```
Jack-Khuu: Unrelated to this PR, but might be worth creating a no-log mode if logging gets taxing down the lien

src/forge/actors/policy.py
```diff
@@ -221,12 +220,21 @@ def start_processing(self):
             self._run_task = asyncio.create_task(self.run())
 
     @endpoint
-    async def generate(self, prompt: str, priority: int = 0) -> List[CompletionOutput]:
+    async def generate(self, prompt: str, priority: int = 0) -> RequestOutput | None:
```
Jack-Khuu: Thanks

src/forge/actors/policy.py
```diff
@@ -115,6 +111,9 @@ def __post_init__(self):
         self._policy_proc: ProcMesh | None = None
         self._worker_procs: ProcMesh | None = None
         self.weights_version: int = 0
+        self._updating_weights: bool = False
+        self._request_queue: list[tuple[str, int, asyncio.Future]] = []
```
Jack-Khuu: Would love a comment describing the `_request_queue` since we also have `requests` as a class member

src/forge/actors/policy.py
```diff
@@ -313,25 +329,89 @@ async def run(self):
                 engine_core_timestamp=outputs.timestamp,
                 iteration_stats=None,
             )
+
             for request_output in processed_outputs.request_outputs:
                 if request_output.finished:
-                    _, fut = self.requests.pop(request_output.request_id)
-                    fut.set_result(request_output)
+                    if request_output.request_id in self.requests:
```
Jack-Khuu: Can combine conditionals?

Just a sanity check: What is the state where `if request_id  not in requests`?

src/forge/actors/policy.py
```diff
@@ -440,15 +520,18 @@ async def update(self, version: int):
         if self.torchstore is None:
             raise Exception("No torchstore configured, skipping model update")
 
-        logger.debug(
-            f"Starting model update from torchstore with key: {self.state_dict_key}{DELIM}{version}"
-        )
-
+        key = f"{self.state_dict_key}{DELIM}{version}"
```
Jack-Khuu, line 435: Let's make this a function since the caller also uses it

apps/grpo/main.py
casteryh: > This should also be call technically even though choose works since replicas=1

As a side note, even if we do `call()`, what we are doing here is all the trainers training on the same batch right?

The replicas are just for fault tolerance? In this regard, if we want different trainers to train on different batches, the trainers themselves have to pull the batches right?

An alternative way to do this is we split the `batch` into `microbatch`es and then call `choose()` on each `microbatch`. After the whole `batch` is done. We then do an `all_reduce` (or other forms of reduce) to average the weights.

src/forge/actors/policy.py
casteryh: Definitely make this a function, and we should do `f"{version}{DELIM}{self.state_dict_key}"` instead.
Which is the correct hierarchy we should use given the new `keys(prefix)` in torchstore api.

src/forge/actors/policy.py
```diff
@@ -313,25 +329,89 @@ async def run(self):
...
+            self.logger.info(f"Weight update completed (now v{self.weights_version})")
+        except Exception as e:
+            self.logger.error(f"Weight update failed: {e}")
+            self._updating_weights = False
+            raise
+
+        self._updating_weights = False
+
+        # Schedule queue processing as a separate task to avoid blocking the endpoint
+        if self._request_queue:
```
casteryh: I am not sure about this point. But there are several gotchas when you use `create_task`.
To err on the safe side, I would find a way to store the reference to the task so that it doesn't get GC'd mid-execution.

apps/grpo/main.py
casteryh: Seems like we are using `ts.initialize()` now and there is a global singleton torchstore. But I will let @LucasLLC  weigh in.

apps/grpo/main.py
```diff
@@ -148,78 +143,115 @@ class Trainer(ForgeActor):
     model_name: str
     learning_rate: float = 1e-5
     beta: float = 0.1
-    epsilon: float = 0.1
     device: torch.device | None = None
+    store: MultiProcessStore | None = None
```
pbontrager: This isn't the current recommended way to use store now. You should just call it as a singleton inside of the trainer

apps/grpo/main.py
```diff
@@ -148,78 +143,115 @@ class Trainer(ForgeActor):
...
 
-        # Initialize loss
-        self.loss = SimpleGRPOLoss(self.epsilon, self.beta)
-
-        self.logger.info(f"Model initialized on {self.device}")
+        self.loss = SimpleGRPOLoss(self.beta)
+
+        self.logger.info(f"Trainer model initialized on {self.device}")
+
+    def _qwen3_hf_to_vllm(self, saved_sd):
```
pbontrager: Can you put this in the trainer.py file? We'll need to reuse this with titan and this will merge nicely with Pradeep's PR

apps/grpo/main.py
```diff
@@ -148,78 +143,115 @@ class Trainer(ForgeActor):
...
         self.optimizer.zero_grad()
         loss.backward()
-
-        # # Gradient clipping (optional but recommended for stability)
-        # torch.nn.utils.clip_grad_norm_(self.model.parameters(), max_norm=1.0)
-
         self.optimizer.step()
 
-        return {"loss": loss.item()}
+        return loss.detach()
```
pbontrager: I think item also detaches and moves to cpu

apps/grpo/main.py
```diff
@@ -297,16 +322,24 @@ class DatasetActor(ForgeActor):
     revision: str = "main"
     data_split: str = "train"
     streaming: bool = True
-    model: str = "Qwen/Qwen3-1.7B-Base"
+    tokenizer: str = "Qwen/Qwen3-1.7B"
```
pbontrager: nit: I don't like this name change

apps/grpo/main.py
```diff
@@ -434,30 +467,33 @@ async def continuous_rollouts():
...
         while True:
-            batch = await replay_buffer.sample.choose(curr_policy_version=0)
+            batch = await replay_buffer.sample.choose(
+                curr_policy_version=policy_version
+            )
             if batch is None:
                 await asyncio.sleep(0.1)
             else:
-                training_result = await trainer.train_step.choose(batch)
+                loss = sum(await trainer.train_step.call(batch))
```
pbontrager: This shouldn't be returning a list to sum over right? Their should be 1 replica for this service?

apps/grpo/main.py
```diff
@@ -483,10 +519,10 @@ async def continuous_training():
         )
 
 
-@parse
-def recipe_main(cfg: DictConfig) -> None:
-    asyncio.run(main(cfg))
+if __name__ == "__main__":
 
+    @parse
```
pbontrager, line 483: Can you explain this change?

src/forge/actors/policy.py
```diff
@@ -239,12 +236,24 @@ def start_processing(self):
             self._run_task = asyncio.create_task(self.run())
 
     @endpoint
-    async def generate(self, prompt: str, priority: int = 0) -> List[CompletionOutput]:
+    async def generate(self, prompt: str, priority: int = 0) -> RequestOutput:
```
pbontrager, line 236: Why was this necessary? This pattern is never great for readability

src/forge/actors/policy.py
```diff
@@ -428,23 +437,19 @@ async def _load_tensor_parallel_state_dict(
...
-        )
-
+        key = f"{self.state_dict_key}{DELIM}{version}"
         model = self.worker.model_runner.model
         current_state_dict = model.state_dict()
-
+        start = time.time()
         await self._load_tensor_parallel_state_dict(current_state_dict, version)
-        logger.debug("Successfully updated model weights from torchstore")
+        self.logger.debug(
```
pbontrager, line 440: Better to log this to wandb then spam the terminal

src/forge/actors/policy.py
```diff
@@ -118,6 +114,7 @@ def __post_init__(self):
         self._policy_proc: ProcMesh | None = None
         self._worker_procs: ProcMesh | None = None
         self.weights_version: int = 0
+        self.running: bool = False
```
pbontrager: Where is this used?

src/forge/actors/policy.py
```diff
@@ -331,25 +340,28 @@ async def run(self):
                 engine_core_timestamp=outputs.timestamp,
                 iteration_stats=None,
             )
+
             for request_output in processed_outputs.request_outputs:
                 if request_output.finished:
-                    _, fut = self.requests.pop(request_output.request_id)
-                    fut.set_result(request_output)
+                    if request_output.request_id in self.requests:
```
pbontrager: When is this not true?

apps/grpo/main.py
```diff
@@ -148,78 +143,115 @@ class Trainer(ForgeActor):
...
 
-        return {"loss": loss.item()}
+        return loss.detach()
 
     @endpoint
-    async def push_weights(self):
-        pass
+    async def push_weights(self, version: int):
+        """Update policy model weights with trainer's current weights."""
+        start_time = time.time()
```
allenwang28:
```suggestion
        # TODO - issues/148 followup
        start_time = time.time()
```

Just for my future reference, tagging some pieces for observability

apps/grpo/main.py
```diff
@@ -243,18 +275,11 @@ class ComputeAdvantages(ForgeActor):
     @endpoint
     async def compute(self, group: Group) -> list[float]:
         # TODO: add batch processing
```
allenwang28, line 238:
```suggestion
        # TODO: issues/120 add batch processing
```

src/forge/actors/policy.py
```diff
@@ -428,23 +437,19 @@ async def _load_tensor_parallel_state_dict(
...
-
-        logger.debug(
-            f"Starting model update from torchstore with key: {self.state_dict_key}{DELIM}{version}"
-        )
-
+        key = f"{self.state_dict_key}{DELIM}{version}"
         model = self.worker.model_runner.model
         current_state_dict = model.state_dict()
-
+        start = time.time()
```
allenwang28, line 438:
```suggestion
        # TODO - issues/148
        start = time.time()
```

src/forge/actors/policy.py
allenwang28: why do we need to do `self.logger` instead of just `logger`?

apps/grpo/main.py
allenwang28: Service call currently returns a list regardless of num_replicas

apps/grpo/main.py
joecummings: u right

src/forge/actors/policy.py
joecummings: With the creation of ForgeActor, each actor has it's own logger and should use that (b/c it logs information about which actor is logging) instead of a global logger.

src/forge/actors/policy.py
joecummings: It's used in the generation loop, was missing before but technically should be false before we start generating things.

src/forge/actors/policy.py
allenwang28: ah, the ForgeActor logger was implemented in a way where it's supposed to just work with `logger` (ie no need to `self.logger`) but please let me know if that doesn't work

apps/grpo/main.py
pbontrager: I was wrong earlier, this should be a choose call. I think only policy.update_weights should be a call

src/forge/actors/trainer.py
```diff
@@ -268,3 +268,49 @@ def push_weights(self) -> None:
     async def cleanup(self) -> None:
         if self.engine.checkpointer:
             self.engine.checkpointer.close()
+
+
+def _qwen3_hf_to_vllm(
```
Ritesh1905, line 273: May be for the next PR: Thoughts on having this in the generator/policy side? Ideally trainer should be agnostic to how the poilicy packs/transforms certain tensors.

src/forge/data/rewards.py
```diff
@@ -17,62 +16,69 @@ def __init__(self, tolerance: float = 1e-6, partial_credit: float = 0.1):
...
-            matches = re.findall(pattern, text)
-            if matches:
-                return self._to_float(matches[-1])
-        return None
-
     def __call__(self, prompt: str, response: str, target: str) -> float:
         """Compute math correctness reward."""
-        # Parse expected
-        expected_answer = self._to_float(target)
+        target_number = self._to_float(target)
```
Ritesh1905, line 21: Curious: Do we a plan to have math verifier libraries which provides certain things (ExactMatch, partial match etc..) out of the box?

PR #142, opened 2025-09-10
fix the path for the tokenizer

AlirezaShamsoshoara:
Fixing the default path for the llama3.1-8B tokenizer as the everything regarding the model will be downloaded to `/tmp/Meta-Llama-3.1-8B-Instruct`.


Jack-Khuu:
Already in main

PR #146, opened 2025-09-10
Trainer Loss

pbontrager:
- Updates trainer for GRPO Loss
- Update replay buffer to do batching
- Update rl app test to do one training step (This app is meant for testing)

The goal of this PR is to enable titan training as well as keep the data definition and the loss easily accessible to the user. The loss, data type, and collation are left to the user so they can co-design all three to their specific data and are less likely to have to touch the trainer.

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -206,9 +206,7 @@ class RLTrainer(ForgeActor):
             self.engine.checkpointer.close()
 
 
-def _qwen3_hf_to_vllm(
-    sd: dict[str, torch.Tensor], num_layers: int
-) -> dict[str, torch.Tensor]:
+def _qwen3_hf_to_vllm(sd: dict[str, Tensor], num_layers: int) -> dict[str, Tensor]:
     """Convert transformers state dict to vLLM format. Specifically, this fuses
     QKV projection and MLP gate_up_proj layers.
 


src/forge/actors/trainer.py
```diff
@@ -155,30 +155,34 @@ def forward_backward(
...
     @endpoint
-    def train_step(self, batch) -> None:
-        # Move tensors to the appropriate device
-        for k, v in batch.items():
-            if isinstance(v, torch.Tensor):
-                batch[k] = v.to("cuda")  # TODO: hardcoded for now
-
+    def train_step(
+        self, inputs: list[dict[Tensor]], targets: list[dict[Tensor]]
+    ) -> None:
```
Jack-Khuu, line 169: Update return type

apps/rl/main.py
```diff
@@ -13,32 +13,169 @@
...
+    @property
+    def response_tensor(self):
+        tensor = torch.tensor(self.response_tokens, dtype=torch.long)
+        if tensor.shape[0] < self.response_len:  # right pad
+            diff = self.response_len - tensor.shape[0]
+            tensor = F.pad(tensor, (0, diff), value=self.pad_id)
+        return tensor
+
+
+def collate(batches: list[list[Episode]]):
```
Jack-Khuu, line 66: return type?

apps/rl/main.py
```diff
@@ -13,32 +13,169 @@
...
+            collate=collate,
             **cfg.replay_buffer,
         ),
     )
-    print("Services initialized....")
+    print("Services initialized...")
+
+    print("Collecting Data...")
+    g = torch.manual_seed(0)
+    global_batch_size = cfg.replay_buffer.batch_size * cfg.replay_buffer.dp_size
```
allenwang28, line 156: no need to change right now, but wouldn't the global batch size technically be a trainer config and not a replay buffer config? Is it possible to tell omegaconf like "make this value the same as the one defined elsewhere"?

pbontrager: Not a trainer config anymore since replay_buffer is it's own service and the trainer doesn't handle data or loading anymore

PR #147, opened 2025-09-11
Refactor replay buffer to use KV buffer

DNXie:
This PR refactors replay buffer storage to KV-based buffer. 
We can’t integrate torchstore yet since the necessary APIs aren’t implemented (numel, delete, etc.). This PR refactors replay buffer storage to KV-based data structure, so we’ll be able to switch the backend easily once torchstore is ready.

- Added `StoreInterface` for KV store abstraction and further integration of torchstore
- Implemented `KVStore` as a temporary KV store backend
- Added unit tests for `KVStore` in `test_kv_store.py`
- Refactored `ReplayBuffer` to use`StoreInterface` instead of a local list
- Updated `test_replay_buffer.py` and `test_toy_rl.py` accordingly
- Updated all callsites of `ReplayBuffer` in `apps.grpo.main`, `apps.rl.main`, `apps.toy_rl.main`

**Test**
```
pytest tests/unit_tests/test_kv_store.py
pytest tests/unit_tests/test_replay_buffer.py
pytest tests/unit_tests/rl/test_toy_rl.py
python -m apps.rl.main --config apps/rl/llama3_8b.yaml
python -m apps.toy_rl.main
```


src/forge/actors/replay_buffer.py
```diff
@@ -50,18 +56,21 @@ async def sample(self, curr_policy_version: int, batch_size: int | None = None):
...
             return None
 
+        keys = await self.store.keys()
+
         # TODO: Make this more efficient
-        idx_to_sample = self.sampler(range(len(self.buffer)), k=total_samples)
+        idx_to_sample = self.sampler(range(len(keys)), k=total_samples)
         # Pop episodes in descending order to avoid shifting issues
-        popped = [self.buffer.pop(i) for i in sorted(idx_to_sample, reverse=True)]
+        sorted_idxs = sorted(idx_to_sample, reverse=True)
```
casteryh: sorting no longer needed with a KV store

src/forge/actors/replay_buffer.py
```diff
@@ -5,34 +5,40 @@
...
         random.seed(self.seed)
         self.sampler = random.sample
 
     @endpoint
     async def add(self, episode) -> None:
-        self.buffer.append(episode)
+        await self._add(episode)
+
+    async def _add(self, episode) -> None:
+        key = f"rb_ep_{await self.store.numel()}_{uuid.uuid4().hex}"
```
casteryh: What's the point of `await self.store.numel()`?
Also this may be expensive.

src/forge/actors/replay_buffer.py
```diff
@@ -81,38 +90,46 @@ async def evict(self, curr_policy_version: int) -> None:
...
-        self.buffer = [
-            trajectory
-            for trajectory in self.buffer
-            if (curr_policy_version - trajectory.policy_version) <= self.max_policy_age
-        ]
+    async def _evict(self, curr_policy_version: int) -> None:
+        keys = await self.store.keys()
+        for key in keys:
+            episode = await self.store.get(key)
+            if (curr_policy_version - episode.policy_version) > self.max_policy_age:
```
casteryh: You can store the policy_version in the key, so you don't need to do a `get` here.

src/forge/actors/replay_buffer.py
```diff
@@ -81,38 +90,46 @@ async def evict(self, curr_policy_version: int) -> None:
...
+    async def _clear(self) -> None:
+        await self.store.delete_all()
 
     @endpoint
     async def state_dict(self) -> dict[str, Any]:
+        keys = await self.store.keys()
+        episodes = [await self.store.get(k) for k in keys]
         return {
-            "buffer": self.buffer,
+            "buffer": episodes,
```
casteryh, line 119: If you want actual full recovery, you should also store the key here. Or derive the key in a deterministic way.

src/forge/actors/replay_buffer.py
casteryh: Re: recovery and determinism, maybe you could use uuid5 or something like highway hash. But I am not sure how important is determinism.

src/forge/data/stores.py
```diff
@@ -0,0 +1,68 @@
...
+# LICENSE file in the root directory of this source tree.
+
+
+from typing import Any
+
+<<<<<<< HEAD
+from forge.interfaces import StoreInterface
+=======
+from src.forge.interfaces import StoreInterface
+>>>>>>> 05dd33b7eb5574b9db6f8ec0ee2417919946b0fe
```
casteryh: please resolve

tests/unit_tests/test_kv_store.py
```diff
@@ -0,0 +1,147 @@
...
+            await store.pop("nonexistent_key")
+
+    @pytest.mark.asyncio
+    async def test_overwrite_existing_key(self, store: KVStore) -> None:
+        """Test overwriting an existing key."""
+        await store.put("key1", "value1")
+        await store.put("key1", "new_value")
+
+        result = await store.get("key1")
+        assert result == "new_value"
```
casteryh: The test is fine as is.
I just want to note that maybe we should never change the value of an existing key. We don't know how much consistency guarantee torchstore provides at this point. Doing this is playing with fire.

tests/unit_tests/test_replay_buffer.py
```diff
@@ -29,7 +36,6 @@ async def test_add(self, replay_buffer: ReplayBuffer) -> None:
         trajectory = Trajectory(policy_version=0)
         await replay_buffer.add.call_one(trajectory)
         assert replay_buffer._numel.call_one().get() == 1
```
casteryh, line 37: please add back test.
To retrieve the element, you can do `replay_buffer.sample(1)`.

tests/unit_tests/test_replay_buffer.py
```diff
@@ -39,8 +45,6 @@ async def test_add_multiple(self, replay_buffer) -> None:
         await replay_buffer.add.call_one(trajectory_0)
         await replay_buffer.add.call_one(trajectory_1)
         assert replay_buffer._numel.call_one().get() == 2
```
casteryh, line 52: test

src/forge/actors/replay_buffer.py
casteryh: Maybe you can add a counter in the ReplayBuffer class

Then derive the key using uuid5 and/or highway hash and/or your favorite hash, with the following 3 pieces of information
- the counter
- the rank of the current worker
- the content of the value.

This will generally avoid duplicate keys even if you have episodes with the same content coming in.

tests/unit_tests/test_replay_buffer.py
```diff
@@ -113,9 +117,15 @@ async def test_sample_with_evictions(self, replay_buffer) -> None:
...
         replay_buffer = await mesh.spawn(
-            "replay_buffer", ReplayBuffer, batch_size=2, max_policy_age=1, dp_size=3
+            "replay_buffer",
+            ReplayBuffer,
+            store=store,
+            batch_size=2,
+            max_policy_age=1,
+            dp_size=3,
         )
         await replay_buffer.setup.call()
```
casteryh, line 160: Consider adding tests for save/load of state dicts and verify it does have deterministic behavior.

src/forge/actors/replay_buffer.py
DNXie: I don't think we need deterministic at this stage. Let's keep things simple. I have dropped the `await self.store.numel()` to make things efficient. Thanks for pointing this out.

src/forge/actors/replay_buffer.py
```diff
@@ -50,18 +56,21 @@ async def sample(self, curr_policy_version: int, batch_size: int | None = None):
...
+        await self._evict(curr_policy_version)
 
-        if total_samples > len(self.buffer):
+        total_available = await self.store.numel()
+        if total_samples > total_available:
             return None
 
+        keys = await self.store.keys()
+
         # TODO: Make this more efficient
```
DNXie: @joecummings Do we still need this TODO?

src/forge/actors/replay_buffer.py
DNXie: I suppose this requires supporting `prefix` in the `get` method? Are we sure that torchstore is going to support that?

tests/unit_tests/test_kv_store.py
DNXie: Dropped.

src/forge/actors/replay_buffer.py
casteryh: Even if they don't, you can still define a function like (not tested lol, maybe don't use regex)
```
self._policy_version_pattern = re.compile(r"^replay_buffer\.policy_version_(\d+)")
def _parse_policy_version(self, key)
     return re.search(self._policy_version_pattern, key)
```
so that you can just get the keys and just do `_parse_policy_version(key)` (and skip or raise an error if there is no match, depending on your assumption about what's in the store)
without actually getting the policy_version.

src/forge/actors/replay_buffer.py
DNXie: Updated the code.

tests/unit_tests/test_replay_buffer.py
DNXie: Added.

src/forge/actors/replay_buffer.py
DNXie: Thanks for catching this. removed. Now the code looks much cleaner

src/forge/interfaces.py
```diff
@@ -208,6 +208,134 @@ def __call__(self, observation: Observation) -> float:
         pass
 
 
+class StoreInterface(ABC):
```
joecummings, line 211: For now, pair this down to the *exact* APIs we will be using in the ReplayBuffer - no more, no less. We can always update the interface later.

src/forge/actors/replay_buffer.py
```diff
@@ -81,38 +86,51 @@ async def evict(self, curr_policy_version: int) -> None:
...
-        self._evict(curr_policy_version)
+        await self._evict(curr_policy_version)
 
-    def _evict(self, curr_policy_version: int) -> None:
-        self.buffer = [
-            trajectory
-            for trajectory in self.buffer
-            if (curr_policy_version - trajectory.policy_version) <= self.max_policy_age
-        ]
+    async def _evict(self, curr_policy_version: int) -> None:
```
joecummings, line 89: We control this endpoint, so you can just pass in the keys that we already calculated above.

src/forge/actors/replay_buffer.py
```diff
@@ -81,38 +86,51 @@ async def evict(self, curr_policy_version: int) -> None:
...
+        keys = await self.store.keys()
+        for key in keys:
+            episode = await self.store.get(key)
```
joecummings: This is fine for now, but leave a comment that we could store each key as a uuid + the policy version and make this more efficient.

src/forge/actors/replay_buffer.py
```diff
@@ -81,38 +86,51 @@ async def evict(self, curr_policy_version: int) -> None:
...
+                await self.store.delete(key)
 
     @endpoint
-    async def _getitem(self, idx: int):
-        return self.buffer[idx]
+    async def _getitem(self, key: str):
+        return await self.store.get(key)
+
+    @endpoint
+    async def items(self) -> list[tuple[str, Any]]:
```
joecummings: Why do we need to add an items method here? I'd prefer to constrain until a use-case arrives.

src/forge/data/stores.py
```diff
@@ -0,0 +1,64 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+from typing import Any
+
+from src.forge.interfaces import StoreInterface
+
+
+class KVStore(StoreInterface):
```
joecummings, line 13: Similar to my other comment, pair this down as much as possible especially b/c it's pretty much just a dictionary.

src/forge/actors/replay_buffer.py
```diff
@@ -5,34 +5,40 @@
...
             self.seed = random.randint(0, 2**32)
         random.seed(self.seed)
         self.sampler = random.sample
 
     @endpoint
     async def add(self, episode) -> None:
-        self.buffer.append(episode)
+        await self._add(episode)
+
+    async def _add(self, episode) -> None:
```
joecummings: Do we need this as an internal method?

src/forge/actors/replay_buffer.py
joecummings, line 89: We control this internal method so you can pass in the keys from above, which we already calculated.

src/forge/interfaces.py
DNXie: I’ve already pared the interface down to only the APIs we need. My concern is that methods like `numel` and `delete` are essential for the replay buffer’s functionality (e.g., eviction, checking buffer size) but aren’t yet implemented in TorchStore. If we remove these from the interface, the buffer implementation won’t be able to operate consistently.

src/forge/actors/replay_buffer.py
DNXie: Yes. Added. 
Once torchstore support fetching with "prefix", this would be much easier.

src/forge/actors/replay_buffer.py
DNXie: We actually don't need. Removed

src/forge/actors/replay_buffer.py
DNXie: It was used in a test case. Now I have updated the test case to not use it. Removed the method as well.

src/forge/actors/replay_buffer.py
DNXie: Let's leave this to future work. Things would be easier if torchstore supports prefix.

src/forge/data/stores.py
DNXie: Replied under your other comment

src/forge/actors/replay_buffer.py
```diff
@@ -5,34 +5,36 @@
...
+    store: StoreInterface
     batch_size: int
     max_policy_age: int
     dp_size: int = 1
     seed: int | None = None
 
-    @endpoint
-    async def setup(self) -> None:
-        self.buffer: list = []
+    def __post_init__(self):
```
DNXie, line 29: @joecummings  I changed the `setup` to `post_init` because I found that the `setup` is not called in many of the scripts when we are using ReplayBuffer. And if it is not called, things may not be initialized correctly (e.g., sampler). Let me know if this cause any concerns.

src/forge/actors/replay_buffer.py
DNXie: Actually, we have to re-fetch the keys after eviction because `_evict` may delete some entries. We fetch once before `_evict` to know what to check for eviction, then fetch again after to ensure we only sample from the remaining keys. This prevents trying to access keys that no longer exist.

src/forge/actors/replay_buffer.py
casteryh: As a general comment: `_evict()` before getting keys is not a reliable way to ensure we don't get outdated policies.
Since we have several await points between `_evict()` and `keys()`. Unless you want to put an async lock on self.store, which you probably don't. This is beyond the scope of this PR though, please and an TODO here. cc @joecummings

casteryh: Even nothing is concurrent at all at this point. I think we should at least keep in mind we will need to support concurrency in the very near future. Also it's not necessarily harder to write concurrently correct program. Albeit we do need to be more careful.

src/forge/actors/replay_buffer.py
casteryh: I think we need to reconsider the whole eviction logic. See my previous point re: concurerncy.

src/forge/actors/replay_buffer.py
casteryh: > Yes. Added. Once torchstore support fetching with "prefix", this would be much easier.

Coming soon!

src/forge/actors/replay_buffer.py
DNXie: Good point. Will add this TODO before landing.

src/forge/actors/replay_buffer.py
```diff
@@ -5,34 +5,36 @@
...
 
 from forge.controller import ForgeActor
+from forge.interfaces import StoreInterface
 
 
 @dataclass
 class ReplayBuffer(ForgeActor):
     """Simple in-memory replay buffer implementation."""
 
+    store: StoreInterface
```
joecummings: nit: maybe call this the backend?

wdyt @LucasLLC ?

src/forge/actors/replay_buffer.py
```diff
@@ -81,38 +83,47 @@ async def evict(self, curr_policy_version: int) -> None:
...
-        self.buffer.clear()
+        await self._clear()
+
+    async def _clear(self) -> None:
+        await self.store.delete_all()
 
     @endpoint
     async def state_dict(self) -> dict[str, Any]:
+        keys = await self.store.keys()
+        episodes = [(k, await self.store.get(k)) for k in keys]
```
joecummings: This is not ideal IMO - is there a way we could dump / serialize the contents of the store?

@LucasLLC ?

src/forge/actors/replay_buffer.py
DNXie: Changed `store` -> `backend`

src/forge/actors/replay_buffer.py
DNXie: Added

src/forge/data/stores.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    A simple single-node key-value (KV) store implementation of StoreInterface.
+
+    This acts as a temporary backend for the replay buffer until torchstore
+    supports the full set of operations we need (delete, pop, keys, numel, etc.).
```
vidhyav, line 18: Can you explain the consistency semantics and the thread safety of the interface?

For example, once put finishes, any future gets will always see the effect of the put.

DNXie: The current `KVStore` implementation is a simple in-memory dictionary with async methods, intended as a temporary backend until `torchstore` is ready. It does not provide thread safety or strong consistency guarantees in the presence of concurrent access. Specifically:

* If multiple coroutines access or modify the store concurrently, race conditions may occur (e.g., a `get` may see stale or missing data if a `delete` or `put` happens at the same time).
* In a single-threaded `asyncio` event loop, as long as each operation is awaited, the store behaves as expected: once a `put` completes, subsequent `get`s will see the new value.
* However, if the store is accessed from multiple threads or if multiple async tasks interleave operations without awaiting, consistency is not guaranteed.

The plan is to switch to torchstore once the key APIs like `delete` and `numel` are ready, which should provide proper concurrency and consistency guarantees.

PR #149, opened 2025-09-11
[RFC] Defining core abstractions

Ritesh1905:
Summary:
- Defining core abstractions. 
- Hoping to use this PR to get feedback on the abstractions
- Not picky about the terminology (Prompt vs experience vs completion vs minibatch etc.) So please suggest.

Test Plan:
Unit tests in the next PR once we have the stubs implemented

### DIFF
--- a/src/forge/data_models/api.py
+++ b/src/forge/data_models/api.py
@@ -5,7 +5,8 @@
 # LICENSE file in the root directory of this source tree.
 
 from abc import ABC, abstractmethod
-from typing import Dict, Sequence, Tuple
+from collections.abc import Iterator, Sequence
+from typing import Dict, Tuple
 
 import torch
 
@@ -15,6 +16,7 @@ from forge.data_models.minibatch import Minibatch
 
 from forge.data_models.prompt import Prompt
 from forge.data_models.scored_completion import ScoredCompletion
+from torch.utils.data import DataLoader, IterableDataset
 
 
 # TODO: This file needs should not be in the data_models folder/package
@@ -86,3 +88,24 @@ class Scorer(ABC):
         Optionally override for efficient batch scoring.
         """
         return [self.score(c) for c in completions]
+
+
+class PromptDataset(IterableDataset):
+    """
+    Users should inherit from this and implement __iter__.
+    """
+
+    def __iter__(self) -> Iterator[Prompt]:
+        """
+        defines how to generate or yield SimpleElement objects.
+        """
+        raise NotImplementedError
+
+
+class PromptDataLoader(DataLoader):
+    """
+    subclass of DataLoader to handles batching, parallelism, and other data serving concerns.
+    """
+
+    def __init__(self, dataset: PromptDataset, **kwargs):
+        super().__init__(dataset, **kwargs)


src/forge/data_models/minibatch.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def from_experiences(
+    exps: Sequence[Experience], max_seq_len: int, pad_val: int = 0
+) -> Minibatch:
+    """
+    Convert a list of experiences to a minibatch.
+    """
+
+    def pack_sequence(
```
felipemello1, line 49: should we use packed dataset? https://github.com/meta-pytorch/forge/blob/029b1bca5a4ac4c65ee254acee3e4ce0a48c88e9/src/forge/data/datasets/packed.py#L508

src/forge/data_models/api.py
```diff
@@ -0,0 +1,111 @@
...
+    def snapshot_weights(
+        self,
+    ) -> Dict[str, Tuple[torch.Tensor, torch.dtype, torch.Size]]:  # TODO: RDMA buffer
+        """
+        applying accumulated gradients to the model parameters.
+
+        the return type is a tuple of weights buffer, dtype, and shape of the original tensor.
+        """
+        # TODO: NEEDS fixing: the weights_handle should be remote handle, like RDMA Buffer handle
+        pass
```
casteryh, line 50: why do we also apply gradients in snapshot_weights here? Or is it just wrong docstring?

src/forge/data_models/api.py
```diff
@@ -0,0 +1,111 @@
...
+class Generator(ABC):
+    @abstractmethod
+    def generate(self, prompt: Prompt, **kwargs) -> list[Completion]:
+        """
+        Generate a completion given a prompt.
+        Args:
+            prompt: The input prompt.
+            **kwargs: Additional model-specific generation parameters.
+        Returns:
+            str: The generated text.
```
casteryh, line 62: Is this supposed to be one completion or multiple completions? docstring says one completion, return type says list of completions.
I think list of completions makes more sense.

src/forge/data_models/minibatch.py
Ritesh1905: Thanks Felipe, I will take a look to see if I can use packed dataset instead?

src/forge/data_models/distributed_metric.py
```diff
@@ -0,0 +1,64 @@
...
+
+import torch
+import torch.distributed as dist
+
+
+@dataclass
+class DistributedMetric(ABC):
+    """Metrics that are calculated in distributed fashion.
+
+    Metrics computed in each rank are going to be wrapped in DistributedMetric
```
casteryh, line 18: I get the point but I am confused about the api.
Can you give a more detailed example here?

src/forge/data_models/experience.py
```diff
@@ -0,0 +1,68 @@
...
+
+from dataclasses import dataclass
+from typing import Optional, Sequence
+
+import torch
+from forge.data_models.scored_completion import ScoredCompletion
+
+
+@dataclass
+class Experience:
```
casteryh, line 15: Let's say we have a multi-turn conversation. Does that correspond to one Experience, or multiple Experiences?

casteryh: For example, are we supposed to have the following mask?
`[0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0, 1, 1, 1, 0,  0, 1, 1]`

src/forge/data_models/prompt.py
```diff
@@ -0,0 +1,66 @@
...
+@dataclass
+class Message:
+    """A single message in a conversation."""
+
+    chunks: Sequence[str]
+    role: Role
+
+
+@dataclass
+class Prompt:
```
casteryh, line 29: I don't know about the name `Prompt`.

src/forge/generators/vllm_generator.py
```diff
@@ -0,0 +1,30 @@
...
+from typing import List
+
+import torch
+from forge.data_models.api import Generator
+
+from forge.data_models.completion import Completion
+from forge.data_models.prompt import Prompt
+
+
+class VLLMGenerator(Generator):
```
casteryh, line 16: This and the Generator in api.py are supposed to be Monarch Actors, no?

src/forge/data_models/api.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def update_weights(
+        self, weights_handle: dict[str, tuple[torch.Tensor, torch.dtype, torch.Size]]
```
allenwang28, line 68: Our `update_weights` implementation is very similar, except the weights handle is essentially a handle [given and tracked by torchstore](https://github.com/meta-pytorch/forge/blob/c50a2a1b6a43f53a097c22558c4c9ec3e1de3246/src/forge/actors/policy.py#L426)

although now that I look at it, we should probably be passing the version explicitly rather than tracking implicitly in the policy cc @joecummings @pbontrager @pradeepfn

src/forge/data_models/api.py
```diff
@@ -0,0 +1,111 @@
...
+from forge.data_models.scored_completion import ScoredCompletion
+from torch.utils.data import DataLoader, IterableDataset
+
+
+# TODO: This file needs should not be in the data_models folder/package
+
+
+class Trainer(ABC):
+    @abstractmethod
+    def accummulate_gradients(self, minibatch: Minibatch) -> LossOutput:
```
allenwang28, line 27: why are `accumulate/apply` kept separate? 

aside, we should be able to implement this in our Titan trainer by pulling out the [`loss.backwards`](https://github.com/meta-pytorch/forge/blob/main/src/forge/actors/trainer.py#L162) into its own function

cc @wwwjn @felipemello1

src/forge/data_models/api.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @abstractmethod
+    def apply_gradients(self) -> None:
+        """
+        applying accumulated gradients to the model parameters.
+        """
+        pass
+
+    @abstractmethod
+    def snapshot_weights(
```
allenwang28, line 41: Is the contract here that the snapshotted weights never change so long as this handle exists?

src/forge/data_models/api.py
```diff
@@ -0,0 +1,111 @@
...
+
+        the return type is a tuple of weights buffer, dtype, and shape of the original tensor.
+        """
+        # TODO: NEEDS fixing: the weights_handle should be remote handle, like RDMA Buffer handle
+        pass
+
+
+class Generator(ABC):
+    @abstractmethod
+    def generate(self, prompt: Prompt, **kwargs) -> list[Completion]:
```
allenwang28, line 55: our [`generate`](https://github.com/meta-pytorch/forge/blob/c50a2a1b6a43f53a097c22558c4c9ec3e1de3246/src/forge/actors/policy.py#L248) returns the `RequestOutput` in order to access the tokenized input IDs cc @pbontrager @Jack-Khuu 

(but I see that the `Completion` is **not** the vLLM `CompletionOutput`)

src/forge/data_models/experience.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    The Experience data class to be used by the trainer.
+
+    Experiences are usually generated from a scored completion and running various post processing steps.
+    """
+
+    # Concatenated prompt and sample token ids.
+    ids: torch.Tensor
```
allenwang28, line 23: we currently [do this concat](https://github.com/meta-pytorch/forge/blob/c50a2a1b6a43f53a097c22558c4c9ec3e1de3246/apps/grpo/main.py#L177-L189) in the `train_step` which I don't love

src/forge/data_models/experience.py
allenwang28, line 15: overall I like this representation. It seems like the right primitive to pass to the trainer, avoiding the trainer having to care about implementation details on getting to the right format.

Also like the idea of `from(...)` but my stylistic preference is like
```
@dataclass
class Experience:

    @classmethod
    def from_scored_completions(cls, ...) -> Experience:
        ...
```

cc @joecummings

src/forge/data_models/distributed_metric.py
```diff
@@ -0,0 +1,64 @@
...
+    according to how they are going to be aggregated. For example, average log prob
+    can be wrapped as `Fraction(Sum((logp * mask).sum()), Sum(mask.sum()))` where
```
allenwang28, line 20:
```
Fraction(Sum((logp * mask).sum()), Sum(mask.sum()))
```

Is this how it's represented? I get the intention but I kind of hate the math DSL it implies lol

cc @felipemello1 

I wonder if there's any way we can use DTensor for this more elegantly, cc @LucasLLC

src/forge/data_models/completion.py
```diff
@@ -0,0 +1,31 @@
...
+
+from dataclasses import dataclass
+from typing import Optional
+
+import torch
+from forge.data_models.prompt import Prompt
+
+
+@dataclass
+class Completion:
```
allenwang28, line 15: similar to `Experience`, I think we should incorporate atomic data representations into Forge with implemented converters like
```
@dataclass
class Completion:
    def from_vllm_outputs(cls, RequestOutput) -> "Completion":
        ...
```

cc @pbontrager  @Jack-Khuu  @joecummings

src/forge/data_models/distributed_metric.py
felipemello1: I already did some work on metric logging aggregation. Its well tested. I think we should check if we can/should use it here: https://github.com/meta-pytorch/forge/tree/main/src/forge/data/dataset_metrics#2-metricsaggregator

When i did it, it was heavily focused on dataset, but it doesnt have to be. It should work for observability, rewards, etc.

src/forge/data_models/api.py
Ritesh1905: Mainly computational efficiency. 
This method would help reduces the frequency of expensive parameter updates
and potentially enables better utilization of hardware (Can batch operations like all-reduce in distributed scenarios). Essentially for every batch/minibatch you can don't need to update the params.. you can just accumulate them and then apply once a suitable number of batches has been trained.

Something like below...

```
For step in range(20):
    for mini_batch in range(10):
         # accumulate for N mini batches
          trainer.accummulate_gradient(mini_batch).
          
    trainer.apply_graidents()
        
```

We can expose another `step` API in trainer... which would be a combination of accumulate and apply.

src/forge/data_models/api.py
Ritesh1905: Yes. For a single RL loop run. So long the learner and policy have the same remote handle reference. then this can be achieved in 3 steps

1. At the init, controller provides the same remote handle to learner and policy
2. When it's time to update weights, controller requests trainer to push weights to remote buffer.
3. Controller requests policy to read the buffer and update weights.

src/forge/data_models/api.py
Ritesh1905: Good catch. Wrong doc string, needs fixing.

src/forge/data_models/api.py
Ritesh1905: > (but I see that the Completion is not the vLLM CompletionOutput)

This is intentional and the idea is to have the api very generic. Like we can have sglang/HF bases generators and not just vLLM based.

src/forge/data_models/api.py
Ritesh1905: It is list of completions. doc string needs fixing.

src/forge/data_models/api.py
Ritesh1905: Right. I have a TODO in the doc string, this needs to be a handle.

Also this needs to be abstracted to. Like we need to have a weights buffer abstractions and the implementation can figureout if it's a handle, raw/bytes etc. 

Will update.

src/forge/data_models/distributed_metric.py
Ritesh1905: I have something in my private branch which provides a motivation for how this will be used. https://github.com/meta-pytorch/forge/blob/rithesh/reinforce/src/forge/trainers/huggingface_trainer.py#L61-L119

Happy to chat further.

src/forge/data_models/distributed_metric.py
Ritesh1905: I have something in my private branch which provides a motivation for how this will be used. https://github.com/meta-pytorch/forge/blob/rithesh/reinforce/src/forge/trainers/huggingface_trainer.py#L61-L119

This is specifically needed if you are using the computational efficient accummulate and apply gradients approach rather than the single train step for every batch. Happy to chat further.

src/forge/data_models/api.py
allenwang28: got it, that makes sense to me

src/forge/data_models/prompt.py
Ritesh1905: I am not picky about the names. The RL world has tons of jargon. Happy to pick one based on what team, agrees. (applies to everything in this diff)

src/forge/generators/vllm_generator.py
Ritesh1905: Nope. The intention is to distinguish between the implementation and the actors would be a very skinny version, where they are injecting a specific generator/trainer. keeping the actor as skinny as possible allows us to inject any generator implementation.

```
class Policy(Actor):
    def __init__(self, generator: Generator):
        super().__init__()
        rank = current_rank()
        self.rank = rank.rank
        self.local_rank = rank["gpus"]
        self.world_size = rank.extent.nelements
        self._set_env_vars()
        self.generator = generator

    @endpoint
    async def generate(self, prompt: Prompt) -> list[Completion]:
        return self.generator.generate(prompt)

    @endpoint
    async def update_weights(
        self, weights_buffer: dict[str, tuple[torch.Tensor, torch.dtype, torch.Size]]
    ):
        return self.generator.update_weights(weights_buffer)

    def _set_env_vars(self):
        os.environ["WORLD_SIZE"] = str(self.world_size)
        os.environ["RANK"] = str(self.rank)
        os.environ["LOCAL_RANK"] = str(self.local_rank)
```

src/forge/data_models/api.py
allenwang28: ok, yeah that makes sense. We're calling it [`push_weights`](https://github.com/meta-pytorch/forge/blob/a6ca59173a58c51de7db2e1fd27d942f56203756/apps/grpo/main.py#L211) now but `snapshot` may be more accurate

src/forge/data_models/api.py
allenwang28: ok our titan integration doesn't yet have grad accumulation: https://github.com/meta-pytorch/forge/pull/146 - it seems possible to add this API surface along with the high level `step`

src/forge/data_models/experience.py
Ritesh1905: The Experience class is generic and is agnostic to single-turn or multi-turn structure. It is up to your data processing pipeline to decide how to chunk conversations into Experience instances.

If you want to train on multi-turn conversations, you would typically:
- Concatenate all turns into a single sequence of token ids.
- Set the mask and weights appropriately 
- Store this as a single Experience.

OR

If you want to treat each turn as a separate training example, you could split the conversation into multiple Experience instances, one per turn.

Jack-Khuu:
Different Version of this has landed in main

PR #150, opened 2025-09-11
Some minor cleanup to  README.md

JenniferWang:
As title suggest. this is based on my latest attempt to set up the dev enviornment

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,9 @@ conda activate forge
 
 Optional: By default, the packages installation uses conda. If user wants to install system packages on the target machine instead of conda, they can pass the `--use-sudo` to the installation script: `./script/install.sh --use-sudo`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,22 +22,18 @@ conda activate forge
...
 ```
 
 If you need to re-build the wheels for whatever reason, you can do so with:
 ```bash
 ./scripts/build_wheels.sh
 ```
 
-Since the vLLM wheel is too large for GitHub, we uploaded it as a release:
-```
-$ gh release create v0.0.0 assets/wheels/vllm-*.whl --title "Forge Wheels v0.0.0"
```
allenwang28, line 37: Can we leave this line in? If/when I need to run this again I use this to copy/paste lol

Issue #144, opened 2025-09-10
HostMesh integration


allenwang28:
# Context
[`HostMesh`](https://github.com/meta-pytorch/monarch/blob/fe9b99947779b3755c37f8fa65edea79d19b32a0/python/monarch/_src/actor/host_mesh.py#L48) in Monarch represents a "Host" resource, providing a utility to spawn `proc_mesh` and `actor`s on those hosts. This gives users fine-grained control over where actors should be placed, for i.e. colocation etc.

Forge is currently `proc_mesh` centric - services are created without any host mechanism applied:
```
s = await spawn_service(ServiceConfig(replicas=2, procs_per_replica=2, with_gpus=True), MyActor, ...)
```

This issue tracks the work needed to support `HostMesh` in Forge services:

```
s = await spawn_service(ServiceConfig(
    replicas=2,
    hosts_per_replica=2,
    procs_per_host=8,
    with_gpus=True),
    ...)
```

See EX545081 for a high level diagram for how services work.

# Technical Requirements
## Resource Management/Scheduling and Service Decoupling
HostMeshes are allocated via `torchx`, which acts as the abstraction layer between application code (i.e. Forge) and the underlying scheduler (i.e. Kubernetes, SLURM, etc.). 

Forge is currently configured so that the underlying [proc_mesh](https://github.com/meta-pytorch/forge/blob/ce91430307547726ff172548cde82d6fe7b4ae31/src/forge/types.py#L94) [talks directly to the scheduler](https://github.com/meta-pytorch/forge/blob/ce91430307547726ff172548cde82d6fe7b4ae31/src/forge/controller/proc_mesh.py#L93), but this might not make sense as there likely needs to be some sort of global coordination of resources.

We should be mindful of complexity - this doesn't need to be as comprehensive or general purpose as Ray is (i.e. I hope we don't write our own binpacking algorithm), and we should feel free to change our service definition to provide the right level of control.

## Replica Colocation
Replica colocation includes 2 pieces of functionality in my mind:
- **GPU-aware scheduling**: assuming I have 2 hosts in my cluster, and I want to spawn 4 inference servers with 4 GPUs each, I expect that my service will use those 2 hosts fully
[...]

PR #151, opened 2025-09-12
Initial multi-host support

allenwang28:
This PR introduces initial multi-host support. For context, see https://github.com/meta-pytorch/forge/issues/144

To summarize, the main limitations we have around Monarch's HostMesh right now are:
1. You can't spawn multiple proc meshes on a HostMesh
2. Actors spawned on different remote meshes cannot communicate with each other

You can, however, facilitate communications from the main/client/controller. This PR creates multiple host meshes as we will normally, but there are a few limitations:
- Only supports SLURM remote scheduling for now
- For worker `os.env`/PTD setup, I saw strange vLLM initialization issues if I tried to set environment variables after the `proc_mesh` was already created. The ideal path:
  - Create a host mesh
  - Get the master addr/port from a host in the mesh
  - Create the proc mesh, setting the env variable
- Can't show a real multi-host example until Titan integration is landed in the example.
- Can't use TorchStore because this requires remote actors that can communicate with each other. We can work around this with DCP save/load to NFS

Key API changes:
1. `ProcessConfig` uses `num_hosts` to determine whether or not to use a remote allocation. `None` means spawn locally, `num_hosts >=1` will run an actual remote allocation
2. `ServiceConfig` correspondingly introduces `hosts_per_replica` to utilize `num_hosts`

Other extras:
- Added in gitignore for some SLURM stuff
- Added in a launcher that runs the same GRPO example, just with multi-host enabled
- Preps vLLM to work with DeepSeek. This won't work because we need 2 hosts, and we can't set up PTD correctly, but we can run it with 1 host and see an OOM (expected)
  - Also added a config for Qwen 32B. It works locally w/ 4 hosts, but is currently set with `hosts_per_replica=1` just to test multi-host
  - Policy currently places the workers on the remote host, but the controller is placed on the client. This is because of the HostMesh limitations

For reviewers, I'd urge to not focus too much on provisioner because that'll be changed once we have proper HostMesh support!

Sample logs: P1944199171

### DIFF
--- a/src/forge/controller/provisioner.py
+++ b/src/forge/controller/provisioner.py
@@ -142,6 +142,7 @@ class Provisioner:
             else:
                 host_mesh = this_host()
                 gpu_manager = self._host_gpu_map[self._this_host_id]
+                host_mesh._host_id = self._this_host_id
 
             if with_gpus:
                 # The ideal path here:


src/forge/actors/policy.py
```diff
@@ -196,7 +215,7 @@ async def setup(self):
 
         self.request_id = 0
         self.requests: Dict[str, tuple[None | ParentRequest, asyncio.Future]] = {}
-        self.vllm_args = await self.policy_worker.get_vllm_args.choose()
+        self.vllm_args = EngineConfig.as_engine_args(self.engine_config)
```
allenwang28: note, I added `as_engine_args` because when I was trying DeepSeek I saw some weird pickling issues when trying to do the `get_vllm_args.choose()`. I previously spent several hours trying to debug it until I decided it wasn't worth it.

apps/vllm/main.py
```diff
@@ -11,14 +11,20 @@
...
 from forge.actors.policy import Policy
 from forge.cli.config import parse
+from forge.controller.provisioner import shutdown
 from forge.controller.service import ServiceConfig, shutdown_service, spawn_service
 
 from omegaconf import DictConfig
 from src.forge.data.utils import exclude_service
 from vllm.outputs import RequestOutput
 
+os.environ["HYPERACTOR_MESSAGE_DELIVERY_TIMEOUT_SECS"] = "600"
```
LucasLLC, line 25: I think this snuck in from BS that I did and might not be necessary. Unless you found something?

apps/grpo/main.py
```diff
@@ -480,7 +481,11 @@ async def continuous_training():
             shutdown_service(compute_advantages),
             shutdown_service(ref_model),
             shutdown_service(reward_actor),
+            return_exceptions=True,
         )
+        # TODO - add a global shutdown that implicitly shuts down all services
+        # and remote allocations
+        await shutdown()
```
joecummings, line 483: This would be awesome!

apps/vllm/deepseek_r1.yaml
```diff
@@ -0,0 +1,22 @@
...
+    enable_expert_parallel: true
+    # enforce_eager: true
+  sampling_config:
+    n: 2
+    guided_decoding: false
+    max_tokens: 512
+  available_devices: null
+  service:
+    procs_per_replica: 8
+    hosts_per_replica: 2
```
joecummings, line 16: nit: have a consistent ordering here - either procs first or hosts first.

apps/vllm/main.py
```diff
@@ -27,28 +33,29 @@ async def run(cfg: DictConfig):
...
+
+            print("\nGeneration Results:")
+            print("=" * 80)
+            for batch, response in enumerate(response_output.outputs):
+                print(f"Sample {batch + 1}:")
+                print(f"User: {prompt}")
+                print(f"Assistant: {response.text}")
+                print("-" * 80)
+
+            print("\nShutting down...")
```
joecummings: I know this isn't you but can you move it down to the branch where it's actually shutting down
?

src/forge/actors/policy.py
```diff
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from __future__ import annotations
```
joecummings, line 7: ?

src/forge/controller/provisioner.py
```diff
@@ -0,0 +1,236 @@
...
+class GpuManager:
+    """Tracks and assigns GPU devices on a host mesh.
+
+    This currently mimics the `gpu_manager` in system_controllers - we will
+    consolidate as part of the "proper HostMesh integration" work.
+
+    """
+
+    def __init__(self):
+        # TODO - extend this to support multiple HostMeshes too
```
joecummings: ?

src/forge/controller/provisioner.py
```diff
@@ -0,0 +1,236 @@
...
+
+    async def create_host_mesh(self, name: str, num_hosts: int) -> HostMesh:
+        """Creates a remote server and a HostMesh on it."""
+        # no need to lock here because this is already locked behind `get_proc_mesh`
+        logger.debug(f"Creating remote server for alloc {name}")
+        appdef = hyperactor.host_mesh(
+            image="test", meshes=[f"{name}:{num_hosts}:gpu.small"]
+        )
+        for role in appdef.roles:
+            # Note - this is hardcoded to SLURM
```
joecummings, line 95: n00b question: is there a way to query this information from a Python API?

src/forge/types.py
```diff
@@ -91,14 +91,9 @@ class State:
 class ProcessConfig:
     """A proc_mesh config for the torchx scheduler."""
 
-    scheduler: Literal["mast", "local"] = "local"
     num_procs: int = 1
     with_gpus: bool = False
-    num_hosts: int = 1
```
joecummings, line 97: Will this no longer work on mast then?

src/forge/actors/policy.py
allenwang28: ah this is needed for `EngineConfig` so we can do 
```
@classmethod
def as_engine_args(cls, config: Mapping | EngineConfig) ...
```

otherwise it'd need to be like
```
@classmethod
def as_engine_args(cls, config: Mapping | "EngineConfig") ...
```

and Python complains about the latter

src/forge/controller/provisioner.py
allenwang28: old comment, removed

src/forge/controller/provisioner.py
allenwang28: i'm not sure tbh, torchx doesn't have this which makes me think there isn't

src/forge/types.py
allenwang28: yeah i'm gonna build MAST integration in another PR

apps/vllm/main.py
allenwang28: ah, so this app isn't running with torchstore, it's that when I try and run DeepSeekv3 it takes like an hour to download the weights. I added this here so it doesn't crash

apps/grpo/multihost.yaml
```diff
@@ -0,0 +1,81 @@
+# GRPO Training Configuration
```
pbontrager: Maybe name it qwen3_multinode

launcher/job.sbatch
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
pbontrager, line 1: I was under the impression that torchx meant we didn't need this? Either way, I think this should be in the GRPO app and take the config still as a conditional.

src/forge/controller/proc_mesh.py
```diff
@@ -4,137 +4,27 @@
...
-        if env:
-            p = await ProcMesh.from_alloc(alloc, setup=partial(_setup_env, env=env))
-        else:
-            p = await ProcMesh.from_alloc(alloc)
-        await p.logging_option(stream_to_client=True, aggregate_window_sec=3)
-        return p
-    else:
-        raise ValueError("Unsupported scheduler: {}".format(process_config.scheduler))
+    # TODO - remove this
+    return await _get_proc_mesh(process_config)
```
pbontrager, line 24: What's the reason for this? Can't we just point directly to it?

src/forge/controller/provisioner.py
```diff
@@ -0,0 +1,235 @@
...
+        self.available_gpus -= set(gpus)
+        return [str(gpu) for gpu in gpus]
+
+    def release_gpus(self, gpu_ids: list[str]) -> None:
+        """Releases the given GPU devices."""
+        for gpu_id in gpu_ids:
+            self.available_gpus.add(int(gpu_id))
+
+
+class Provisioner:
```
pbontrager, line 71: At this point it would be useful to have a diagram of all the concepts related to services so it'll be easier to maintain.

launcher/job.sbatch
allenwang28: yeah it's convoluted, we run sbatch to schedule the controller, then the controller calls sbatch through torchx.

We need the controller to run on a GPU node so that Monarch's build doesn't complain because it's built with tensor engine. I'm not sure what the right long-term solution is quite yet

src/forge/controller/proc_mesh.py
allenwang28: I promise I'll fix this later lol

apps/vllm/main.py
LucasLLC: From my understanding, this timeout is decoupled from the e2e latency of an endpoint

launcher/job.sbatch
allenwang28: I am going to leave it for now, but will think on this more

Issue #133, opened 2025-09-05
[RFC] Services are all you need?


Jack-Khuu:
_(Based on conversations with @pbontrager and @DNXie)_

Currently, Forge users are exposed to 2 concepts: [Services](https://github.com/meta-pytorch/forge/blob/d4011ea56f36333d191a8b0dd64a21c8e0df9757/src/forge/controller/service/service.py#L53) (spawns processes and replicas) and [ForgeActor](https://github.com/meta-pytorch/forge/blob/4372a5490d748889070fa6374499a274185eb540/src/forge/controller/actor.py#L21) (class doing the "work"). **Should we flatten this mental model so that users only think of everything as a Service instead?**

From a user's perspective, Services are a wrapper around Actor instances, with the routing setup logic intentionally abstracted away from the user. The API for spawning a Service even _literally_ takes the underlying Actor as an argument.

https://github.com/meta-pytorch/forge/blob/4372a5490d748889070fa6374499a274185eb540/src/forge/controller/service/spawn.py#L22-L24

---

### **Option A: Just a rebrand**
- `Service` => `ServiceRouter`
- `ForgeActor` => `Service`
- Nothing is functionally changed; just a rename
- @DNXie has started on something similar in https://github.com/meta-pytorch/forge/pull/124

```python
async def spawn_service( 
     service_cfg: ServiceConfig, actor_def: Type[Service], **actor_kwargs 
 ) -> ServiceInterface: 

# Where ServiceInterface is mentally mapped as the functionality of both the Router and Service
```

### **Option B: Flatten the layers**
Currently, the architecture is layered such that each class has a distinct responsibility: 
- **`ServiceInterface`** wraps `Service` and exposes select meta APIs from it's nested child classes (actor, replica, etc)
- **`Service`** has a `List[Replica]` and deals with routing and load balancing
  -  _Recall that the final Service implementation should be an Actor itself_
- **`Replica`** has `ForgeActor` and manages the request queueing 
- **`ForgeActor`** has the actual endpoints that users write and adds Logging
- **`Actor`** is your base Monarch Actor 

This looks great from a "Single Responsibility Design" aspect, but **should we flatten the overall architecture to reduce this depth**? Effectively moving from "instantiate a new class and adding it as a class member" to "populating a field"

Specifically, combining the interfaces of `Service` and `ForgeAgent`, call it `SuperService` for the sake of discussion

``` python
class SuperService(Actor):
    # All the interfaces in ForgeActor
[...]

PR #153, opened 2025-09-12
Refactor service spawning: add ForgeActor.options().as_service() API

DNXie:
This PR refactors how services are spawned for `ForgeActor` instances, providing a more actor-centric and user-friendly API. It is based on the discussion in https://github.com/meta-pytorch/forge/issues/133

**Before:**

```python
policy = await spawn_service(
    ServiceConfig(num_replicas=1, procs_per_replica=2),
    Policy,
    **cfg.policy,
)
shutdown_service(policy)
```

**After:**

**Option 1**: pass service configs
```python
policy = await Policy.options(num_replicas=1, procs_per_replica=2).as_service(**cfg.policy)
```

**Option 2**: pass ServiceConfig object directly
```python
cfg = ServiceConfig(num_replicas=1, procs_per_replica=2)
policy = await Policy.options(service_config=cfg).as_service(**cfg.policy)
```

**Option 3:** use the default configuration

```python
# by default uses num_replicas=1, procs_per_replica = 1
policy = await Policy.as_service(**cfg.policy)
```

**To shutdown**: 
```python
policy.shutdown()
```

**Changes introduced:**

* Added `ForgeActor.options()` class method, which returns a pre-configured `ForgeActor` subclass.
* Updated `ForgeActor.as_service()` to spawn the service asynchronously and return a `ServiceInterface`.
* Added `ServiceInterface.shutdown()` to allow shutting down the underlying service directly.
* Updated test files `tests/unit_tests/test_service.py` and `tests/integration_tests/test_policy_update.py`.
* Added two new unit tests in `tests/unit_tests/test_service.py` to verify all three construction options (explicit `ServiceConfig`, implicit kwargs, and defaults) work correctly.
* Updated all service usages in code to align with the new API.
* Removed `spawn_service()` and `shutdown_service()`.

**Test:**
```shell
pytest tests/integration_tests/test_policy_update.py
python -m apps.rl.main --config apps/rl/llama3_8b.yaml
python -m apps.vllm.main --config apps/vllm/llama3_8b.yaml
```

### DIFF
--- a/tests/unit_tests/test_service.py
+++ b/tests/unit_tests/test_service.py
@@ -74,7 +74,7 @@ async def test_actor_def_type_validation():
         await InvalidActor.options(procs_per_replica=1, num_replicas=1).as_service()
 
 
-@pytest.mark.timeout(10)
+@pytest.mark.timeout(20)
 @pytest.mark.asyncio
 async def test_service_with_explicit_service_config():
     """Case 1: Provide a ServiceConfig directly."""
@@ -89,7 +89,7 @@ async def test_service_with_explicit_service_config():
         await service.shutdown()
 
 
-@pytest.mark.timeout(10)
+@pytest.mark.timeout(20)
 @pytest.mark.asyncio
 async def test_service_with_kwargs_config():
     """Case 2: Construct ServiceConfig implicitly from kwargs."""
@@ -109,7 +109,7 @@ async def test_service_with_kwargs_config():
         await service.shutdown()
 
 
-@pytest.mark.timeout(10)
+@pytest.mark.timeout(20)
 @pytest.mark.asyncio
 async def test_service_options_missing_args_raises():
     """Case 3: Error if neither service_config nor required args are provided."""
@@ -117,7 +117,7 @@ async def test_service_options_missing_args_raises():
         await Counter.options().as_service()  # no args, should raise before service spawn
 
 
-@pytest.mark.timeout(10)
+@pytest.mark.timeout(20)
 @pytest.mark.asyncio
 async def test_service_default_config():
     """Case 4: Construct with default configuration using as_service directly."""


src/forge/controller/actor.py
```diff
@@ -41,6 +43,79 @@ def __init__(self, *args, **kwargs):
...
+            """
+
+            _actor_def = cls
+            _service_interface: ServiceInterface | None
+
+            def __init__(self) -> None:
+                self._service_interface = None
+
+            @classmethod
+            async def as_service(cls, **actor_kwargs) -> "ConfiguredService":
```
allenwang28: I would add `.as_service(...)` as a `classmethod` on the actor itself

If I did `Policy.as_service()` for instance, I would expect it to spawn with the defaults of `procs_per_replica=1, num_replicas=1` etc. 

Then all `.options(...)` does is change class level attributes

allenwang28: (otherwise to use `.as_service()` I'll always need to do `options` first

DNXie: Reasonable. Thanks for the suggestion. Just updated the code to support that.

DNXie:
@allenwang28 I removed `spawn_service`  and `shutdown_service` as we don't need them anymore. How about [`spawn_service_v2`](https://github.com/meta-pytorch/forge/blob/c50a2a1b6a43f53a097c22558c4c9ec3e1de3246/src/forge/controller/service/spawn.py#L58), [`shutdown_service_v2`](https://github.com/meta-pytorch/forge/blob/c50a2a1b6a43f53a097c22558c4c9ec3e1de3246/src/forge/controller/service/spawn.py#L89)? Actually the question is: do we need [`ServiceActor`](https://github.com/meta-pytorch/forge/blob/c50a2a1b6a43f53a097c22558c4c9ec3e1de3246/src/forge/controller/service/service.py#L609) (as well as [`ServiceInterfaceV2`](https://github.com/meta-pytorch/forge/blob/main/src/forge/controller/service/interface.py#L203))? It is never used. 

And the way how [`ServiceActor` initiates `Service`](https://github.com/meta-pytorch/forge/blob/cfd9677784ee29353f53b6496068ebaa7e6e593c/src/forge/controller/service/service.py#L631), looks buggy to me.

apps/grpo/main.py
```diff
@@ -351,39 +350,20 @@ async def main(cfg: DictConfig):
...
-            DatasetActor,
-            **exclude_service(cfg.dataset),
+        DatasetActor.options(**cfg.dataset.service).as_service(
+            **exclude_service(cfg.dataset)
         ),
-        spawn_service(
-            ServiceConfig(**cfg.policy.service),
-            Policy,
-            **exclude_service(cfg.policy),
+        Policy.options(**cfg.policy.service).as_service(**exclude_service(cfg.policy)),
```
allenwang28, line 360: hmm would it make more sense to have `service` as a standalone config? I really hate that we're doing `**exclude_service(...)` in every line.

It would look like:

```
policy:
  ...
  # (removed the service here)

services:
  policy:
    ...      
```

so then our code can just be

```
Policy.options(**cfg.service.policy).as_service(**cfg.policy)
```

src/forge/controller/actor.py
```diff
@@ -41,6 +43,116 @@ def __init__(self, *args, **kwargs):
...
+                raise ValueError(
+                    "Must provide either `service_config` or (num_replicas + procs_per_replica)."
+                )
+            cfg = ServiceConfig(
+                num_replicas=num_replicas,
+                procs_per_replica=procs_per_replica,
+                **service_kwargs,
+            )
+
+        class ConfiguredService:
```
allenwang28: Can we remove the `ConfiguredService` piece altogether? `options()` should return a `type["ForgeActor"]`

tests/unit_tests/test_service.py
```diff
@@ -61,26 +62,84 @@ async def add_to_value(self, amount: int, multiplier: int = 1) -> int:
 @pytest.mark.timeout(10)
 @pytest.mark.asyncio
 async def test_actor_def_type_validation():
-    """Test that spawn_service validates actor_def is a subclass of ForgeActor."""
+    """Test that .options() rejects classes that are not ForgeActor subclasses."""
 
-    # Service can only spawn ForgeActor subclasses
+    # Only spawn ForgeActor subclasses has options method
```
allenwang28:
```suggestion
    # Only `ForgeActor`s can be spawned as services
```

src/forge/controller/actor.py
DNXie: The main reason we have the `ConfiguredService` wrapper is to allow direct access to actor endpoints via instance attributes (e.g., service.value.choose()) while still carrying the `ServiceConfig` on the class.

If we remove the wrapper and make `options()` return a subclass of `ForgeActor` directly, the endpoints (like [`counter.value`](https://github.com/meta-pytorch/forge/blob/2e97f6767b79ba75f5b02060c384786646dd3b82/tests/unit_tests/test_service.py#L35)) are only accessible through the underlying ServiceInterface. So we need to change the statement
```
await service.value.choose()
```
to 
```
await service._service_interface.value.choose()
```


I attempted to delegate [`EndpointProperty`](https://github.com/meta-pytorch/monarch/blob/dc471be3fe63430de6b39305df071985b65cdf7c/python/monarch/_src/actor/endpoint.py#L211) access via `__getattr__` like this 
```python
def __getattr__(self, item):
    if self._service_interface is None:
        raise AttributeError(f"Service not started yet; cannot access '{item}'")
    
    attr = getattr(self._service_interface, item)
    from monarch._src.actor.endpoint import EndpointProperty
    if isinstance(attr, EndpointProperty):
        # Call the descriptor's __get__ to bind it
        return attr.__get__(self._service_interface, type(self._service_interface))

    return attr
```

However, it didn’t fully work: Python descriptors like `EndpointProperty` only bind correctly when accessed through the class or a properly initialized ServiceInterface instance. The `__get__` call in `__getattr__` does not fully replicate the descriptor binding behavior.

I’m still getting familiar with the internals of Monarch and some of the Python descriptor mechanics, so I'd love to hear any suggestions if there’s a cleaner way to handle this.

apps/grpo/main.py
DNXie: I like it too. Although given the scope of this PR, I feel it is better to open a followup(separate) PR for this change. What do you think?

src/forge/controller/actor.py
allenwang28: hmm, `EndpointProperty` binding is only needed after we do `as_service()` right? 

this is how I imagine the `options` piece:

```
class ForgeActor(Actor):
    _service_config: ServiceConfig | None = None

    @classmethod
    def options(
        cls,
        *,
        service_config: ServiceConfig | None = None,
        num_replicas: int | None = None,
        procs_per_replica: int | None = None,
        **service_kwargs,
    ) -> Type["ForgeActor"]:
        if service_config:
            config = service_config
        else:
            config = ServiceConfig(num_replicas=num_replicas, procs_per_replica=procs_per_replica)  
        return type(
            f"{cls.__name__}",
            (cls,),
            {"_service_config": config}
        )```

**Comment by allenwang28** [comment]
> Actually the question is: do we need [ServiceActor](https://github.com/meta-pytorch/forge/blob/c50a2a1b6a43f53a097c22558c4c9ec3e1de3246/src/forge/controller/service/service.py#L609) (as well as [ServiceInterfaceV2](https://github.com/meta-pytorch/forge/blob/main/src/forge/controller/service/interface.py#L203))? It is never used.

yeah please leave it for now, I'm gonna clean this all up once HostMesh lands 😅

File: src/forge/controller/actor.py (hunk 3 above)
Comment (DNXie) [reply src/forge/controller/actor.py]: > EndpointProperty binding is only needed after we do as_service() right?

Yes. It is needed whenever we call an endpoint function.

Your proposed implementation seems reasonable. However, it does not handle endpoint binding. After `as_service()`, you still need to go through `service._service_interface` to access endpoints.

Comment (DNXie) [reply src/forge/controller/actor.py]: As we discussed offline, if `as_service` returns a `ServiceInterface` directly, we cannot terminate the service with
```
service.shutdown()
```

because the returned object (`service`) is just a `ServiceInterface`. In that case, we’d have to fall back to
```
shutdown_service(service)
```

Personally, I prefer the `service.shutdown()` style since it feels more natural and object-oriented, but I’m okay with either one.

Comment (allenwang28) [reply src/forge/controller/actor.py]: We can add `def shutdown(self)` directly to `ServiceInterface`

Comment (DNXie) [reply src/forge/controller/actor.py]: Done!

File: src/forge/controller/actor.py [hunk 5]
```diff
@@ -41,6 +44,78 @@ def __init__(self, *args, **kwargs):
...
+        )
+
+    @classmethod
+    async def as_service(cls: Type[T], **actor_kwargs) -> "ServiceInterface":
+        """
+        Convenience method to spawn this actor as a Service using default configuration.
+        If `.options()` was called, it will use the bound ServiceConfig;
+        otherwise defaults to 1 replica, 1 proc.
+        """
+        # Lazy import to avoid top-level dependency issues
```
allenwang28, line 104: this is fine, just curious if this is a circular dependency issue?

DNXie: Yes. There would be a circular dependency issue

apps/grpo/main.py
DNXie: A followup PR for this https://github.com/meta-pytorch/forge/pull/165

PR #154, opened 2025-09-12
Pull VllmConfig construction up to Policy

Jack-Khuu:
Move `VllmConfig` construction into `EngineConfig` at the `Policy` level instead of deferring to the `PolicyWorker`. 
This simplifies the PolicyWorker construction logic from:
```python
class PolicyWorker(ForgeActor):
    vllm_args: EngineConfig | Mapping = EngineConfig()
    ...

    + def __post_init__() ...
```
to just
``` python
class PolicyWorker(ForgeActor):
    vllm_config: VllmConfig
```
---

Misc:
- While doing this it also cleans up the ambiguous intermediate `vllm_args` from Policy
- Fixed SamplingConfig bug where `temp` and `top_p` weren't being passed down from yaml cc: @DNXie 

---
```
python -m pytest tests/unit_tests/test_policy_config.py
python -m apps.vllm.main --config apps/vllm/llama3_8b.yaml
```

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -5,25 +5,24 @@
 # LICENSE file in the root directory of this source tree.
 
 import asyncio
-import logging
 import os
 import sys
+import time
 from collections.abc import Mapping
 from copy import copy
 from dataclasses import asdict, dataclass, field, fields
-from typing import Dict, List
 
 import torch
+import torchstore as ts
 from monarch.actor import current_rank, endpoint, ProcMesh
-from torchstore import MultiProcessStore
-from torchstore._state_dict_utils import DELIM
+from torchstore.state_dict_utils import DELIM
 from vllm.config import VllmConfig
 
 from vllm.engine.arg_utils import EngineArgs
 from vllm.entrypoints.utils import _validate_truncation_size
 from vllm.executor.multiproc_worker_utils import set_multiprocessing_worker_envs
 from vllm.lora.request import LoRARequest
-from vllm.outputs import CompletionOutput
+from vllm.outputs import RequestOutput
 from vllm.sampling_params import GuidedDecodingParams, RequestOutputKind, SamplingParams
 from vllm.transformers_utils.tokenizer_group import init_tokenizer_from_configs
 from vllm.usage.usage_lib import UsageContext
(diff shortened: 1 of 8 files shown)


src/forge/actors/policy.py
```diff
@@ -93,13 +96,21 @@ class EngineConfig(EngineArgs):
...
+    _is_v1_supported_oracle = lambda *_: True
+
     @classmethod
     def from_dict(cls, d: Mapping):
         d = dict(d)
         all_fields = [f.name for f in fields(cls)]
         valid_args = {k: v for k, v in d.items() if k in all_fields}
         return cls(**valid_args)
 
+    def create_vllm_config(self) -> VllmConfig:
```
joecummings, line 105: Will this always be static? Or is there ever any need to customize this from the user's perpective? 

Definitely not saying we should do that now, just curious about the pattern here.

src/forge/actors/policy.py
```diff
@@ -196,36 +208,36 @@ async def setup(self):
 
         self.request_id = 0
         self.requests: Dict[str, tuple[None | ParentRequest, asyncio.Future]] = {}
-        self.vllm_args = await self.policy_worker.get_vllm_args.choose()
+        self.vllm_config = await self.policy_worker.get_vllm_config.choose()
```
Jack-Khuu: Will change to 
``` python
        self.vllm_config = self.engine_config.create_vllm_config()
```

Jack-Khuu:
@allenwang28 Ah I missed this change in #151, part of reason for this mini PR was actually your temp work for testing vllm

I considered doing it as a classmethod like you have it at first, but `EngineConfig.as_engine_args(self.engine_config)` felt a lil verbose

src/forge/actors/policy.py
Jack-Khuu: At least for now `EngineConfig` is a easy abstraction for us to interact with the monster signature of vllmConfig, so this is just an easy way to convert it to the form that vllm actually uses

We could make it static/classmethod, but then the args will just be the members of this class

DNXie:
Thanks for spotting and fixing the issue with `SamplingConfig`! The code LGTM

PR #155, opened 2025-09-15
[WIP] Fast UV Build

pbontrager:
With a working vllm nightly build (thanks @huydhn) we can directly build on uv now. This setup should run very quickly and play nice with more setups. This should be kept in conjunction with the install script for now so we have two different options in case one build approach stops working.

This is currently pinned to 9/12 nightlies. This can be updated in the pyproject whenever we need a newer version but I think it'll be a little more stable now if it doesn't update daily on its own.

TODO: need to update CI and some vLLM API calls for nightlies


pbontrager:
This switches us from vllm stable to nightlies so it'll require some vllm updates to land. Let's do this when we want to switch vllm versions.

PR #157, opened 2025-09-15
[1/N] Core Data Models

Ritesh1905:
A split version of this [RFC PR](https://github.com/meta-pytorch/forge/pull/149)

### DIFF
--- a/src/forge/data_models/scored_completion.py
+++ b/src/forge/data_models/scored_completion.py
@@ -1,3 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
 from dataclasses import dataclass
 
 from forge.data_models.completion import Completion


src/forge/data_models/completion.py
```diff
@@ -0,0 +1,31 @@
...
+
+from dataclasses import dataclass
+from typing import Optional
+
+import torch
+from forge.data_models.prompt import Prompt
+
+
+@dataclass
+class Completion:
```
pbontrager, line 15: Currently our Episode is a combination of Completion and ScoreCompletion and Experience. I think keeping them flat makes customization and logging a bit easier but either way can work.

src/forge/data_models/distributed_metric.py
```diff
@@ -0,0 +1,64 @@
...
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+
+import torch
+import torch.distributed as dist
+
+
+@dataclass
+class DistributedMetric(ABC):
```
pbontrager, line 15: I like this concept and it could be combined with our MetricLogger. But if we're going to handle this in a generic way, I don't think we want to deal with process groups, otherwise you have to set them up with a bunch of services that don't need them. Either we should use these abstractions and return metrics to a MetricService/controller to aggregate or handle it without formal abstractions, in a service specific way, as we do now.

src/forge/data_models/prompt.py
```diff
@@ -0,0 +1,66 @@
...
+@dataclass
+class Message:
+    """A single message in a conversation."""
+
+    chunks: Sequence[str]
+    role: Role
+
+
+@dataclass
+class Prompt:
```
pbontrager, line 29: In general we should match OpenAI requests here as the type. This way we can use the same formatting for local and API judge calls

src/forge/data_models/loss.py
```diff
@@ -0,0 +1,22 @@
...
+
+from dataclasses import dataclass
+
+import torch
+from forge.data_models.distributed_metric import Fraction
+from forge.data_models.minibatch import Minibatch
+
+
+@dataclass
+class LossInput:
```
pbontrager, line 15: This is roughly what we're doing now, except it's trainer_logits + target_minibatch which is a subset of the minibatch you routed for the loss.

PR #158, opened 2025-09-15
[2/N] Core trainer abstraction

Ritesh1905:
A split version of [this RFC](https://github.com/meta-pytorch/forge/pull/149)

### DIFF
--- a/src/forge/data_models/api.py
+++ b/src/forge/data_models/api.py
@@ -1,0 +1,137 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from abc import ABC, abstractmethod
+from typing import Any, Dict, Tuple
+
+from forge.data_models.loss import LossOutput
+from forge.data_models.minibatch import Minibatch
+
+
+# TODO: This file needs should NOT be in the data_models folder/package
+
+
+class Store(ABC):
+    """
+    Abstract base class for a generic key-value store.
+
+    This class defines the interface for a storage backend that can save and retrieve
+    values using string keys. Subclasses should implement the actual storage logic,
+    which could be in-memory, on disk, remote (e.g., RDMA, Redis), or any other backend.
+
+    Example use cases include storing model weights, configuration objects, or any
+    other data that needs to be accessed by key.
+
+    Methods:
+    put(key: str, value: Any) -> None
+        Store a value under the specified key.
+
+    get(key: str) -> Any
+        Retrieve the value associated with the specified key.
+
+    Subclasses must implement both methods.
+    """
+
+    @abstractmethod
+    def put(self, key: str, value: Any) -> None:
+        """Store a value under a key."""
+        pass
+
+    @abstractmethod
+    def get(self, key: str) -> Any:
+        """Retrieve a value by key."""
+        pass
+
+
+class WeightsBuffer:
+    """
+    Concrete class for managing model weights using a generic key-value Store backend.
+    This class provides a simple interface to store and retrieve model weights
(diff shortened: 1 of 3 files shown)


Ritesh1905:
> same as https://github.com/meta-pytorch/forge/pull/159/files ?

Yes, I thought it would create a stacked diff. But seems like this is cumulative. :( 

let me know if there is an easy way to fix this.

src/forge/data_models/api.py
```diff
@@ -0,0 +1,137 @@
...
+
+        Args:
+            minibatch (Minibatch): The minibatch of data to use for gradient computation.
+        Returns:
+            LossOutput: The computed loss and any additional outputs needed for logging or analysis.
+        """
+        pass
+
+    @abstractmethod
+    def apply_gradients(self) -> None:
```
pbontrager, line 118: In general this looks fine, my main concern is whether this would be compatible with Compile and Pipeline parallel APIs? @H-Huang

src/forge/data_models/api.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Apply accumulated gradients to the model parameters.
+        This method should update the model's parameters using the gradients that have
+        been accumulated so far (e.g., by calling an optimizer step). After this call,
+        the accumulated gradients should be cleared/reset.
+        """
+        pass
+
+    @abstractmethod
+    def snapshot_weights(self) -> WeightsBuffer:
```
pbontrager, line 128: This would likely push weights to store for checkpoint handling and weight sync to take over.

src/forge/data_models/api.py
```diff
@@ -0,0 +1,137 @@
...
+from typing import Any, Dict, Tuple
+
+from forge.data_models.loss import LossOutput
+from forge.data_models.minibatch import Minibatch
+
+
+# TODO: This file needs should NOT be in the data_models folder/package
+
+
+class Store(ABC):
```
pbontrager, line 17: I left my comment in 3/N but to repeat here, is it valuable to abstract the buffer too? It's as core to the library as Monarch.

src/forge/data_models/api.py
```diff
@@ -0,0 +1,137 @@
...
+        """Store a value under a key."""
+        pass
+
+    @abstractmethod
+    def get(self, key: str) -> Any:
+        """Retrieve a value by key."""
+        pass
+
+
+class WeightsBuffer:
```
pbontrager, line 49: I don't follow the reason to have this extra layer? Also a buffer is what holds some individual data, vs this would be the entire store?

src/forge/data_models/api.py
pbontrager: Similar to update_weights in the policy, this will be somewhat dependent on the internal state of apply_gradients where you want to call it right after apply_gradients is done (without awaiting it) and then not call apply_gradients again until it has completed. Not as complex as the policy side, but something to keep in mind.

src/forge/data_models/api.py
Ritesh1905: >  It's as core to the library as Monarch.

buffer is just a wrapper on top of store, hence I did not do that. can you elaborate on your reasoning for abstracting buffer?

[EDIT]: Don't have an opinion but does not hurt to abstract the buffer too.

src/forge/data_models/api.py
Ritesh1905: At a high-level...

`Store` is a generic key-value storage abstraction. It can store any kind of data (strings, tensors, configs, etc.), not just model weights.

`WeightsBuffer` is a specialized abstraction focused on the logic and conventions for storing and retrieving model weights. It may add domain-specific features, validation, serialization, or metadata handling that are unique to weights.

src/forge/data_models/api.py
Ritesh1905: I don't have a concern if we want to expose another `Step` API for trainer.

PR #159, opened 2025-09-15
[3/N] Core generator abstraction

Ritesh1905:
A split version of [this RFC PR](https://github.com/meta-pytorch/forge/pull/149)

### DIFF
--- a/src/forge/data_models/api.py
+++ b/src/forge/data_models/api.py
@@ -7,8 +7,11 @@
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Tuple
 
+from forge.data_models.completion import Completion
+
 from forge.data_models.loss import LossOutput
 from forge.data_models.minibatch import Minibatch
+from forge.data_models.prompt import Prompt
 
 
 # TODO: This file needs should NOT be in the data_models folder/package
@@ -133,5 +136,40 @@ class Trainer(ABC):
         The returned buffer can be used to transfer weights between components or for checkpointing.
         Returns:
             WeightsBuffer: A handle or reference to the stored weights buffer.
+        """
+        pass
+
+
+class Generator(ABC):
+    """
+    Abstract base class for a model generator in RL or sequence modeling workflows.
+    This class defines the interface for any generator implementation, which is responsible
+    for producing completions (e.g., text, actions) given a prompt, and for updating its
+    internal model weights. Subclasses should implement the actual logic for generation
+    and weight updates, which may vary depending on the underlying model or framework.
+    """
+
+    @abstractmethod
+    def generate(self, prompt: Prompt) -> list[Completion]:
+        """
+        Generate completions given a prompt.
+        This method should use the current model to produce one or more completions
+        (e.g., text outputs, actions) based on the provided prompt.
+        Args:
+            prompt (Prompt): The input prompt or context for generation.
+        Returns:
+            list[Completion]: A list of generated completions corresponding to the prompt.
+        """
+        pass
+
+    @abstractmethod
+    def update_weights(self, weights_handle: WeightsBuffer) -> None:
+        """
+        Update the weights of the model using the provided weights buffer.
+        This method should update the generator's internal model parameters using
+        the weights stored in the given WeightsBuffer (which may be local or remote).
+        Args:
+            weights_handle (WeightsBuffer): A handle or reference to the weights buffer
+                                            containing the new model weights.
         """
         pass
--- a/src/forge/generators/vllm_generator.py
+++ b/src/forge/generators/vllm_generator.py
@@ -1,0 +1,31 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import List
+
+from forge.data_models.api import Generator, WeightsBuffer
+
+from forge.data_models.completion import Completion
+from forge.data_models.prompt import Prompt
+
+
+class VLLMGenerator(Generator):
+    def __init__(self, model_path: str):
+        self.model_path = model_path
+
+    def generate(self, prompt: Prompt) -> List[Completion]:
+        """
+        Generate completions for a given prompt using vLLM.
+        """
+        return []
+
+    def update_weights(self, weights_handle: WeightsBuffer) -> None:
+        """
+        Update the weights of the model.
+        Args:
+            weights: A remote handle to weights buffer
+        """
+        pass
--- a/src/forge/generators/__init__.py
+++ b/src/forge/generators/__init__.py
@@ -1,0 +1,5 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.


src/forge/trainers/huggingface_trainer.py
```diff
@@ -0,0 +1,54 @@
...
+from forge.data_models.api import Trainer, WeightsBuffer
+
+from forge.data_models.distributed_metric import Fraction, SumDistributedMetric
+
+from forge.data_models.loss import LossOutput
+from forge.data_models.minibatch import Minibatch
+from forge.stores.in_memory_store import InMemoryStore
+
+
+class HuggingFaceTrainer(Trainer):
```
felipemello1, line 18: At first glance, I am a bit skeptical that having multiple Trainers would work in practice. Is the idea that there would be a HF and a Titan trainer? Could you share a bit about why we need the Trainer abstraction?

Ritesh1905: Yes providing an optionality for the end-users. if they want to compare two runs or set bench marks or just want to try them out.

The idea is to define the abstraction and enable the policy and learner actors to inject an implementation of the trainer. Essentially what trainer/generator to use can be provided from config.

felipemello1: Wouldn't they have significantly different config signatures / distributed APIs / library requirements, etc? Do you see it working for a V0 of Forge?

src/forge/data_models/completion.py
```diff
@@ -0,0 +1,31 @@
...
+from dataclasses import dataclass
+from typing import Optional
+
+import torch
+from forge.data_models.prompt import Prompt
+
+
+@dataclass
+class Completion:
+    """A model-generated completion for a given prompt."""
```
felipemello1, line 16: i think we already have a dataclass for this. Not sure if its "Episode". But it makes sense to have a completion class.

src/forge/data_models/distributed_metric.py
```diff
@@ -0,0 +1,64 @@
...
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+
+import torch
+import torch.distributed as dist
+
+
+@dataclass
+class DistributedMetric(ABC):
```
felipemello1, line 15: I think we should use https://github.com/meta-pytorch/forge/tree/main/src/forge/data/dataset_metrics

But lets connect and see if you think it makes sense!

src/forge/data_models/experience.py
```diff
@@ -0,0 +1,68 @@
...
+
+from dataclasses import dataclass
+from typing import Optional, Sequence
+
+import torch
+from forge.data_models.scored_completion import ScoredCompletion
+
+
+@dataclass
+class Experience:
```
felipemello1, line 15: This is very similar to Completion. I wonder if we should just keep one. But its also a bit similar to "Episode".

felipemello1: Need to see how this plays with data packing. It could be that the Packer takes list[Experience]. But i dont think it makes sense for this class to have concat logic if we are having packing.

src/forge/trainers/huggingface_trainer.py
Ritesh1905: Yes, they would sure have different configs and we need have a config abstraction too.

> Do you see it working for a V0 of Forge?

the current apps(grpo) use HF, and I just saw another PR to make the grpo app running for titan. so we already have two variations of it.

src/forge/data_models/loss.py
```diff
@@ -0,0 +1,22 @@
...
+
+@dataclass
+class LossInput:
+    minibatch: Minibatch
+    trainer_logits: torch.Tensor
+
+
+@dataclass
+class LossOutput:
+    loss: Fraction
```
felipemello1, line 22: its interesting to have a LossInput and LossOutput class. I am just afraid that these abstractions would add too much hierarchy. i.e. instead of `loss(logits, targets, mask)`, we have `loss(LossInput(MiniBatch(something_else)))`.

But the LossOutput makes more sense to me because we may be outputting multiple numbers. We should double check, because if its only a couple, then having a dataclass might be an overkill

src/forge/trainers/huggingface_trainer.py
felipemello1: @allenwang28 , do you know if we will support both Titan and HF after we work on Titan integration? At least in torchtune finding all the bugs/optimizations/features for a single framework was a lot of work.

src/forge/data_models/experience.py
Ritesh1905: This is modeled in a way that, Completion is what we receive from the generator and the experience is what we feed into trainer (in-between it goes through scoring, post-processing, etc.).

src/forge/data_models/minibatch.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def from_experiences(
+    exps: Sequence[Experience], max_seq_len: int, pad_val: int = 0
+) -> Minibatch:
+    """
+    Convert a list of experiences to a minibatch.
+    """
+
+    def pack_sequence(
```
felipemello1, line 49: i think we will have to revisit it after we do dataset packing. My idea is to have PackedDataset(iterator), where the iterator is a replay buffer. The replay buffer outputs an Experience, and outputs a PackedMiniBatch (or something like that)

src/forge/data_models/prompt.py
```diff
@@ -0,0 +1,66 @@
...
+    USER = "user"
+    ASSISTANT = "assistant"
+    NONE = "none"
+
+
+@dataclass
+class Message:
+    """A single message in a conversation."""
+
+    chunks: Sequence[str]
```
felipemello1, line 24: whats a chunk? Is it for example 2 consecutive messages of an user?

src/forge/data_models/prompt.py
felipemello1, line 1: At least in torchtune, we needed different types of messages/prompt handling. This makes sense to me for a V0, but i wonder if users will want to do more with this: https://github.com/pytorch/torchtune/blob/main/torchtune/data/_messages.py

Maybe @joecummings has some thoughts?

Ritesh1905: You are right. I thought I left a TODO that we need support other formats like image (bytes) etc.

src/forge/data_models/prompt.py
Ritesh1905: I think the inital ideas was mark each chunk as trainable (mask). I probably complicated this by. we can just use str for now.

src/forge/data_models/api.py
```diff
@@ -0,0 +1,175 @@
...
+
+from forge.data_models.loss import LossOutput
+from forge.data_models.minibatch import Minibatch
+from forge.data_models.prompt import Prompt
+
+
+# TODO: This file needs should NOT be in the data_models folder/package
+
+
+class Store(ABC):
```
LucasLLC, line 20: This looks really close to the KVStore we're building in `https://github.com/meta-pytorch/forge/pull/147`

src/forge/data_models/api.py
```diff
@@ -0,0 +1,175 @@
...
+        """Store a value under a key."""
+        pass
+
+    @abstractmethod
+    def get(self, key: str) -> Any:
+        """Retrieve a value by key."""
+        pass
+
+
+class WeightsBuffer:
```
LucasLLC, line 52: I think this is a reasonable abstraction in forge, although I do wonder if this is something torchstore should support to begin with? 
e.g. `(e.g., in-memory, RDMA, file system, torchstore etc.)` 

These are all things torchstore should support OOB.

src/forge/data_models/api.py
```diff
@@ -0,0 +1,175 @@
...
+        Retrieve the weights stored under the specified key.
+        Args:
+            key (str): The key for which to retrieve the weights.
+        Returns:
+            The weights object or reference associated with the key.
+        """
+        return self.store.get(key)
+
+
+class Trainer(ABC):
```
LucasLLC, line 96: +1

src/forge/stores/in_memory_store.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from forge.data_models.api import Store
+
+
+class InMemoryStore(Store):
```
LucasLLC, line 10: This is the second PR I see an in memory store implementation, and a bit concerned that this boiler plate code is being generated as a result of torchstore! 

Do you think it'd be a good idea to add a 'single process in memory' version of this to torchstore?

src/forge/data_models/api.py
Ritesh1905: ack. Will rebase and remove this once the other PR gets merged.

src/forge/trainers/huggingface_trainer.py
LucasLLC: Having a trainer abstraction feels like a huge benefit -- users may have lots of different reasons for trying different trainers and they should occupy similar shapes

src/forge/generators/vllm_generator.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, model_path: str):
+        self.model_path = model_path
+
+    def generate(self, prompt: Prompt) -> List[Completion]:
+        """
+        Generate completions for a given prompt using vLLM.
+        """
+        return []
+
+    def update_weights(self, weights_handle: WeightsBuffer) -> None:
```
pbontrager, line 25: Currently we have update_weights in Forge as well but I don't think it's compatible with efficient weight updates when the policy is using a continuous batching system. Essentially vLLM decodes a fixed number of tokens per step, it then removes completed requests from the stack and adds new request to fill the stack back up. So the batch is never done. This means that you either have to exhaust the request queue before calling update, which isn't very efficient but gives you full control, or you have to update mid decoding.

Ideally you want to be able to either update as soon as you can (e.g. after the weights have been pre-fetched) or you want want to call generate with a policy request (policy.generate(prompt, policy_version) and then update the policy as soon as all the old policy requests are gone. You would also want to do this at different times for every replica in the service as their queue's would run out at different times.

All of this to say, this logic would be much easier to internalize within the generator. If we put it in the controller, we'll have to have a separate loop that reads a bunch of state information about the generator and calls this update.

src/forge/generators/vllm_generator.py
```diff
@@ -0,0 +1,31 @@
...
+
+from typing import List
+
+from forge.data_models.api import Generator, WeightsBuffer
+
+from forge.data_models.completion import Completion
+from forge.data_models.prompt import Prompt
+
+
+class VLLMGenerator(Generator):
```
pbontrager, line 15: Shouldn't this RFC just define the interfaces? So instead of VLLMGenerator it would just be a Generator abstract class.

src/forge/stores/in_memory_store.py
pbontrager, line 10: I'm not sure this one is worth adding an abstraction for as torchstore is as fundamental to our stack as monarch. We'd essentially just be updating the interface to match store's interface.

src/forge/generators/vllm_generator.py
pbontrager: nit: the input parameter would likely be the policy version or key in store

PR #161, opened 2025-09-15
GRPO Titan RL Trainer

joecummings:
## What does this PR do?

1. This PR integrates torchtitan as the backend for our trainer actor and our reference model actor. Crucially, this gives us the ability to run w/ multiple model parallelisms, which is needed for multi-node setup. 
2. This PR also incorporates some syntactic sugar in the form of `hf://`, which allows users to specify a model from the Hugging Face hub and it will automatically either download it if it does not exist locally or find it in the catch and point the trainer / reference model to that directory.

### How was this PR tested?

This PR was tested for "runnability" using the following combinations of parallelisms on a single node. 

Trainer | Reference | Policy | Tested
-- | -- | -- | --
single | single | single | ✅
single | single | TP 2 | ✅
single | single | TP 4 | ✅
DP 2 | DP 2 | single | ✅
DP 2 | DP 2 | TP 2 | ✅
TP 2 | TP 2 | TP 2 | ✅

In addition, I incorporated a unit test for the config `hf://` specification. 

### FAQs

1. **Why is the titan trainer and reference model *slower* than Hugging Face trainer and reference model?** Presumably, this is b/c we have to a) run in fp32 for now (see #166 for updates on changing this) and b) b/c we now have to convert first back to the Hugging Face format before pushing weights.
2. **Why does loss parallel not work?** Loss parallel is only guaranteed to work *OOTB* with PyTorch's cross entropy loss b/c PyTorch distributed does some automatic sharding logic for all underlying aten ops [here](https://github.com/pytorch/pytorch/blob/v2.8.0/torch/distributed/tensor/parallel/loss.py). Since we write our own GRPO loss, we don't get this for free. A potential follow-up would be to look into enabling this through our own resharding logic in Forge or upstreaming a more general fix to PyTorch for the underlying aten ops we need.
3. **How does the user know what the various knobs from Titan they can play with?** Currently, there are no docs available for ForgeEnginer, ForgeTrainer, etc. This is a huge risk from the UX side of using Forge b/c the user will have to navigate themselves through the torchtitan codebase to figure this out. cc @mjtrm
4. **Why do we push compute logprobs onto the controller GPU? ** I agree this is not ideal. In an effort to keep the ReferenceModel idempotent, I changed it so that it just returns logits. In addition, this lets us reuse the compute_logprobs found in grpo/main.py. However, this is incredible slow and has a lot more computation on rank0 than I would prefer. cc @Jack-Khuu  we should perhaps change this s.t. we create a ReferenceLogprobs Actor that does the logits and logprobs calculation. Less idempotent, but more efficient.

### Is this blocked by anything?

YES: https://github.com/meta-pytorch/torchstore/pull/32 cc @LucasLLC

### DIFF
--- a/src/forge/actors/reference_model.py
+++ b/src/forge/actors/reference_model.py
@@ -95,7 +95,8 @@ class ReferenceModel(ForgeActor):
             # (jackkhuu) Not sure if either context are needed for inference here
             with self.engine.train_context(optional_context_parallel_ctx):
                 with self.engine.maybe_enable_amp:
-                    logits = model_parts[0](input_ids)
+                    with torch.inference_mode():
+                        logits = model_parts[0](input_ids)
         if isinstance(logits, DTensor):
             logits = logits.full_tensor()
         return logits


pradeepfn:
awesome results!.

@joecummings  would you be able to copy/paste the configs we have to change for DP/TP. thanks.

vidhyav:
Curious, what data did you test this with?

joecummings:
> Curious, what data did you test this with?

Everything is with GSM8K dataset

apps/grpo/main.py
```diff
@@ -117,82 +97,64 @@ def new_group(
...
-            dtype=torch.bfloat16,
-            trust_remote_code=True,
-        ).to(self.device)
-        self.model.train()
-
-        self.optimizer = torch.optim.AdamW(
-            self.model.parameters(), lr=self.learning_rate
-        )
-        self.optimizer.zero_grad()
+def collate(batches: list[list[Episode]]):
```
allenwang28, line 100: not for this PR, I would prefer this conversion to be a classmethod on Episode cc @Jack-Khuu

src/forge/actors/trainer.py
```diff
@@ -32,16 +30,12 @@
...
 )
-from torchtitan.distributed import utils as dist_utils
 from torchtitan.experiments.forge.engine import ForgeEngine
 from torchtitan.experiments.forge.job_config import ForgeJobConfig
 
 from forge.controller import ForgeActor
 from forge.data.utils import batch_to_device
 
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
```
allenwang28, line 43: keep this, the logging will work if you `super().__init__()` in `__post_init__()`

src/forge/actors/replay_buffer.py
```diff
@@ -36,7 +36,9 @@ async def add(self, episode: "Episode") -> None:
         self.buffer.append(episode)
 
     @endpoint
-    async def sample(self, curr_policy_version: int, batch_size: int | None = None):
+    async def sample(
+        self, curr_policy_version: int, batch_size: int | None = None
+    ) -> tuple[tuple[Any, ...], ...] | None:
```
allenwang28, line 41: not for this PR,  but it seems really fragile for sample to be returning something defined by a function in the app, with no standard interface. 

cc @Jack-Khuu - can we keep a note of this?

src/forge/actors/trainer.py
joecummings: I'm confused tho - why do we need this if there's already a logger defined on the Actor?

PR #164, opened 2025-09-16
add hf_transfer (faster model download from hf)

felipemello1:
make forge download go brr

more info: https://github.com/pytorch/torchtune/pull/2046

llama 8b: 45s -> 15s


joecummings:
The recommended future-looking solution is XET: https://huggingface.co/docs/huggingface_hub/en/guides/download#faster-downloads

However, we've run into a lot of problems trying to get it working consistently.

PR #167, opened 2025-09-17
Adds basic LLMJudges

Jack-Khuu:
Adds a basic LLMJudge (vllm generation from `Policy`) based on the methods listed in https://hazyresearch.stanford.edu/blog/2025-06-18-weaver

Note: Policy should probably be renamed, but this isn't the PR to do it and there's separate abstraction discussions in https://github.com/meta-pytorch/forge/pull/149

### DIFF
--- a/src/forge/data/judge.py
+++ b/src/forge/data/judge.py
@@ -7,7 +7,11 @@
 from dataclasses import dataclass
 from enum import Enum
 
-from vllm.outputs import RequestOutput
+try:
+    from vllm.outputs import RequestOutput
+except ImportError as e:
+    print(f"Failed to import RequestOutput from vllm.outputs: {e}")
+    RequestOutput = "RequestOutput"
 
 from forge.controller.service.interface import ServiceInterface
 


src/forge/data/judge.py
```diff
@@ -0,0 +1,87 @@
...
+                matching += 1
+            print(output.text)
+
+        return matching > (len(outputs.outputs) // 2)
+
+    async def _first_sample(self, response: str, outputs: RequestOutput) -> bool:
+        """
+        Returns whether there is a match to the first output
+        """
+        first_output = outputs.outputs[0]
```
Ritesh1905, line 70: nit: may be handle the edge case when the outputs are empty?

src/forge/data/judge.py
```diff
@@ -0,0 +1,87 @@
...
+class EvaluationMethodology(str, Enum):
+    """Evaluation methodology for LLM Judge."""
+
+    MAJORITY = "Majority"
+    FIRST_SAMPLE = "First"
+    PASS_N = "Pass N"
+
+
+@dataclass
+class LLMJudge:
```
allenwang28, line 28: if LLMJudge here is evaluating responses i.e. Pass@1, Majority, First Sample, etc. - aren't these metrics rather than judges?

src/forge/data/judge.py
```diff
@@ -0,0 +1,87 @@
...
+    """Simple interface for Judges utilizing LLMs."""
+
+    judge_model: ServiceInterface
```
allenwang28, line 31: instead of this class holding a ServiceInterface, can we instead pass the generated responses directly to evaluate()?

src/forge/data/judge.py
Jack-Khuu: Aren't metrics just for logging? 

The Judges can be used as part of the generation evaluation (analogous to Rewards)

src/forge/data/judge.py
Jack-Khuu: So have the main loop directly manage calling all N weak verifiers in the rollout?

I don't have a strong preference here, but doing so does increases the boilerplate/management load

src/forge/data/judge.py
allenwang28: yeah Metric is sort of a loaded term unfortunately. Maybe it's `EvaluationMetric`?

src/forge/data/judge.py
allenwang28: hmm, so this is my mental model of how Weaver works:

1. Generator generates K responses
2. K responses go through N verifiers, producing KN verifier results
3. KN results gets distilled down to a scalar 0/1 through weaver

Meanwhile, for pass@1, majority, first sample, and pass @K, they're more like "assuming we know the answer already, was the generator able to produce the correct results in K tries?"

so when it comes to this PR, it depends on what we're trying to accomplish - is it step 2?

Jack-Khuu: Step 2 (will update to return a list/tensor of length K) where N-judges correspond to N verifiers

Good catch though, I should generalize this to K responses

allenwang28: ok, so if we're shooting for 2 then I wouldn't necessarily include pass@1, majority, first sample, pass@K etc. which are separate from the verifiers

What we should show is like N different models from [RewardBench](https://huggingface.co/spaces/allenai/reward-bench) as individual services

if we want to introduce a Judge concept, then I think we should do two things:
- Rename Policy* to Inference*
- Make the Judge a special instance of Policy that uses `generate` to turn the final result into scalars or w/e is needed

Jack-Khuu:
Will remake PR, since first N commits were going in a different direction than needed

PR #169, opened 2025-09-17
Update monarch wheel to include process_allocator

allenwang28:
This includes the process_allocator binding in the Monarch install - previously you had to build this yourself.


JenniferWang:
Does this mean -- for the change to take effect, run `./scripts/build_wheels.sh`?
How is this different from say, `pip install --upgrade monarch`? 
I also noticed that my `pip list` shows `monarch` not `torchmonarch-nightly` (as specified in the `pyproject.toml`)

allenwang28:
@JenniferWang you really only need to run
```
pip uninstall -y monarch
pip install assets/wheels/monarch-*
```
from your conda env

You're not seeing the `torchmonarch-nightly` because the way we build wheels now shows it as `monarch` instead (which I agree is confusing)

vidhyav:
Wait can I just run the rebuild script?

allenwang28:
> Wait can I just run the rebuild script?

yes, although transparently it's configured to just work on my dev GPU so your mileage may vary lol

PR #172, opened 2025-09-17
Refactor Service routing into separate router classes

DNXie:
This PR refactors the Service routing logic into separate router classes.

**Changes:**

- Moved the routing logic into independent router class: `RoundRobinRouter`, `LeastLoadedRouter`, and `SessionRouter`. 
- Added a `Router` interface.
- Refactored `Service` class accordingly.
- Added unit tests for the added router classes and integration tests for `SessionRouter` and `LeastLoadedRouter` behavior and fallback routing.

**Test:**
```
```

### DIFF
--- a/tests/unit_tests/test_service.py
+++ b/tests/unit_tests/test_service.py
@@ -660,36 +660,6 @@ async def test_broadcast_call_vs_choose():
 
 
 @pytest.mark.asyncio
-async def test_least_loaded_router_basic():
-    """LeastLoadedRouter picks the replica with lowest load."""
-    replicas = [
-        make_replica(0, load=5),
-        make_replica(1, load=1),
-        make_replica(2, load=3),
-    ]
-    router = LeastLoadedRouter()
-    chosen = router.get_replica(replicas)
-    assert chosen.idx == 1  # lowest load
-
-
-@pytest.mark.asyncio
-async def test_session_router_assigns_and_updates_session_map():
-    """SessionRouter updates session_map and preserves sticky sessions."""
-    replicas = [make_replica(0), make_replica(1)]
-    session_map = {}
-    fallback = LeastLoadedRouter()
-    router = SessionRouter(fallback)
-
-    # First request assigns via fallback
-    r1 = router.get_replica(replicas, sess_id="sess1", session_map=session_map)
-    assert session_map["sess1"] == r1.idx
-
-    # Second request should stick
-    r2 = router.get_replica(replicas, sess_id="sess1", session_map=session_map)
-    assert r1.idx == r2.idx
-
-
-@pytest.mark.asyncio
 async def test_session_router_with_round_robin_fallback():
     """Switch fallback router to round-robin and verify assignment order."""
     # Choose RoundRobinRouter as fallback, r1 and r2 should be assigned to different replicas


src/forge/controller/service/router.py
```diff
@@ -0,0 +1,90 @@
...
+        self._next_idx = 0
+
+    def get_replica(
+        self,
+        replicas: List[Replica],
+        sess_id: str | None = None,
+        session_map: Dict[str, int] | None = None,
+    ) -> Replica:
+        healthy_replicas = [r for r in replicas if r.healthy]
+        if not healthy_replicas:
```
allenwang28, line 29: we should just pass the healthy_replicas into the router

src/forge/controller/service/router.py
```diff
@@ -0,0 +1,90 @@
...
+
+        return replica
+
+
+class LeastLoadedRouter(Router):
+    """Always routes to the replica with the lowest current load."""
+
+    def get_replica(
+        self,
+        replicas: List["Replica"],
```
allenwang28: can this not be `List[Replica]`?

src/forge/controller/service/router.py
```diff
@@ -0,0 +1,90 @@
...
+
+class SessionRouter(Router):
+    """Session-based routing: sticky sessions with a fallback router."""
+
+    def __init__(self, fallback_router: Router):
+        self.fallback_router = fallback_router
+
+    def get_replica(
+        self,
+        replicas: List["Replica"],
```
allenwang28: Same `"Replica"` comment

src/forge/controller/service/router.py
```diff
@@ -0,0 +1,90 @@
...
+        sess_id: str | None = None,
```
allenwang28, line 61: `sess_id: str,`

i.e. don't assume `None` is passable for this or `session_map`. I would also get rid of the checks below

tests/unit_tests/test_service.py
```diff
@@ -634,3 +654,190 @@ async def test_broadcast_call_vs_choose():
 
     finally:
         await service.shutdown()
+
+
+# Rounter Tests
```
allenwang28: nit: typo
```suggestion
# Router Tests
```

tests/unit_tests/test_service.py
```diff
@@ -634,3 +654,190 @@ async def test_broadcast_call_vs_choose():
...
+
+    assert r1.idx == r2.idx == 0
+
+
+# Router integeration tests
+
+
+@pytest.mark.timeout(10)
+@pytest.mark.asyncio
+async def test_round_robin_router_distribution():
```
allenwang28, line 694: for these router integration tests, are these not already covered in the above tests?

src/forge/controller/service/router.py
DNXie: This is because I have a interface for routers: (also see `interface.py`)
```
class Router(ABC):
    """Abstract base class for routing logic."""

    @abstractmethod
    def get_replica(
        self,
        healthy_replicas: List[Replica],
        sess_id: str | None = None,
        session_map: Dict[str, int] | None = None,
    ) -> Replica:
        """Select a replica from the list based on routing logic."""
        pass

```
For RRRouter and LeastLoadedRouter, this could be None.

tests/unit_tests/test_service.py
DNXie: Most functionalities are covered. But the above tests are only testing router itself. The integration tests are testing routing behaviors through a service. If you have concerns about the CI overhead, we could get ride of these except `test_round_robin_router_distribution`, which is the only test for roundrobin.

src/forge/controller/service/router.py
DNXie: Done.

src/forge/controller/service/router.py
```diff
@@ -0,0 +1,90 @@
...
+        if not healthy_replicas:
+            raise RuntimeError("No healthy replicas available for load balancing")
+
+        self._next_idx = (self._next_idx + 1) % len(healthy_replicas)
+        replica = healthy_replicas[self._next_idx]
+
+        return replica
+
+
+class LeastLoadedRouter(Router):
```
joecummings, line 38: nit: BalancedRouter?

tests/unit_tests/test_service.py
DNXie: I removed one integration tests and 2 other unit tests due to overlapped coverage, so now just one unit test and two integration tests:
- `test_session_router_with_round_robin_fallback`: test different fallback router for the SessionRouter. Also tests the correctness of `LeastLoadedRouter`. 
- `test_round_robin_router_distribution` (integration): the only test case for RR logic
- `test_session_router_assigns_and_updates_session_map_in_service` (integration):  is to test whether the session_map modified in Router can get updated properly in Service class. 

So I think we should keep both.

src/forge/controller/service/interface.py
```diff
@@ -277,3 +280,17 @@ def __getattr__(self, name: str):
         raise AttributeError(
             f"'{self.__class__.__name__}' object has no attribute '{name}'"
         )
+
+
+class Router(ABC):
```
allenwang28, line 285: nit - I think this Router can actually just be in router.py

src/forge/controller/service/router.py
allenwang28: I think `LeastConnectedRouter` would be the most canonically accurate

src/forge/controller/service/router.py
allenwang28: hmm ok, I'm not sure how I feel about that longer term but that's ok for now

tests/unit_tests/test_service.py
allenwang28: sgtm

src/forge/controller/service/interface.py
DNXie: I think we cam keep it in interface.py since that's where all the interfaces are.

src/forge/controller/service/router.py
DNXie: Thanks for the suggestion! 

See Allen's discussion on this in another thread. 

> I think `LeastConnectedRouter` would be the most canonically accurate

I agree. So let's keep it as `LeastConnectedRouter` for now.

PR #176, opened 2025-09-18
DCP for weight sync

LucasLLC:
Switches to DCP for weight sync, adds controller workaround to TS

image

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -15,10 +15,10 @@ from copy import copy
 from dataclasses import asdict, dataclass, field, fields
 
 import torch
+import torch.distributed.checkpoint as dcp
 import torchstore as ts
 from monarch.actor import current_rank, endpoint, ProcMesh
 from torchstore.state_dict_utils import DELIM
-import torch.distributed.checkpoint as dcp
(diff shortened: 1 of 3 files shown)


src/forge/actors/policy.py
```diff
@@ -416,14 +418,27 @@ async def _load_tensor_parallel_state_dict(
             self.vllm_config.parallel_config.tensor_parallel_size, self.rank
         )
 
+        checkpoint_id = f"{self.state_dict_key}{DELIM}{version}"
+        dcp_metadata = None
+        if self.use_dcp:
+            dcp_metadata = await ts.get()
```
joecummings: Interesting - so an empty get just returns metadata? No key required?

joecummings:
Can you document some benchmark numbers here? I heard through the grapevine that DCP is actually WAAAAAAY faster than Torchstore :)

src/forge/actors/policy.py
LucasLLC: noob mistake

LucasLLC:
> Can you document some benchmark numbers here? I heard through the grapevine that DCP is actually WAAAAAAY faster than Torchstore :)

cries in performañol

PR #177, opened 2025-09-18
Add Batch routing support via `@service_endpoint` with configurable batch size and timeout

DNXie:
Add batch routing to `Service` to improve request throughput and maintain session-aware routing.

-  Added new `@service_endpoint` decorator that supports routing configuration (`router`, `batch_size`, `batch_timeout`).
- Introduced `ServiceEndpointProperty` to distinguish between `@endpoint` and `@service_endpoint`.
- Centralized endpoint-to-router mapping in `Service` (`self.routers`) with support for both plain routers and batchers.
- Updated `ServiceInterface` to register endpoints through `_set_router`, ensuring consistent setup for both standard and service endpoints.
-  Extended `_call` and `_get_replica` to handle batch routing, session routing, and fallback routing in a unified way.
- Enhanced `Service.stop` to gracefully shut down any active batchers in addition to replicas.
- Added integration tests to validate:

  * Round-robin distribution with and without batching
  * Correct batch flushing when `batch_size` is reached
  * Independent coexistence of multiple endpoints with different batch sizes/routers

**Test**

```bash
pytest tests/unit_tests/test_router.py
```


src/forge/controller/service/router.py
```diff
@@ -88,3 +90,111 @@ def get_replica(
...
+            sess_ids = []
+            start_time = time.time()
+
+            # Wait for first request
+            fut, healthy_replicas, sess_id, session_map = await self._queue.get()
+            batch.append((healthy_replicas, sess_id, session_map))
+            futs.append(fut)
+            sess_ids.append(sess_id)
+
+            while True:
```
allenwang28, line 189: probably add a `self._running` or something equivalent, therefore instead of 

```
while True:
    ...
```

we can 
```
while self._running:
    ...
```

so we can cleanly shut this down

src/forge/controller/service/router.py
```diff
@@ -88,3 +90,111 @@ def get_replica(
...
+                try:
+                    timeout = max(0, self.batch_max_wait_s - (time.time() - start_time))
```
allenwang28: ChatGPT says that `time.monotonic()`may be better here

src/forge/controller/service/router.py
```diff
@@ -88,3 +90,111 @@ def get_replica(
...
+                    (
+                        fut,
+                        healthy_replicas,
+                        sess_id,
+                        session_map,
+                    ) = await asyncio.wait_for(self._queue.get(), timeout)
```
allenwang28: hmm, is this operation saying "wait for timeout or until queue.get() finishes"? 

If that is, then I think it makes sense -
1. get first request
2. wait for up to `timeout - elapsed` or the first request
3. repeat until time our or we hit batch size

src/forge/controller/service/router.py
```diff
@@ -88,3 +90,111 @@ def get_replica(
...
+                    futs.append(fut)
+                    sess_ids.append(sess_id)
+
+                    if len(batch) >= self.batch_max_size:
+                        break
+                except asyncio.TimeoutError:
+                    break
+
+            # One routing decision for the whole batch
+            healthy_replicas = batch[-1][0]  # use most recent replica state
```
allenwang28: hmm, the router should also get a reference to the service so it can poll for healthy_replicas and session_map directly

src/forge/controller/service/router.py
```diff
@@ -88,3 +90,111 @@ def get_replica(
...
+
+    async def get_replica(
+        self,
+        healthy_replicas: List[Replica],
+        sess_id: Optional[str] = None,
+        session_map: Optional[Dict[str, int]] = None,
+    ) -> Replica:
+        """Enqueue request and wait until batch assigns a replica."""
+        loop = asyncio.get_event_loop()
+        fut = loop.create_future()
```
allenwang28: can we not just do `asyncio.Future()`?

src/forge/controller/service/router.py
DNXie: Yes, `await asyncio.wait_for(self._queue.get(), timeout)` means exactly what you said. It mean: wait until a new item is put into the queue, but give up and raise TimeoutError if nothing arrives within timeout seconds.

I spotted a bug in my previous implementation. I should move `start_time = time.time()` to **after (not before)** it receives the first item. Now the logic should be exactly what you described.

src/forge/controller/service/router.py
DNXie: Added. And added `shutdown` function to shutdown a batchrouter.

src/forge/controller/service/router.py
DNXie: hmmm.. But service would have a `self.router` for sure. Now router has a `self.service`? Not sure if this is a good practice

DNXie: Alternatively maybe we could pass two function pointers to the `BatchRouter.__init__` 
```
get_healthy_replicas=lambda: service.get_healthy_replicas(),
get_session_map=lambda: service.session_map,
```

allenwang28: hmm yeah that could work. Nit it'd probably be
```
router = Router(
    get_healthy_replicas=self.get_healthy_replicas,
    get_session_map=self.get_session_map,
)
```

src/forge/controller/service/router.py
allenwang28: hmm for some reason i'm not seeing these changes, did you push them?

DNXie: Oops. my local branches are a little messed up. Fixed now.

src/forge/controller/service/router.py
DNXie: Updated `BatchRouter`. It is tested in `test_batchrouter_callable_updates` (in `test_service.py`).

Since `BatchRouter` is not yet integrated with the service, this test runs the router **standalone** rather than through the service class. I found a way to verify batching, replica selection, and session map updates independently.

src/forge/controller/service/service.py
```diff
@@ -111,6 +112,13 @@ async def __initialize__(self):
         self._default_router = RoundRobinRouter()
         self._session_router = SessionRouter(fallback_router=LeastLoadedRouter())
 
+        # Batching
+        self._max_batch_size = self._cfg.max_batch_size
```
allenwang28: so a complication here is that the `max_batch_size` and `batch_max_wait_s` are dependent on the endpoint itself! Not a global service level concept.

so we would need to introduce some way to mark the endpoints we want batched, because not all of them should be. Here's what I had before:

```
class ReferenceActor(Actor):
    @service_endpoint(router=BatchedRouter(timeout_in_s=5, batch_size=4)) # batch either when you hit 5s, or batch size is 4
    def forward(self, token_batch: list[int]) -> list[torch.Tensor]:
         ...
```

DNXie: I see. One way I can think of is to move the routing logic entirely to [ServiceEndpoint](https://github.com/meta-pytorch/forge/blob/c868aeea68d0480178b67acdf673fdd6bad3927f/src/forge/controller/service/interface.py#L80)?

```python
class ServiceEndpoint(Generic[P, R]):
    """Service endpoint that owns its own routing + batching logic."""

    def __init__(
        self,
        service,
        endpoint_name: str,
        router=None,
        max_batch_size: int = 1,
        batch_max_wait_s: float = 0.01,
    ):
        self.service = service
        self.endpoint_name = endpoint_name
        self.router = router or RoundRobinRouter()  # default
        self.max_batch_size = max_batch_size
        self.batch_max_wait_s = batch_max_wait_s

        # Only set up batching infra if batching is enabled
        if self.max_batch_size > 1:
            self._batch_queue = asyncio.Queue()
            self._batch_task = asyncio.create_task(self._batch_loop())
        else:
            self._batch_queue = None
            self._batch_task = None

    async def route(self, *args: P.args, **kwargs: P.kwargs) -> R:
        """Route request to one replica using this endpoint’s router."""
        sess_id = kwargs.pop("sess_id", None)

        if sess_id is not None:
            replica = await self.get_replica(sess_id)
            return await self.service._call(sess_id, replica, self.endpoint_name, *args, **kwargs)

        if self.max_batch_size > 1:
            fut = asyncio.get_event_loop().create_future()
            # enqueue the future and request payload
            await self._batch_queue.put((fut, args, kwargs))
            return await fut

        # STATELESS and no batching -> route & call immediately
        replica = await self.get_replica(None)
        return await self.service._call(None, replica, self.endpoint_name, *args, **kwargs)


    async def fanout(self, *args: P.args, **kwargs: P.kwargs) -> List[R]:
        # stay the same because it iterates all replicas
        ...

    async def get_replica(self, sess_id: str | None) -> "Replica":
        """Encapsulate replica selection logic (session vs stateless)."""
        healthy = self.service._get_healthy_replicas()

        return await self.router.get_replica(healthy, sess_id, self.service._session_replica_map)

    async def _batch_loop(self):
       ...


```

allenwang28: hmm yeah I think that makes sense!

DNXie:
@allenwang28 I found out that this [try-except clause ](https://github.com/meta-pytorch/forge/blob/e76f2a096a9cd5a498aee8f768435070c44dd53d/src/forge/controller/service/service.py#L186-L197)actually doesn't catch error.

Since I moved this logic to `ServiceEndpoint` with `self.max_attempts=2`. Our existing test cases doesn't cover this logic. I was trying to test whether it really works with this test case:


```python
class Counter(ForgeActor):
  @endpoint()
  async def flaky(self):
      if not hasattr(self, "_flaky_calls"):
          self._flaky_calls = 0
      self._flaky_calls += 1
      if self._flaky_calls == 1:
          raise RuntimeError("fail first attempt")
      return "ok"
```

```python
@pytest.mark.timeout(10)
@pytest.mark.asyncio
async def test():
    service = await Counter.options(procs=1, num_replicas=2).as_service(v=0)

    try:
        service.max_attempts = 2

        result = await service.flaky.route()
        assert result == "ok"  # success after retry
    finally:
        await service.shutdown()
```

But somehow it doesn't go through. It seems like this error is handled in monarch level?
```bash
INFO     forge.controller.service.endpoint:endpoint.py:109 Get replica 1 on attempt 0
[0] CRITICAL:root:Unhandled exception in actor endpoint
[0] Traceback (most recent call last):
[0]   File "/home/dxie/.fbpkg_conda_envs/forge-a7401c7/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 831, in instrumented
[0]     result = await the_method(*args, **kwargs)
[0]   File "/home/dxie/forge/tests/unit_tests/test_router.py", line 64, in flaky
[0]     raise RuntimeError("fail first attempt")
[0] RuntimeError: fail first attempt
```

src/forge/controller/service/service.py
DNXie: Done!

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,368 @@
...
+                   - "leastloaded": pick the replica with the lowest load.
+                 Default: "round_robin".
+
+    Supported methods:
+        - `route`: Send a request to a single replica, chosen by the configured router
+        (e.g. round-robin, least-loaded).
+        - `fanout`: Broadcasts the request to all healthy replicas.
+
+    Notes:
+        - Support `@endpoint()` and `@service_endpoint(router='..')` decorators.
```
allenwang28:
```suggestion
        - Supports Monarch's `@endpoint()` as well as service's `@service_endpoint(router='..')` decorators.
```

for documentation posterity, let's specify that the @endpoint is Monarch's endpoint

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,368 @@
...
+
+class ServiceEndpoint(Generic[P, R]):
+    """
+    This extends Monarch's actor APIs for service endpoints.
+    ServiceEndpoint provides the basic, non-batched routing API for Forge services.
+
+    Args:
+        service: The underlying service object that owns replicas.
+        endpoint_name (str): The name of the endpoint method.
+        router (str, optional): Routing strategy for stateless requests.
```
allenwang28: hmm what do you think about providing a Router object instead? Like instead of "round_robin" as input the user provides RoundRobinRouter

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,368 @@
...
+
+from .router import LeastLoadedRouter, RoundRobinRouter, Router, SessionRouter
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+class ServiceEndpoint(Generic[P, R]):
+    """
+    This extends Monarch's actor APIs for service endpoints.
```
allenwang28, line 26: ServiceEndpoint extends Monarch's native actor APIs for service functionality.

Services provide fault tolerance and load balancing on top of Monarch actors, exposing the following endpoints:
- route() - (insert the definition here)
- fanout() - (insert the definition here)

Note that Monarch's native actor APIs are not accessible through service endpoints.

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,368 @@
...
+        # Primary router (stateless routing)
+        self.router = self._resolve_router(router)
+
+        # Session-aware router for sticky sessions
+        self.session_router = SessionRouter(fallback_router=self.router)
+
+        # Number of routing attempts (initial + retries)
+        self.max_attempts = 1
+
+    def _resolve_router(self, router_name: str) -> Router:
```
allenwang28: the router input comment from above will remove the need for this step

src/forge/controller/service/interface.py
```diff
@@ -182,10 +85,33 @@ def __init__(self, _service, actor_def):
...
+                    )
+                else:
+                    endpoint = ServiceEndpoint(
+                        self._service,
+                        attr_name,
+                        router=cfg["router"],
+                    )
+
+            elif isinstance(attr_value, EndpointProperty):
+                # Decorated with @endpoint
```
allenwang28:
```suggestion
                # This was defined as a standard Monarch endpoint
```

src/forge/controller/service/interface.py
```diff
@@ -182,10 +85,33 @@ def __init__(self, _service, actor_def):
         # Inspect the actor_def directly to find endpoints
         for attr_name in dir(actor_def):
             attr_value = getattr(actor_def, attr_name)
-            if isinstance(attr_value, EndpointProperty):
+            if hasattr(attr_value, "_service_endpoint_config"):
```
allenwang28: can we somehow make it `if isinstance(attr_value, ServiceEndpointProperty)` for consistency?

DNXie: Done!

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,167 @@
...
+        return await self.actor_mesh.call.call_one(
+            sess_id, self.endpoint_name, *args, **kwargs
+        )
+
+    async def call(self, *args: P.args, **kwargs: P.kwargs) -> List[R]:
+        """Broadcasts a request to all healthy replicas and returns the results as a list."""
+        result = await self.actor_mesh.call_all.call_one(
+            self.endpoint_name, *args, **kwargs
+        )
+        return result
```
allenwang28, line 113: can we copy paste the other adverbs above? like the `fanout()` / `route()` / the other Monarch adverbs causing an error etc.

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,167 @@
...
+        method: Any,
+        propagator: Propagator,
+        explicit_response_port: bool,
+        *,
+        router: Router = RoundRobinRouter(),
+        batch_size: int = 1,
+        batch_timeout: float = 0.01,
+    ) -> None:
+        super().__init__(method, propagator, explicit_response_port)
+        self._service_endpoint_config = dict(
```
allenwang28: why does this need to be a dict? Can we not just do `self.router = router` etc.?

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,167 @@
...
+
+    async def call(self, *args: P.args, **kwargs: P.kwargs) -> List[R]:
+        """Broadcasts a request to all healthy replicas and returns the results as a list."""
+        result = await self.actor_mesh.call_all.call_one(
+            self.endpoint_name, *args, **kwargs
+        )
+        return result
+
+
+class ServiceEndpointProperty(EndpointProperty, Generic[P, R]):
```
allenwang28, line 146: hmm, I don't think we actually need to inherit from the base `EndpointProperty` but let me know if I'm wrong

src/forge/controller/service/service.py
```diff
@@ -108,7 +109,8 @@ async def __initialize__(self):
 
         # Initialize the routers
         self._default_router = RoundRobinRouter()
-        self._session_router = SessionRouter(fallback_router=LeastLoadedRouter())
+        self._session_router = SessionRouter(fallback_router=self._default_router)
+        self.routers: dict[str, Router | Batcher] = {}
```
allenwang28, line 116: let's mark as a comment that this keeps the map between the registered endpoints and the routing functions

src/forge/controller/service/service.py
```diff
@@ -138,6 +140,56 @@ async def __initialize__(self):
...
+                {
+                    "router": Router,
+                    "batch_size": int,
+                    "batch_timeout": float
+                }
+        Returns:
+            Router | Batcher instance stored in self.routers
+        """
+
+        # If router already exists, ignore
```
allenwang28: this seems like it would be masking a bug. Why should we expect a router to exist already if we're only calling this when we create a service?

src/forge/controller/service/service.py
```diff
@@ -168,7 +220,7 @@ async def _call(self, sess_id: str | None, function: str, *args, **kwargs):
             if ctx:
                 sess_id = ctx["session_id"]
 
-        replica = await self._get_replica(sess_id)
+        replica = await self._get_replica(sess_id=sess_id, endpoint_name=function)
```
allenwang28: can we move the entire `_get_replica(...)` logic here into `_call`? since the `_get_replica()` logic is now contained in the batcher, I think we don't need this separate `_get_replica()` functionality.

Additionally, can we rename the `_call(...)` and `_call_all(...)` to `_route()` and `_fanout()` respectively?

src/forge/controller/service/endpoint.py
DNXie: I think the purpose is for it to behave exactly like [`EndpointProperty`](https://github.com/meta-pytorch/monarch/blob/0a842ad3c4abfb3983eb7ba55ec327c42553ff44/python/monarch/_src/actor/endpoint.py#L232-L260) other than having additional config

src/forge/controller/service/endpoint.py
DNXie: Because if it is a batcher, we need to pass two functions `get_healthy_replicas` and `get_session_map`
```python
class Batcher:
    def __init__(
        self,
        inner_router: Router,
        get_healthy_replicas: Callable[[], List["Replica"]],
        get_session_map: Callable[[], Dict[str, int]],
        batch_size: int = 16,
        batch_timeout: float = 0.01,
    ):
```
These two functions are from a `service` object. So `Batcher` needs to be initialized after Service is initialized.

src/forge/controller/service/service.py
DNXie: Yeah. This should be a bug. Raised an exception instead

src/forge/controller/service/service.py
DNXie: Done

src/forge/controller/service/endpoint.py
allenwang28: ahh ok I understand. the `EndpointProperty` is needed because the actors that are created by the service still need this so that the endpoints are exposed correctly

Could you add a comment in the docstring about this? i.e. why we're inheriting from EndpiontProperty

src/forge/controller/service/endpoint.py
allenwang28: my question is more about why it is a dictionary

src/forge/controller/service/endpoint.py
DNXie: Done

src/forge/controller/service/endpoint.py
DNXie: Oh I see. Yeah we don't need a dict here. We can keep things flat. Done.

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,199 @@
...
+    ) -> None:
+        super().__init__(method, propagator, explicit_response_port)
+        self.router = router
+        self.batch_size = batch_size
+        self.batch_timeout = batch_timeout
+
+
+def service_endpoint(
+    *,
+    router: Router = RoundRobinRouter(),
```
allenwang28: can this instead be a router constructor? Like right now we pass in a full Router object - probably fine to start with, but typically for efficiency reasons we pass in a constructor function, which we create later.

So it'd look like

```
router: Callable[[], Router]
```

and 
```
        class MyForgeActor(ForgeActor):
            @service_endpoint(router=RoundRobinRouter, batch_size=16, batch_timeout=0.05)
            async def predict(self, x): ...
```

src/forge/controller/service/interface.py
```diff
@@ -182,10 +85,15 @@ def __init__(self, _service, actor_def):
         # Inspect the actor_def directly to find endpoints
         for attr_name in dir(actor_def):
             attr_value = getattr(actor_def, attr_name)
-            if isinstance(attr_value, EndpointProperty):
-                # Create a ServiceEndpoint that will route through the Service Actor
-                endpoint = ServiceEndpoint(self._service, attr_name)
-                setattr(self, attr_name, endpoint)
+
+            # ServiceEndpointProperty: created by @service_endpoint
```
allenwang28, line 89: can you make these changes in ServiceV2 too? if it's not too much of a lift - it would make it really easy for us to switch to ServiceV2 whenever we're ready to!

src/forge/controller/service/service.py
```diff
@@ -138,14 +143,61 @@ async def __initialize__(self):
...
+
+        Args:
+            endpoint_name: Name of the endpoint.
+            prop: Optional ServiceEndpointProperty object with router, batch_size,
+                  and batch_timeout attributes.
+        """
+
+        # If router already exists, raise an exception
+        if endpoint_name in self.routers:
+            raise ValueError(f"Router already exists for endpoint: {endpoint_name}")
```
allenwang28: nit: maybe this should be an AssertionError

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,126 @@ def get_replica(
...
+                        break
+                except asyncio.TimeoutError:
+                    break
+
+            session_map = self.get_session_map()
+            healthy_replicas = self.get_healthy_replicas()
+
+            # One routing decision for the whole batch
+            replica = self.inner_router.get_replica(healthy_replicas, None, session_map)
+
```
allenwang28, line 209: hmm, technically what should happen here is that a batch of requests is sent to the replica. I think what's happening right now is that you're calling the replica's endpoint `batch_size` times, which is exactly the opposite of what we want!

src/forge/controller/service/interface.py
DNXie: Done

src/forge/controller/service/endpoint.py
DNXie: Got it. Done!

src/forge/controller/service/service.py
DNXie: Done

src/forge/controller/service/router.py
DNXie: Yeah you are right. Just updated to make the batcher send the batched requests to the replica. Please check!

src/forge/controller/service/replica.py
```diff
@@ -220,6 +220,11 @@ async def enqueue_request(self, request: ServiceRequest):
         # Accept requests in all other states - let the processing loop handle the rest
         await self.request_queue.put(request)
 
+    async def enqueue_batch(self, requests: list[ServiceRequest]):
```
allenwang28, line 223: hmm, ok this is complicated. The right behavior is actually that, as a user I want to be able to say:

```
class MyActor(ForgeActor):
    @service_endpoint(batch_size=4, duration_time_s=1) # (you can ignore the wording I'm using here)
    def forward(self, batch: list[torch.Tensor]) -> torch.Tensor:
        ...
```

but be able to use it as 
```
service = MyActor.as_service()

result: torch.Tensor = service.forward.route(torch.zeros(10))
```

i.e., you define it as a batch function but get individual results back

DNXie: This is working properly. See the test case in `test_router.py::test_batch_endpoint_returns_individual_results`

PR #178, opened 2025-09-18
Core Data Models

vidhyav:
Rithesh's pr commandeered: #157

### DIFF
--- a/src/forge/data_models/scored_completion.py
+++ b/src/forge/data_models/scored_completion.py
@@ -1,3 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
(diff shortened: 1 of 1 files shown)


allenwang28:
Can we do this incrementally? Let's start a PR that integrates completion and experience, with Experience renamed to Episode (consistent with what we have now)

I want that, plus the ability to convert from e.g. vLLM native outputs to our Forge native outputs.

An example from GRPO main: https://github.com/meta-pytorch/forge/blob/main/apps/grpo/main.py#L311-L313, we should be able to convert from the vLLM request to a `forge.data_types.CompletionOutput`, then create an `Episode` from only forge native data types. Ideally like:

```
responses = await policy.generate.choose(prompt)
completions = forge.data_type.CompletionOutput.from_vllm(responses)
```

etc.

vidhyav:
Huh this was a set of stacked diffs and it is painful to commandeer AND modify. I am going to simply restart a new diff. SIGH.

PR #179, opened 2025-09-18
Core trainer abstraction

vidhyav:
I was unable to commandeer so did this.


Jack-Khuu:
Different Version landed in main

PR #180, opened 2025-09-18
Core generator abstraction. 

vidhyav:
Commandeering.


src/forge/data_models/prompt.py
```diff
@@ -0,0 +1,66 @@
...
+
+class Role(Enum):
+    SYSTEM = "system"
+    USER = "user"
+    ASSISTANT = "assistant"
+    NONE = "none"
+
+
+@dataclass
+class Message:
```
Ritesh1905, line 21: Change this to 

```
class Message(TypedDict):
    role: str
    content: str | dict[str, Any]
    tools: dict[str, Any] | None
```

- Role can be enum.
- remove metadata from prompt

PR #181, opened 2025-09-18
Update logic to only track policy version in the main controller 

JenniferWang:
## Before

Policy version is tracked in two places: main controller and Policy Actor (which increments the version number upon `update_weights` call). This feels fragile and unnecessary.

## After
This PR is not to debate the need for a versioning number but rather assuming the concept is valid, only track it in the controller. 

## Tests

Run on both grpo and toy_rl. But mainly look at progress on the toy_rl

https://www.internalfb.com/phabricator/paste/view/P1954914879

No regression 
Before: https://meta.wandb.io/jiyue/sumdigits-training/runs/omqmpkq1?nw=nwuserjiyue
image

After: https://meta.wandb.io/jiyue/sumdigits-training/runs/xv5dduh5?nw=nwuserjiyue
image

### DIFF
--- a/apps/toy_rl/sumdigits.py
+++ b/apps/toy_rl/sumdigits.py
@@ -397,13 +397,7 @@ async def main(cfg: DictConfig):
 
     # ---- Setup services ---- #
     await ts.initialize()
-    (
-        dataloader,
-        policy,
-        trainer,
-        replay_buffer,
-        reward_actor,
-    ) = await asyncio.gather(
+    (dataloader, policy, trainer, replay_buffer, reward_actor,) = await asyncio.gather(
         DatasetActor.options(**cfg.services.dataset).as_service(**cfg.dataset),
         Policy.options(**cfg.services.policy).as_service(**cfg.policy),
         Trainer.options(**cfg.services.trainer).as_service(**cfg.trainer),


apps/grpo/main.py
```diff
@@ -358,6 +358,7 @@ async def continuous_rollouts():
                 return
             prompt, target = sample["request"], sample["target"]
             responses = await policy.generate.choose(prompt)
+            # TODO: this shall be part of the responses metadata instead of a separate call
```
joecummings, line 289: 👍

src/forge/actors/replay_buffer.py
```diff
@@ -87,11 +91,18 @@ async def evict(self, curr_policy_version: int) -> None:
         self._evict(curr_policy_version)
 
     def _evict(self, curr_policy_version: int) -> None:
+        buffer_len_before_evict = len(self.buffer)
```
joecummings, line 97: What do you think about leaving this information at the top level script like this? https://github.com/meta-pytorch/forge/pull/181/files#diff-3bdea8452f8678efbb5410b0618ec2942e39e2678db862e9abf4c2fabf507fd8R338

It lets the user decide how much to log instead of opting them into this information ever time they sample.

src/forge/actors/replay_buffer.py
```diff
@@ -119,3 +131,6 @@ async def state_dict(self) -> dict[str, Any]:
     async def load_state_dict(self, state_dict: dict[str, Any]) -> None:
         self.buffer = state_dict["buffer"]
         random.setstate(state_dict["rng_state"])
+
+    def __post_init__(self):
```
joecummings: nit: Move this up to be the first method.

JenniferWang:
I'd like to test on https://github.com/meta-pytorch/forge/pull/183 as well...

src/forge/actors/replay_buffer.py
```diff
@@ -106,6 +117,7 @@ async def _numel(self) -> int:
     async def clear(self) -> None:
         """Clear the replay buffer immediately - dropping all episodes."""
         self.buffer.clear()
+        logger.info("replay buffer cleared")
```
joecummings: Similar to my other comment, could we leave this out? Or at least switch it to be debug only?

apps/toy_rl/sumdigits.py
```diff
@@ -464,21 +464,18 @@ async def continuous_rollouts():
...
                 loss = await trainer.train_step.choose(batch[0])
                 training_step += 1
                 mlogger.log("loss/training_step", loss, training_step)
-                print(f"loss/training_step: {loss} at {training_step}")
-                await trainer.push_weights.call(policy_version)
-                policy_version += 1
-                await policy.update_weights.call()
+                print(f"loss/training_step: {loss} at training step {training_step}")
+                await trainer.push_weights.call(training_step)
+                await policy.update_weights.call(training_step)
```
Ritesh1905, line 477: _# TODO: remove this line? this clears the buffer so it's always on-policy_

This was intentional so that the (at least the ideas was that the) metrics can converge quickly.

PR #183, opened 2025-09-19
Sumdigits exmaple with GRPO

Ritesh1905:
The reward-mean converges in less than 15 training steps and takes ~5 mins 

wandb run: https://meta.wandb.io/torchforge/sumdigits-training/runs/uxzowpkp?nw=nwuserrithesh

image

- Made some corrections to GRPO loss function.
- Updates the sumdigits example to use GRPO loss
- Introduces a small curriculum setup.

TODO:

- Lot of code redundancy, which will be fixed once the abstractions are landed.

### DIFF
--- a/src/forge/losses/reinforce_loss.py
+++ b/src/forge/losses/reinforce_loss.py
@@ -5,9 +5,9 @@
 # LICENSE file in the root directory of this source tree.
 
 import torch
+from torch import nn
 
 from forge.util.ops import selective_log_softmax
-from torch import nn
 
 
 class ReinforceLoss(nn.Module):
--- a/src/forge/util/ops.py
+++ b/src/forge/util/ops.py
@@ -8,7 +8,7 @@ import torch
 import torch.nn.functional as F
 
 
-def selective_log_softmax(logits, index) -> torch.Tensor:
+def selective_log_softmax(logits: torch.Tensor, index: torch.Tensor) -> torch.Tensor:
     """
     A memory-efficient implementation of the common `log_softmax -> gather` operation.
 


src/forge/util/model_utils.py
joecummings, line 1: nit: util/model_utils.py -> utils/ops.py

src/forge/util/model_utils.py
```diff
@@ -0,0 +1,51 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import torch.nn.functional as F
+
+
+def selective_log_softmax(logits, index) -> torch.Tensor:
```
joecummings: Could be include a quick test for this as well? Plus some typing for logits and index

src/forge/losses/grpo_loss.py
```diff
@@ -9,21 +9,70 @@
 
 
 class SimpleGRPOLoss(nn.Module):
-    """Simplified GRPO Loss for simplified single step updates
-    Inspired by the Hugging Face TRL implementation:
-        https://github.com/huggingface/trl/blob/417915a3e4d3e3bc8d7b196594308b8eabf928be/trl/trainer/grpo_trainer.py#L1624.
+    """
+    Group Relative Policy Optimization (GRPO) Loss Function.
```
joecummings: Let's discuss offline. In general, I think we should have a GRPOLoss that follows exactly the original spec: https://huggingface.co/docs/trl/en/grpo_trainer#loss-types and then something like DrGRPO which is shown to be more stable wrt length and difficulty bias: https://huggingface.co/docs/trl/en/grpo_trainer#loss-types

src/forge/losses/grpo_loss.py
```diff
@@ -9,21 +9,70 @@
...
+        """
+        # Compute log ratio: log(π_θ/π_ref) = log(π_θ) - log(π_ref)
+        log_ratio = logprobs - ref_logprobs
+
+        # Policy gradient loss: -log_ratio * advantages
+        per_token_policy_loss = log_ratio * advantages
+
+        # KL penalty: KL(π_θ || π_ref) ≈ π_θ * (log(π_θ) - log(π_ref))
+        # For log probabilities, this becomes: exp(logprobs) * (logprobs - ref_logprobs)
+        kl_penalty = torch.exp(logprobs) * (logprobs - ref_logprobs)
```
Ritesh1905: @joecummings , my bad, This is incorrect... It should be  same as original implementation.

```
log_ratio = ref_logprobs - logprobs
kl_penalty = torch.exp(log_ratio) - log_ratio - 1
```

I will revert the chnages to this file

Ritesh1905: reverted changes to this file.

PR #184, opened 2025-09-19
Weight loading working correctly with tp: use vllm builtin load_weights()

casteryh:
Add a flag use_vllm_bultin_load to policy and trainer.
When set to true, will use vllm builtin load_weights() method to exchange weights.
In particular, this works correctly with TP.
Tested with sumdigits example tp_size = 2.
https://meta.wandb.io/torchforge/sumdigits-training/runs/f9jb060e/panel/2pa6e8ptg?nw=nwuseryuxuanh
avg_reward >0.9 in 3k steps
W B Chart 9_23_2025, 10_22_37 PM

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -18,21 +18,6 @@ from dataclasses import asdict, dataclass, field, fields
 import torch
 import torch.distributed.checkpoint as dcp
 import torchstore as ts
-
-from forge.actors._torchstore_utils import (
-    DcpHandle,
-    extract_param_name,
-    get_param_key,
-    get_param_prefix,
-    load_tensor_from_dcp,
-)
-
-from forge.controller import ForgeActor, get_proc_mesh, stop_proc_mesh
-from forge.data.sharding import VLLMSharding
-from forge.data_models.completion import Completion
-from forge.data_models.prompt import to_prompt
-from forge.interfaces import Policy as PolicyInterface
-from forge.types import ProcessConfig
 from monarch.actor import current_rank, endpoint, ProcMesh
 from torchstore.state_dict_utils import DELIM
 from vllm.config import VllmConfig
@@ -56,6 +41,21 @@ from vllm.v1.engine.processor import Processor
 from vllm.v1.request import Request
 from vllm.v1.structured_output import StructuredOutputManager
 from vllm.worker.worker_base import WorkerWrapperBase
+
+from forge.actors._torchstore_utils import (
+    DcpHandle,
+    extract_param_name,
+    get_param_key,
+    get_param_prefix,
+    load_tensor_from_dcp,
+)
+
+from forge.controller import ForgeActor, get_proc_mesh, stop_proc_mesh
+from forge.data.sharding import VLLMSharding
+from forge.data_models.completion import Completion
+from forge.data_models.prompt import to_prompt
+from forge.interfaces import Policy as PolicyInterface
+from forge.types import ProcessConfig
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)


apps/grpo/main.py
```diff
@@ -397,20 +397,28 @@ async def continuous_rollouts():
...
-                await trainer.push_weights.call(policy_version)
-                policy_version += 1
-                await policy.update_weights.call()
+                start_time = time.perf_counter()
+                await trainer.push_weights.call(training_step)
+                mlogger.log(
+                    "push_weights_time/training_step",
+                    time.perf_counter() - start_time,
+                    training_step,
+                )
```
JenniferWang: This is great! I think let's split this diff to 
1. add weight sync counter
2. add options to do per-tensor weight sync

src/forge/actors/policy.py
```diff
@@ -364,16 +369,16 @@ async def run(self):
                     fut.set_result(request_output)
 
     @endpoint
-    async def update_weights(self):
+    async def update_weights(self, policy_version: int):
```
JenniferWang, line 391: You probably want to rebase on this https://github.com/meta-pytorch/forge/pull/181
I'll address the comments and merge the PR ASAP

src/forge/actors/policy.py
```diff
@@ -384,7 +393,10 @@ async def update_weights(self, policy_version: int):
             await asyncio.gather(*curr_requests)
 
         logger.debug(f"Starting weight update on {self.__class__.__name__}")
-        await self.policy_worker.update.call(version=policy_version)
+        if self.use_vllm_builtin_load:
```
vidhyav, line 399: Eventually,  this will be the default right?

src/forge/actors/policy.py
```diff
@@ -505,6 +517,27 @@ async def update(self, version: int):
         await self._load_tensor_parallel_state_dict(current_state_dict, version)
         logger.debug(f"Loaded state dict from {key} in {time.time() - start} seconds")
 
+    @endpoint
+    async def _update_hf_nonsharded(self, version: int):
```
vidhyav: Why is this specific to hf??

vidhyav:
Is the source of the problem also because we have deconstructed the vllm engine and using it piecemeal instead of using it as a whole?

src/forge/actors/trainer.py
```diff
@@ -248,6 +251,12 @@ def train_step(
 
     @endpoint
     async def push_weights(self, policy_version: int) -> None:
+        if self.use_vllm_builtin_load:
+            await self._push_weights_hf_nonsharded(policy_version)
+        else:
+            await self._push_weights_sharded(policy_version)
+
+    async def _push_weights_sharded(self, policy_version: int) -> None:
```
JenniferWang: I have some confusion: in my fix, I feel that this is not sharded.  The difference is whether to process the state dict or not. Basically we just need to skip this path `_qwen3_hf_to_vllm` and keep the rest as is?

src/forge/actors/policy.py
casteryh: seems like the plan

src/forge/actors/policy.py
casteryh: this just means we are pushing/reading the state dict using the hugging face format. not titan, not vllm.

JenniferWang: Perhaps call it `update_DEPRECATED` and `update`. I'd like to keep the `DEPRECATED` one just for A/B testing and delete it before the PTC.

JenniferWang: Could you explain the choice between using `get_state_dict/get_state_dict` and the `get/put` API?

JenniferWang: I am also confused at the `load_weights` API -- will it handle sharding itself? If so, should we call this function on the driver worker (0) once?

casteryh: > I am also confused at the `load_weights` API -- will it handle sharding itself? If so, should we call this function on the driver worker (0) once?

every worker(rank) has to call `load_weights`.
After `load_weights()` is called, every worker will figure out its own rank and just read its own shard when `load_weights()` is called.

Moreover, `load_weights()` supports incremental updating, i.e., if there is only one tensor in the passed in weights, it will update that part specifically (it even handles these concatenated weights as well).
For example, if you pass in (I am making up the fqn but you get the point) a single kv pair `"model.layers.0.q_proj.xxx" -> full_tensor`, it will actually update the `q_proj` part of the fused `qkv_proj` weight.

apps/grpo/main.py
casteryh: added this for debugging myself, will do!

src/forge/actors/trainer.py
casteryh: yep, I am bad at naming things. it actually has nothing to do with sharding at this point.
maybe we just call this `push_weights_vllm` vs `push_weights_hf` (or `push_weights_DEPRECATED` vs `push_weights` if you will)

casteryh:
@JenniferWang 
> Could you explain the choice between using get_state_dict/get_state_dict and the get/put API?

`get_state_dict` will get all the weights all at once which we probably don't want.
I don't have problems with `put_state_dict`, but the thing is if I am not using `get_state_dict`, I am not sure how to properly read something written in with `put_state_dict`.
I'd rather have a flat kv structure and everything I can control myself.
Also, the torchstore api explicitly says `get_state_dict` and `put_state_dict` are for testing purposes (last time I checked the codebase). I got no idea why it's used here in the first place.

casteryh:
@vidhyav 
> Is the source of the problem also because we have deconstructed the vllm engine and using it piecemeal instead of using it as a whole?

What we are doing here is we are using vllm workers and let monarch handles the collective operations instead of vllm's default method. That said, we can still use `load_weights()` method on vllm *workers*.

- Ordinary vllm weight loading: vllm using its own collectives to call `load_weights()` on each worker.
- This PR / our approach: monarch calls `load_weights()` on each worker.
- Previous buggy approach: each vllm worker (wrapped in monarch actor) is trying to circumvent the `load_weights()` method all together and directly operate on the `state_dict()` of the underlying `nn.Module`.

apps/toy_rl/sumdigits.py
```diff
@@ -4,30 +4,32 @@
...
 from forge.controller.actor import ForgeActor
 from forge.controller.provisioner import shutdown
+
 from forge.losses.grpo_loss import SimpleGRPOLoss
 from forge.util.metric_logging import get_metric_logger
-
 from forge.util.ops import selective_log_softmax
 from monarch.actor import endpoint
 from omegaconf import DictConfig
+from torch.functional import _return_counts
```
JenniferWang:
```suggestion
```

apps/toy_rl/sumdigits.py
```diff
@@ -338,7 +343,14 @@ def train_step(self, episodes: list[Episode]) -> float:
         return loss.item()
 
     @endpoint
-    async def push_weights(self, version: int):
+    async def push_weights_DEPRECATED(self, policy_version: int):  # noqa: N802
```
joecummings: If we're confident in this fix, we should just fully delete the old way. My thinking is as follows:

1) Gets everyone immediately testing the new version for any bugs 👍 
2) Reduces the chance an end user sees and uses this endpoint 👍 
3) Less code to parse through right now 👍

src/forge/actors/policy.py
```diff
@@ -496,15 +519,37 @@ async def _load_tensor_parallel_state_dict(
...
+        self.logger.debug(f"{matching_keys=}")
+        # TODO: find a way to save the original huggingface parameter names.
+        hf_names = [extract_param_name(key) for key in matching_keys]
+        self.logger.debug(f"{hf_names=}")
+        loaded_weights = set()
+        # We can't pass a generator since vllm load_weights is not async.
+        # Instead, we just call load_weights with one parameter at a time.
+        for name in hf_names:
+            param = await ts.get(get_param_key(version, name))
+            loaded = model.load_weights([(name, param)])
```
joecummings, line 582: This is super cool! I didn't realize you could do it per-param :)

src/forge/actors/policy.py
```diff
@@ -496,15 +519,37 @@ async def _load_tensor_parallel_state_dict(
...
+            del param
+            loaded_weights.update(loaded)
+        self.logger.info(f"Updated {len(loaded_weights)} parameters")
```
joecummings: nit: I prefer the old debug message that prints out the time it took to update the weights

src/forge/actors/_torchstore_utils.py
joecummings, line 1: nit: underscore this file so that users know they should NOT be importing from these functions directly.

apps/toy_rl/sumdigits.py
casteryh: > Gets everyone immediately testing the new version for any bugs 👍

Yes the new one is the default now! I think the plan is keep the DEPRECATED method just for benchmarking purposes now? @JenniferWang

src/forge/actors/policy.py
casteryh: yeah it's surprisingly good

src/forge/actors/policy.py
casteryh: will add it back

apps/grpo/main.py
```diff
@@ -394,6 +393,10 @@ async def continuous_training():
 
     @parse
     def _main(cfg):
-        asyncio.run(main(cfg))
+        with TemporaryDirectory(prefix="forge_run_", dir="/dev/shm") as dcp_path:
```
joecummings: ?

casteryh: yeah this won't work lemme revert

PR #185, opened 2025-09-19
compute log probs in reference actor

tushar00jain:


Jack-Khuu:
cc @joecummings who made the opposite change

PR #191, opened 2025-09-19
metric logger - backend (1/N)

felipemello1:
Implemented it on a toy script. Will add real metrics to grpo main.py.

To log from anywhere:

```
from forge.observability.metrics import record_metric, ReductionType
await record_metric("generate/tokens", value, ReductionType.SUM)
```

## Testing

- 2 actors, each with 2 replicas and 2 ranks. 
- Used 2 backends at the same time (wandb and console). 
- Also tried all 3 distributed methods suggested by wandb. 
- Logged 2 metrics, one with sum reduction and another with mean reduction. 
- API and abstractions can probably can use some polishing, but they should be enough to work for different user needs.

image

## TLDR 3 ways of doing wandb distributed training 
https://docs.wandb.ai/guides/track/log/distributed-training/

### a) Track a single process: 
- only rank 0 instantiates wandb and logs. 
- User can manually reduce and send to rank 0.

Pros:
- Customizations, e.g. std, count, filtering, if we have 2 dp, 8tp, sum_num_tokens should be done for dp only.
- Easy single run to visualize

Cons:
- Needs cross rank reduction logic

### b) Track multiple processes: 
- Every rank instantiates wandb. 
- Runs can be grouped in UI by group name, e.g. “my_group_name”, defined during instantiation.

Pros:
- Maximum visibility per rank

Cons:
- Crowded: 32 ranks = 32 wandb runs
- Reduction is somewhat manual (per graph config - but perhaps we can set via code?) and limited (mean, max, min)

image

### c) Track all processes to a single wandb run ID (experimental): 
- Every rank instantiates wandb. 
- Primary rank generates a run_id and sends it to other ranks.

pros:
- Not sure

Cons:
- Can't use same x-axis
- Doesn't guarantee logging order

image

### 1) Every process should log to the same train step

```
for train_step in range(2):
	for generation_step in range(3):
		time_generation_step = generation_actor.choose()
		logger.log(time_generation_step, train_step)
	time_train_step = train_actor.choose()
	logger.log(time_train_step, train_step)
```

For each train_step we have 3 generation steps. They should all belong to the same train_step x-axis.

### 2) We need a local reduction per train step

If we want the mean `time_generation_step` per train_step, we have to reduce it ourselves. wandb only keeps the latests

### 3) We need a global and a local logger

Local loggers are responsible for:
		i)  per-rank wandb instantiation;
		ii)  collecting local data.

Global loggers are responsible for communicating:
		i) train step + flush command: necessary for all 3 wandb modes; 
		ii) shared wandb run_id: necessary for option C; 
		iii) optionally collecting and reducing to rank 0: Necessary for option A

Q1: Can we get rid of local_logger and just use global_logger for everything?
A1: Probably not. For options (b,c) we need to instantiate per rank. Perhaps we could instantiate N instances in a single rank, but this doesn’t make much sense to me. For option (a), we could use just a global logger, but it seems easier to manage it per rank.

### 4) User must have easy access to a dashboard with a general view of the system

This means that we should **NOT** require internal tools or external infra setup for basic metric logging. WandB seems to be the best fit for this.

## Recommendation

We should support (a) Track a single process with wandb and (b) Track multiple processes with jsonl dump per rank.

## Design

These 3 steps are independent:
a) collecting metrics
b) reducing metrics (we can decide to reduce, log per rank, log only on rank 0)
c) flushing metrics to wandb, scuba, jsonl, etc

Controller has a GlobalLoggingActor, every rank has a LocalFetcherActor + MetricLoggerSingleton

a. When a service is spawned, we also spawn a LocalFetcherActor per rank.
b. After every training step, a GlobalLoggingActor calls every LocalFetcherActor which calls MetricLoggerSingleton.

```
  ┌─────────────────────────────────────────────────────┐
  │                   Controller Process                │
  │ ┌───────────────────────────────────────────────┐   │
  │ │              GlobalLoggingActor               │   │
  │ │  • Coordinates all local loggers              │   │
  │ │  • Applies cross-process reductions           │   │
  │ │  • Dispatches to backends                     │   │
  │ └───────────────────────────────────────────────┘   │
  └─────────────────────────────────────────────────────┘
                           │
            ┌──────────────┼──────────────┐
            │                             │
     ┌──────▼───────┐               ┌─────▼───────┐
     │  Process A   │               │  Process B  │
     │ ┌───────────┐│               │┌───────────┐│
     │ │ Local     ││               ││ Local     ││
     │ │ Fetcher   ││               ││ Fetcher   ││
     │ │ Actor     ││               ││ Actor     ││
     │ └───────────┘│               │└───────────┘│
     │ ┌───────────┐│               │┌───────────┐│
     │ │ Metric    ││               ││ Metric    ││
     │ │ Logger    ││               ││ Logger    ││
     │ │ Singleton ││               ││ Singleton ││
     │ └───────────┘│               │└───────────┘│
     │ ┌───────────┐│               │┌───────────┐│
     │ │ Training  ││               ││ Policy    ││
     │ │ Actors    ││               ││ Actors    ││
     │ └───────────┘│               │└───────────┘│
     └──────────────┘               └─────────────┘

```

### Why?

Each rank can just call the singleton anywhere. This is easy and clean, but a bit counterintuitive at first, i.e. "push metrics to where?" The singleton buffers per-key/train_step, applies local reduction on fetch call.

```
import push_metrics

def push_metrics(key, value, reduction_type):
	metric_logger = MetricLoggerSingleton()
	metric_logger.append(key, value, reduction_type)

def my_random_actor_or_function():
	push_metrics('key', 1, ReductionTypes.SUM)
```

Other alternative is to instantiate a class at every actor and pass it around. But I am not sure how to fetch it from GlobalLoggingActor.
```
def my_random_actor_or_function(my_metric_logger):
	my_metric_logger.push_metrics('key', 1, ReductionTypes.SUM)
```

Q1 : Why not just use the `LocalFetcherActor` and drop the Singleton?
A1:  Because i couldn't find a way to find the `LocalFetcherActor` for each rank from the context

Q2: And how is the `GlobalMetricLogger` aware of `LocalFetcherActor` and `LocalLoggingSingleton`?
A2: in `get_proc_mesh`
```
# For each rank
local_fetcher_actor = await procs.spawn("local_fetcher_actor", LocalFetcherActor)

# Register with global logger
global_logger = await get_or_spawn_controller("global_logger", GlobalLoggingActor)
await global_logger.register_fetcher.call_one(local_fetcher_actor, process_name)
```

Then in the main loop:
```
for train_step in range(3):
	....
	await global_logger.flush.call_one(train_step)
```

### DIFF
--- a/apps/toy_metrics/main.py
+++ b/apps/toy_metrics/main.py
@@ -11,7 +11,7 @@ import time
 
 from forge.controller.actor import ForgeActor
 from forge.controller.provisioner import shutdown
-from forge.observability.metric_actors import get_metric_logger
+from forge.observability.metric_actors import setup_metric_logger
 from forge.observability.metrics import record_metric, ReductionType
 
 from monarch.actor import current_rank, endpoint
@@ -60,13 +60,13 @@ async def main():
     }
 
     service_config = {"procs": 2, "num_replicas": 2, "with_gpus": False}
+    mlogger = await setup_metric_logger()
 
     # Spawn services first (triggers registrations via provisioner hook)
     trainer = await TrainActor.options(**service_config).as_service()
     generator = await GeneratorActor.options(**service_config).as_service()
 
     # Now init config on global (inits backends eagerly across fetchers)
-    mlogger = await get_metric_logger()
     await mlogger.init_backends.call_one(config)
 
     for i in range(3):
--- a/src/forge/controller/provisioner.py
+++ b/src/forge/controller/provisioner.py
@@ -16,19 +16,12 @@ import uuid
 import monarch
 from monarch._src.actor.allocator import RemoteAllocator, TorchXRemoteAllocInitializer
 from monarch._src.actor.shape import NDSlice, Shape
-from monarch.actor import (
-    Actor,
-    endpoint,
-    get_or_spawn_controller,
-    HostMesh,
-    ProcMesh,
-    this_host,
-)
+from monarch.actor import Actor, endpoint, HostMesh, ProcMesh, this_host
 from monarch.tools import commands
 from monarch.tools.components import hyperactor
 from monarch.tools.config import Config
 
-from forge.observability.metric_actors import GlobalLoggingActor, LocalFetcherActor
+from forge.observability.metric_actors import setup_metric_logger
 
 from forge.types import ProcessConfig
 
@@ -151,24 +144,6 @@ class Provisioner:
             server_name,
         )
 
-    async def _setup_logging(self, procs: ProcMesh) -> None:
-        """Spawn and register local fetcher for metric logging on each process.
-
-        When any process is created, we create for each rank a LocalFetcherActor and
-        store it at GlobalLoggingActor. Backends (e.g. wandb) should be eagerly instantiated
-        later in main by calling `global_logger.initialize_backends.call_one(logging_config)`
-        """
-
-        local_fetcher_actor = await procs.spawn(
-            "local_fetcher_actor", LocalFetcherActor
-        )
-        procs._local_fetcher = local_fetcher_actor
-
-        global_logger = await get_or_spawn_controller(
-            "global_logger", GlobalLoggingActor
-        )
-        await global_logger.register_fetcher.call_one(local_fetcher_actor, procs)
-
     async def get_proc_mesh(
         self, num_procs: int, with_gpus: bool = False, num_hosts: int | None = None
     ):
@@ -243,7 +218,7 @@ class Provisioner:
                 self._proc_server_map[procs] = server_name
 
         # Spawn local logging actor on each process and register with global logger
-        await self._setup_logging(procs)
+        _ = await setup_metric_logger(procs)
 
         return procs
 
@@ -252,9 +227,7 @@ class Provisioner:
         async with self._lock:
             # Deregister local logger from global logger
             if hasattr(proc_mesh, "_local_fetcher"):
-                global_logger = await get_or_spawn_controller(
-                    "global_logger", GlobalLoggingActor
-                )
+                global_logger = await setup_metric_logger(proc_mesh)
                 await global_logger.deregister_fetcher.call_one(proc_mesh)
 
             if hasattr(proc_mesh, "_gpu_ids"):
--- a/src/forge/observability/metric_actors.py
+++ b/src/forge/observability/metric_actors.py
@@ -8,7 +8,7 @@ import asyncio
 import logging
 from typing import Any, Dict, Optional
 
-from monarch.actor import Actor, endpoint, get_or_spawn_controller
+from monarch.actor import Actor, endpoint, get_or_spawn_controller, ProcMesh, this_proc
 
 from forge.observability.metrics import (
     get_logger_backend_class,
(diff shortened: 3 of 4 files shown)


src/forge/controller/provisioner.py
```diff
@@ -119,6 +126,24 @@ async def create_host_mesh(self, name: str, num_hosts: int) -> HostMesh:
             server_name,
         )
 
+    async def _setup_logging(self, procs: ProcMesh) -> None:
```
felipemello1: ctx: when a service is spawned, we create for each rank a LocalFetcherActor and store it at GlobalLoggingActor.

backends are eagerly instantiated later in main by calling "await global_logger.initialize_backends.call_one(logging_config)"

src/forge/observability/metrics.py
```diff
@@ -0,0 +1,633 @@
...
+                )
+
+        metric_accumulator = ReductionType(first_reduction_type).accumulator_class
+        reduced_value = metric_accumulator.get_reduced_value_from_states(metric_states)
+        reduced_metrics[key] = reduced_value
+
+    return reduced_metrics
+
+
+class LoggerBackend(ABC):
```
felipemello1: tempted to move all backend related things to logger_backends.py

src/forge/controller/provisioner.py
allenwang28: This would be a good comment to put in this function!

src/forge/controller/provisioner.py
```diff
@@ -142,6 +149,28 @@ async def create_host_mesh(self, name: str, num_hosts: int) -> HostMesh:
...
+        """Spawn and register local fetcher for metric logging on each process.
+        When a service is spawned, we create for each rank a LocalFetcherActor and
```
allenwang28: nit: spacing between lines 153 and 154

apps/toy_metrics/main.py
```diff
@@ -0,0 +1,118 @@
...
+        trainer.shutdown(),
+        generator.shutdown(),
+    )
+
+    await shutdown()
+
+
+if __name__ == "__main__":
+    mode = sys.argv[1] if len(sys.argv) > 1 else "wandb_all_log_all"
+    valid_modes = [
```
allenwang28: I'm not convinced that we need configurable backends. We should bias towards simplicity, especially in this first PR and discover what our actual needs are. Can we downscope it to only assume one backend, whatever we think is the most reasonable?

apps/toy_metrics/main.py
```diff
@@ -0,0 +1,118 @@
...
+    }
+    service_config = {"procs": 2, "num_replicas": 2, "with_gpus": False}
+
+    # Spawn services first (triggers registrations via provisioner hook)
+    trainer = await TrainActor.options(**service_config).as_service()
+    generator = await GeneratorActor.options(**service_config).as_service()
+
+    # Now init config on global (inits backends eagerly across fetchers)
+    global_logger = await get_or_spawn_controller("global_logger", GlobalLoggingActor)
+    await global_logger.initialize_backends.call_one(logging_config)
```
allenwang28: I'd prefer all the global logger APIs are hidden, i.e.:

```
# mlogger.py

_global_logger = None

async def initialize():
    global _global_logger
    if _global_logger is None:
        _global_logger = get_or_spawn_controller(...)
    _global_logger.initialize_backends.call_one()


async def flush(step):
    global _global_logger
    if _global_logger is None:
        _global_logger = get_or_spawn_controller(...)
    await _global_logger.flush.call_one(step)
```

src/forge/controller/provisioner.py
allenwang28: nit, it isn't when a service is spawned, it's whenever any process is created (that also happens in the service spawning)

src/forge/controller/provisioner.py
```diff
@@ -215,11 +244,24 @@ def bootstrap(gpu_ids: list[str]):
...
         """Stops a proc mesh."""
         async with self._lock:
+            # Deregister local logger from global logger
+            if hasattr(proc_mesh, "_local_fetcher"):
+                from forge.observability.metric_actors import GlobalLoggingActor
+
+                global_logger = await get_or_spawn_controller(
+                    "global_logger", GlobalLoggingActor
+                )
+                process_name = f"proc_mesh_{id(proc_mesh)}"
```
allenwang28: I think you should be able to just pass the `proc_mesh` here instead?

src/forge/observability/metric_actors.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        self.config = config
+
+        # Validate unique classes
+        classes = [b["class"] for b in config["backends"]]
+        if len(set(classes)) != len(classes):
+            raise ValueError("Duplicate logger_backend classes in config")
+
+        # Init global logger_backends and states where needed
+        from forge.observability.metrics import get_logger_backend_class
```
allenwang28: why do we need a lazy import like this?

apps/toy_metrics/main.py
allenwang28: I would also like to avoid having to send another config for logging, if possible...

src/forge/observability/metric_actors.py
```diff
@@ -0,0 +1,208 @@
...
+            from forge.observability.metrics import reduce_metrics_states
+
+            reduced_metrics = reduce_metrics_states(all_local_states)
+
+            # Log to each global logger_backend
+            for (
+                logger_backend_name,
+                logger_backend,
+            ) in self.global_logger_backends.items():
+                await logger_backend.log(reduced_metrics, step)
```
allenwang28, line 316: as a follow up to the backends discussion, can we, for now, get rid of the backends abstraction and explicitly do the wandb.log(...) here?

apps/toy_metrics/main.py
```diff
@@ -0,0 +1,91 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
allenwang28, line 1: for later - let's move this into an integration test (make things assertable) and remove this folder

src/forge/observability/metric_actors.py
```diff
@@ -0,0 +1,335 @@
...
+    MetricCollector,
+    reduce_metrics_states,
+)
+
+logger = logging.getLogger(__name__)
+
+_global_logger = None
+
+
+async def setup_metric_logger(
```
allenwang28, line 24: nit - should be renamed to something like `get_or_create_metric_logger` which is more representative to what this is doing

src/forge/observability/metric_actors.py
```diff
@@ -0,0 +1,335 @@
...
+    def __init__(self, global_logger: Optional["GlobalLoggingActor"] = None) -> None:
+        self.global_logger = global_logger
+        _is_initialized = False
+
+    @endpoint
+    async def flush(
+        self, step: int, return_state: bool = False
+    ) -> Dict[str, Dict[str, Any]]:
+        """Log to local logger backends (if any), reset accumulators and return metric states dict if return_state=True.
+
```
allenwang28, line 126: nit: add a comment that says this should only ever be called by the global logger

Issue #173, opened 2025-09-17
[RFC] Add `MyActor.as_actor()` API


allenwang28:
This RFC proposes two changes:
- Makes what are currently singleton services as actors
- APIs on how to spawn actors

# Services => Actors
Everything in Forge is currently a service, but maybe this should only be restricted to:
- vLLM generators
- any environments, i.e. coding env etc.

Pros:
- More accurate conceptually - singletons as services is overkill

Cons:
- Introduces more concepts (user has to reason about the fact that there are actors, and there are services which are composed of multiple actors)

# How to spawn Actors
Currently, services are spawned as

```
service = await MyActor.options(replicas=1, procs_per_replica=4, hosts_per_replica=2).as_service()
```

To maintain consistency, we could spawn singletons as:
```
actor = await MyActor.options(procs=4, hosts=2).as_actor()
```

Pros:
- Consistency with how we're reasoning about services

Cons:
- Hides Monarch APIs (i.e. hides away proc_meshes/etc.)

If we like this, then I'd also propose that services drop the `_per_replica` suffix, so it would then become:

```
service = await MyActor.options(replicas=1, procs=4, hosts=2).as_service()
```

allenwang28:
cc @vidhyav @Ritesh1905  @JenniferWang @Jack-Khuu @pbontrager @joecummings 

some of whom I've had a related conversation with :)

Jack-Khuu:
Everything about this looks like a strict improvment to me. 

One option that comes to mind is to push this even further: Do non-power users need to interact with the concept of service after construction? (lmk if I'm barking up the same tree again)

``` python
actor1 = await MyActor.options(replicas=1, procs=4, hosts_per_replica=2).spawn() # Service under the hood
actor2 = await MyActor.options(replicas=1, procs=4).spawn() 

actor2.entrypoint.choose() === actor2.entrypoint.call()

actor2.<"service specific func">() # Return singleton results (count) or error out 
```

> Hides Monarch APIs (i.e. hides away proc_meshes/etc.)

I don't know if I would call this a con for casuals 😉

allenwang28:
I like the simplicity of your proposal, but the concern I have is that the API surface and how you interact with the resulting object changes whether or not you're running a service.

vidhyav:
This is great!

Another way to look at it is as follows: Service endpoints are good for the following use cases: 

[...]

PR #195, opened 2025-09-19
Add `as_actor` single-actor mode

DNXie:
**Summary:**
This PR adds `ForgeActor.as_actor()` and refactors `ForgeActor.options()` and `.as_service()` to improve configuration handling and dynamic subclassing for actors and services. Context: https://github.com/meta-pytorch/forge/issues/173

Key changes include:

* Added `as_actor()` to support launching a single actor directly.
* Support positional argument for `Service` and `Actor`
* Rename `num_hosts` to `hosts` and `num_procs` to `procs` in `ProcessConfig`.
* `.options()` now stores all configuration parameters as class attributes rather than building a full config object immediately.
* Dynamic subclasses are created only during `.as_actor()` or `.as_service()` calls, ensuring each configuration remains isolated.
* Default configuration is applied automatically if `.options()` is not called.

**Changes in behavior:**

So the single actor initialization from 
```
cfg = ProcessConfig(...)
actor = await MyForgeActor.launch(process_config=cfg, **actor_kwargs)
```
become 
``` 
actor = await MyForgeActor.options(procs=1, ...).as_actor(**actor_kwargs)
```

**Usage Examples 1 (Actor):**
```python
# Pre-configure a single actor
actor = await MyForgeActor.options(procs=1, hosts=1).as_actor(...)
await actor.shutdown()

# Default usage without calling options
actor = await MyForgeActor.as_actor(...)
await actor.shutdown()
```

Log:
```bash
Spawning single actor Counter
```

**Usage Examples 2 (Service):**

```python
# Pre-configure a service with multiple replicas
service = await MyForgeActor.options(num_replicas=3, procs=2).as_service(...)
await service.shutdown()

# Default usage without calling options
service = await MyForgeActor.as_service(...)
await service.shutdown()
```
Log when `num_replicas=3`

The printed class name is its original class name instead of xxService (See https://github.com/meta-pytorch/forge/pull/193)
```bash
```

**Usage Examples 3 (Positional argument):**
This means you can now do:

```python
await Counter.as_service(10)
```

instead of having to use keyword-only arguments like:

```python
await Counter.as_service(v=10)
```

**Test**
```
```

### DIFF
--- a/tests/unit_tests/test_service.py
+++ b/tests/unit_tests/test_service.py
@@ -69,6 +69,7 @@ def make_replica(idx: int, healthy: bool = True, load: int = 0) -> Replica:
         idx=idx,
         proc_config=ProcessConfig(),
         actor_def=Counter,
+        actor_args=(),
         actor_kwargs={},
     )
     replica.state = ReplicaState.HEALTHY if healthy else ReplicaState.UNHEALTHY


Ritesh1905:
Wondering what is the motivation behind this? Why would one choose an actor creation directly over service? Single actor seems to be a special case of service?

allenwang28:
> Wondering what is the motivation behind this? Why would one choose an actor creation directly over service? Single actor seems to be a special case of service?

For context: https://github.com/meta-pytorch/forge/issues/173

You're right though, the rationale is that only vLLM should be a service currently. Trainer for e.g. will not really take advantage of fault tolerance or routing, so we should always expect it to be a singleton.

src/forge/controller/actor.py
```diff
@@ -88,11 +125,11 @@ def options(
                 **service_kwargs,
             )
 
-        return type(
-            f"{cls.__name__}Service",
-            (cls,),
-            {"_service_config": cfg},
-        )
+            return type(
+                f"{cls.__name__}Service",
```
allenwang28: @DNXie I still don't understand why we can't return just `cls.__name__` here, please let me know if I'm missing something. 

If I do something like this:
```
class Test:
    _option = None

    @classmethod
    def options(cls, option):
        return type(
            cls.__name__,
            (cls,),
            {"_option": option},
        )

t = Test
print(t)
print(t._option)


t2 = Test.options(option="hi")
print(t2)
print(t2._option)
```

I see 
```
<class '__main__.Test'>
None
<class '__main__.Test'>
hi
```

Can't we do something like here too?

src/forge/controller/actor.py
```diff
@@ -57,7 +58,11 @@ def options(
         Returns a subclass of this ForgeActor with a bound ServiceConfig.
         The returned subclass can later be launched via `.as_service()`.
 
-        Usage (choose ONE of the following forms):
+        Usage modes:
+
+        ---- Service Mode (default) ----
```
allenwang28: since the user specifies the service or actor mode by `as_service()` or `as_actor()` I'm not sure there's truly a default

src/forge/controller/actor.py
```diff
@@ -57,7 +58,11 @@ def options(
...
+        Use when deploying a replicated service (multiple replicas, each with N procs).
+
             # Option A: construct ServiceConfig implicitly
```
allenwang28: In a follow up, we should remove `ServiceConfig` and `ProcessConfig` altogether

src/forge/controller/actor.py
```diff
@@ -73,10 +78,42 @@ def options(
...
+            raise ValueError(
+                "Cannot pass both `service_config` and `process_config`. "
+                "Use either `service_config` for service mode or `process_config` for single actor mode."
+            )
+
+        if process_config is not None:
+            return type(
+                f"{cls.__name__}Actor",
+                (cls,),
+                {"_process_config": process_config},
```
allenwang28: we should have something like this:

```
class ForgeActor(Actor):
    _procs: int = 1
    _replicas: int = 1
    _hosts: int = 0
    _with_gpu: bool = False
    
    @classmethod
    def options(cls, ...):
        return type(
            cls.__name__,
            (cls,),
            {"_replicas": replicas, "_procs": procs, "_hosts": hosts, "_with_gpus": with_gpus}
        )
```

and then the service or actor spawning doesn't need the process config or service config anymore!

DNXie: Done

src/forge/controller/actor.py
DNXie: Removed

src/forge/controller/actor.py
DNXie: Looks like we can! Updated. Please check the latest version.

src/forge/controller/actor.py
```diff
@@ -48,73 +59,64 @@ def __init__(self, *args, **kwargs):
     def options(
         cls: Type[T],
         *,
-        service_config: ServiceConfig | None = None,
-        num_replicas: int | None = None,
-        procs: int | None = None,
-        **service_kwargs,
+        procs: int,
```
allenwang28: also `hosts: int`, `with_gpu: bool` and `num_replicas: int | None`?

src/forge/controller/actor.py
```diff
@@ -48,73 +59,64 @@ def __init__(self, *args, **kwargs):
...
-            # dynamically create a configured subclass for consistency
-            cls = type(f"{cls.__name__}Service", (cls,), {"_service_config": cfg})
+        class_attrs = {k: v for k, v in cls.__dict__.items() if not k.startswith("__")}
+        if "procs" not in class_attrs:
+            class_attrs["procs"] = 1
+        if "num_replicas" not in class_attrs:
+            class_attrs["num_replicas"] = 1
+        cfg = ServiceConfig(**filter_config_params(ServiceConfig, class_attrs))
+
+        service_cls = type(f"{cls.__name__}Service", (cls,), {"_service_config": cfg})
```
allenwang28: Why do we still need `service_cls` here? can the logic of `as_service()` be:

```
@classmethod
async def as_service(cls, **actor_kwargs) -> "ServiceInterface":
    service = Service(cfg, cls, actor_kwargs)
    await service.__initialize__()
    return ServiceInterface(service, cls)
```

src/forge/controller/actor.py
```diff
@@ -181,11 +183,29 @@ async def launch(cls, *, process_config: ProcessConfig, **kwargs) -> "ForgeActor
...
+        The configuration values stored in the subclass returned by `.options()` (like
+        `procs`) are used to construct a ProcessConfig instance.
+        If no configuration was stored, defaults to a single process with no GPU.
         """
-        if actor._proc_mesh is None:
+        class_attrs = {k: v for k, v in cls.__dict__.items() if not k.startswith("__")}
+        cfg = ProcessConfig(**filter_config_params(ProcessConfig, class_attrs))
+
+        logger.info("Spawning single actor %s", cls.__name__)
+        actor = await cls.launch(process_config=cfg, **actor_kwargs)
```
allenwang28: hmm maybe we can modify the `def launch()` above to simplify things?

Like this:
```
@classmethod
async def launch(cls, *args, **kwargs) -> "ForgeActor":
    proc_mesh = await get_proc_mesh(process_config=ProcessConfig(procs=cls._procs, hosts=cls._hosts, with_gpu=cls._with_gpu))
    actor_name = kwargs.pop("name", cls.__name__)
    actor = await proc_mesh.spawn(actor_name, cls, *args, **kwargs)
    actor._proc_mesh = proc_mesh
    if hasattr(proc_mesh, "_hostname") and hasattr(proc_mesh, "_port"):
        host, port = proc_mesh._hostname, proc_mesh._port
        await actor.set_env.call(addr=host, port=port)
    await actor.setup.call()
    return actor
    
```

src/forge/controller/actor.py
DNXie: I kind of put them all in `**kwargs` since only `procs` is required for both service and actor. Do you think it is better to explicitly list them?

src/forge/controller/actor.py
DNXie: You are right! Removed.

src/forge/controller/actor.py
DNXie: Done!

src/forge/controller/actor.py
allenwang28: yes, please explicitly list them

src/forge/controller/actor.py
```diff
@@ -48,68 +59,57 @@ def __init__(self, *args, **kwargs):
...
         from forge.controller.service import Service, ServiceInterface
 
-        # Use _service_config if already set by options(), else default
-        cfg = getattr(cls, "_service_config", None)
-        if cfg is None:
-            cfg = ServiceConfig(num_replicas=1, procs=1)
-            # dynamically create a configured subclass for consistency
-            cls = type(f"{cls.__name__}Service", (cls,), {"_service_config": cfg})
+        class_attrs = {k: v for k, v in cls.__dict__.items() if not k.startswith("__")}
+        if "procs" not in class_attrs:
```
allenwang28: follow up comment on explicit attributes, this for e.g. is unclear and can be pretty brittle

src/forge/controller/actor.py
```diff
@@ -167,7 +167,13 @@ async def launch(cls, *, process_config: ProcessConfig, **kwargs) -> "ForgeActor
         a homogeneous set of actors on a single proc mesh.
 
         """
-        proc_mesh = await get_proc_mesh(process_config=process_config)
+        # Build process config from class attributes with defaults
+        cfg = ProcessConfig(
+            procs=getattr(cls, "procs", 1),
```
allenwang28: I generally try and use `getattr` as little as possible. If it's used too much it can mask real errors that can be really hard to debug later.

src/forge/controller/actor.py
```diff
@@ -181,11 +187,26 @@ async def launch(cls, *, process_config: ProcessConfig, **kwargs) -> "ForgeActor
...
+        The configuration values stored in the subclass returned by `.options()` (like
+        `procs`) are used to construct a ProcessConfig instance.
+        If no configuration was stored, defaults to a single process with no GPU.
         """
-        if actor._proc_mesh is None:
+        logger.info("Spawning single actor %s", cls.__name__)
+        actor = await cls.launch(**actor_kwargs)
+
+        # Patch shutdown to bypass endpoint system
+        actor.shutdown = types.MethodType(
```
allenwang28: hmm this is a hack, we shouldn't be doing this. I'm guessing it's because we want to preserve the ability to 

```
svc = MyActor.as_service()

await svc.shutdown()
```

?

DNXie: No, `as_service` returns a `ServiceInterface`. So when we call `service.shutdown()`, we are actually calling [ServiceInterface.shutdown](https://github.com/meta-pytorch/forge/blob/main/src/forge/controller/service/interface.py#L173)

The reason I have to do this hacky thing is: 
Without it, `actor.shutdown()` gives me this error:
```bash
RuntimeError: Actor <class 'tests.unit_tests.test_service.Counter'>.shutdown is not annotated as an endpoint. To call it as one, add a @endpoint decorator to it, or directly wrap it in one as_endpoint(obj.method).call(...)
```

If I simply decorate `shutdown` with `@endpoint`, we'd have to call it like 
```
await actor.shutdown.call()
```
But it would still give error:
```bash
AssertionError("Called shutdown on a replica with no proc_mesh.")
```

Any suggestions?

src/forge/controller/actor.py
DNXie: This is a fallback when the user doesn’t specify configs via `.options()`. In this case, the original `ForgeActor` class doesn’t have attributes like `procs`. If we are getting rid of `getattr`, one way I can think of is to add these attributes to `ForgeActor` class like 
```python
class ForgeActor(Actor):
    procs: int = 1
    hosts: int | None = None
    with_gpus: bool = False
    num_replicas: int = 1

    def __init__(self, *args, **kwargs):
```

But either way,  it means the default values are specified in three places:

1. In [types.py](https://github.com/meta-pytorch/forge/blob/12234732207ed7abc9a3604922487c108fe027dc/src/forge/types.py#L94-L96)
3. As default values in `.options()`
4. As attributes on the `ForgeActor` class OR here in `launch`.

I’m not sure if there’s a cleaner way to handle this. I’ve updated the code accordingly (get rid of `getattr`), please take a look and let me know if you have any suggestions or improvements.

src/forge/controller/actor.py
DNXie: Removed in the latest version

src/forge/controller/actor.py
allenwang28: ah I see. Ok in that case, I think what we should do is not do `actor.shutdown()` for now, and just rely on eg

```
await RLTrainer.stop(trainer)
```

for now. Maybe what we can do next is have the provisioner keep track of all of the proc meshes, and do a global `shutdown()`? Including all the services etc. we can discuss more, just want to unblock this PR

DNXie: Sounds good. Done!

src/forge/controller/actor.py
```diff
@@ -154,7 +162,7 @@ async def set_env(self, addr: str, port: str):
         os.environ["MASTER_PORT"] = port
 
     @classmethod
-    async def launch(cls, *, process_config: ProcessConfig, **kwargs) -> "ForgeActor":
+    async def launch(cls, **kwargs) -> "ForgeActor":
```
allenwang28: can you add *args here? This solves the *args related TODO that's listed here!

src/forge/controller/actor.py
```diff
@@ -48,68 +54,69 @@ def __init__(self, *args, **kwargs):
...
-
-            # Option B: provide an explicit ServiceConfig
-            cfg = ServiceConfig(num_replicas=1, procs=2, ..)
-            service = await MyForgeActor.options(service_config=cfg).as_service(...)
-            await service.shutdown()
-
-            # Option C: skip options, use the default service config with num_replicas=1, procs=1
-            service = await MyForgeActor.as_service(...)
-            await service.shutdown()
+        Returns a dynamically created subclass of this ForgeActor with bound configuration.
```
allenwang28:
```suggestion
        Returns a version of ForgeActor with configured resource attributes.
```

src/forge/controller/actor.py
DNXie: Added in `launch` and `as_actor`. Also tested in `test_as_actor_with_kwargs_config`

tests/unit_tests/test_service.py
```diff
@@ -31,8 +31,18 @@
...
     """Test actor that maintains a counter with various endpoints."""
 
-    def __init__(self, v: int):
+    def __init__(self, v: int, *args, **kwargs):
         self.v = v
+        self.args = args
+        self.kwargs = kwargs
+
+    @endpoint
+    async def get_args(self):
```
allenwang28: let's remove these tests, i think it's fine without

Issue #190, opened 2025-09-19
In torchtitan's trainer, `_qwen3_hf_to_vllm` `num_layers` is hardcoded to 28


allenwang28:
See [here](https://github.com/meta-pytorch/forge/blob/d55de5b8da5a1c1d34934a380649aea940db214d/src/forge/actors/trainer.py#L218)

This works for 1.8B but fails relatively silently for other models. Obv this should be deduced from the model name somehow

PR #199, opened 2025-09-19
dynamically determine num_layers in _qwen3_hf_to_vllm

casteryh:
This is fixing #190

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -6,6 +6,7 @@
 
 import math
 import os
+import re
 import time
 from collections.abc import Mapping
 from dataclasses import dataclass, field, fields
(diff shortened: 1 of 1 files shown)


src/forge/actors/trainer.py
```diff
@@ -215,7 +216,7 @@ async def push_weights(self, policy_version: int) -> None:
             )
         hf_state_dict = self.engine.checkpointer.sd_adapter.to_hf(flattened_state_dict)
         # TODO: Figure out how to gracefully handle which model to-vLLM conversion is needed
-        vllm_ready_hf_sd = _qwen3_hf_to_vllm(sd=hf_state_dict, num_layers=28)
+        vllm_ready_hf_sd = _qwen3_hf_to_vllm(sd=hf_state_dict)
```
Ritesh1905, line 219: Could this just be simplified by using `num_layers=self.model.config.num_hidden_layers`? see [this](https://github.com/meta-pytorch/forge/blob/main/apps/toy_rl/sumdigits.py#L344-L346)

Ideally you should not be needing this method in the trainer at all. The trainer should be agnostic to the type/arch of generator.

allenwang28: this should work with torchtitan:

```
num_layers=self.engine.model_args.n_layers)
```

casteryh: I guess simply reading the max num layers from regex has the advantage that it's agnostic of the trainer implementation as long as the state dict is in huggingface format.
Let me know what you think. @Ritesh1905 @allenwang28

casteryh:
closed since the change is already in #215

PR #202, opened 2025-09-19
Creates Judge Example as a wrapper on Policy

Jack-Khuu:
_Update(10/10) : This is a very much simplified version of the previous iterations of the PRs. It provides just a basic example of LLM Judges for GRPO_

**Judges** can both be used as "Verifiers" or "Graders". This PR adds to the sandbox, a `CorrectnessJudge` example of how an LLM Judge can be used in GRPO (note that this PR does not integrate)

It takes as input (prompt + response) generated from a model, and returns whether the model thinks it accurately responded to the prompt. Results can then be used to make decisions during GRPO whitening

---
```
python -m tests.sandbox.vllm.judge --config tests/sandbox/vllm/qwen3_4b.yaml
```
image


apps/vllm/judge.py
```diff
@@ -0,0 +1,108 @@
...
+        policy,
+        prompt_wrapper=basic_selector_wrapper(cfg.policy.engine_config.model),
+        output_postprocessor=unroll_response,
+    )
+
+    print(f"Prompt: {prompt}")
+    print(f"Responses: {responses}\n")
+
+    try:
+        async with policy.session():
```
Ritesh1905: why do you need this?

src/forge/actors/generative_judge.py
```diff
@@ -0,0 +1,52 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Callable, Optional
+
+
+@dataclass
+class GenerativeJudge:
```
Ritesh1905: This does not seem to be an actor? does not inherit `ForgeActor?`

apps/vllm/judge.py
```diff
@@ -0,0 +1,108 @@
...
+    return [output.text for output in response.outputs]
+
+
+async def run(cfg: DictConfig):
+    prompt = "What is the capital of Japan?"
+    responses = ["Aardvark", "Durian", "Tokyo"]
+
+    print("Spawning service...")
+    policy = await Policy.options(**cfg.services.policy).as_service(**cfg.policy)
+    evaluate = GenerativeJudge(
```
Ritesh1905: Could you help me understand why can't we actor'ify the generative judge? policy/service config can be passed to the GenerativeJudgeActor, and let it figure put the the actor creator, session semantics (if that is needed) etc?

apps/vllm/judge.py
```diff
@@ -0,0 +1,108 @@
...
+
+from vllm.transformers_utils.tokenizer import get_tokenizer
+
+
+def basic_selector_wrapper(model: str) -> Callable[[str, list[str]], str]:
+    """
+    Note: This is not a "good" prompt setup, it just demonstrates how to make one
+    """
+
+    def _wrapper(prompt: str, responses: list[str]) -> str:
```
allenwang28: hmm, this is a good start and I think we can improve as well. IIUC in LLMs as verifiers we have two tracks:
1. Reward models
2. LLM as a judge

what seems to differ is the prompt you input and whether or not you have to massage the outputs? Is there a way we can minimize the user code to focus on just that?

apps/vllm/judge.py
allenwang28: I think Jack's intention here is that the generators can also be used as the judge (e.g. how it's done here)

but intuitively I feel like they should be kept separate. I feel like a judge specific actor and a reward model specific actor makes sense too, with all of the boilerplate kept with those implementations

apps/vllm/judge.py
Jack-Khuu: We don't, will remove

apps/vllm/judge.py
Jack-Khuu: > Is there a way we can minimize the user code to focus on just that?

Agreed that the wrapper seems funky since i wanted to test generic models. In practice a user would just pass in a `(str, list[str]) -> str` to the constructor

We can bake in a default of "LLM as a judge" `Judge(model_name, policy_config)` and reduce the scope of class?

Jack-Khuu:
> lol is this working correctly?

I wrote this prompt from the deep archives of my mind and I'm also shocked that the prompting worked.

src/forge/actors/generative_judge.py
Jack-Khuu: Will discuss in the other comment

apps/vllm/judge.py
Jack-Khuu: > Could you help me understand why can't we actor'ify the generative judge? 
> generators can also be used as the judge

The idea behind taking in a hydrated generator was to make it easy to enable uses of an existing policy (or policy version) as a discriminator. 

That said we absolutely can make this an actor and push the setup inside. I wanted to avoid some of the boiler plate, but if we're fine with it, I'll send up a JudgeActor, RewardModelActor

allenwang28: > enable uses of an existing policy (or policy version) as a discriminator.
Is this something that's common in the literature? If it's not I sort of would wanna wait and see until this is requested

>  I'll send up a JudgeActor, RewardModelActor
I think `JudgeActor` and `RewardModelActor` are reasonable. I also think it's reasonable if you want to do another PR now that renames Policy to VLLM / VLLMWorker etc. as a PR before this

src/forge/actors/generative_judge.py
```diff
@@ -0,0 +1,172 @@
...
+            **policy_cfg.policy
+        )
+
+        actor_name = kwargs.pop("name", cls.__name__)
+        llm_judge = await judge_procs.spawn(
+            actor_name,
+            cls,
+            prompt_wrapper=prompt_wrapper,
+            output_postprocessor=output_postprocessor,
+            generator=policy,
```
Jack-Khuu: Running into a pickling issue with passing around ServiceInterfaces

```python
  File "/home/jackkhuu/forge/src/forge/actors/generative_judge.py", line 51, in launch
    llm_judge = await judge_procs.spawn(
  File "/home/jackkhuu/.fbpkg_conda_envs/forge-a7401c7/lib/python3.10/site-packages/monarch/_src/actor/proc_mesh.py", line 254, in spawn
    return self._spawn_nonblocking(name, Class, *args, **kwargs)
  File "/home/jackkhuu/.fbpkg_conda_envs/forge-a7401c7/lib/python3.10/site-packages/monarch/_src/actor/proc_mesh.py", line 366, in _spawn_nonblocking
    return self._spawn_nonblocking_on(self._proc_mesh, name, Class, *args, **kwargs)
  File "/home/jackkhuu/.fbpkg_conda_envs/forge-a7401c7/lib/python3.10/site-packages/monarch/_src/actor/proc_mesh.py", line 386, in _spawn_nonblocking_on
    service = ActorMesh._create(
  File "/home/jackkhuu/.fbpkg_conda_envs/forge-a7401c7/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 1048, in _create
    send(ep, (mesh._class, proc_mesh, controller_controller, *args), kwargs)
  File "/home/jackkhuu/.fbpkg_conda_envs/forge-a7401c7/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 603, in send
    endpoint._send(args, kwargs, port, selection)
  File "/home/jackkhuu/.fbpkg_conda_envs/forge-a7401c7/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 465, in _send
    objects, bytes = flatten((args, kwargs), _is_ref_or_mailbox)
  File "/home/jackkhuu/.fbpkg_conda_envs/forge-a7401c7/lib/python3.10/site-packages/monarch/_src/actor/pickle.py", line 73, in flatten
    pickler.dump(obj)
  File "/home/jackkhuu/.fbpkg_conda_envs/forge-a7401c7/lib/python3.10/site-packages/cloudpickle/cloudpickle.py", line 1303, in dump
    return super().dump(obj)
TypeError: cannot pickle '_asyncio.Future' object
```

src/forge/actors/judge.py
```diff
@@ -0,0 +1,214 @@
...
+        return [output.text for output in outputs]
+
+    @endpoint
+    async def evaluate(
+        self,
+        prompt: str,
+        responses: list[str],
+    ) -> list[str]:
+        wrapped_prompt: str = self._wrap_prompt(prompt, responses)
+        response: List[Completion] = await self.generate._method(self, wrapped_prompt)
```
allenwang28: should we just change the Policy to have _generate with the core logic and a generate endpoint?

src/forge/actors/judge.py
```diff
@@ -0,0 +1,214 @@
...
+    `LLM-based Judges` are typically generative models which are then prompted
+    to evaluate responses. These models NEED prompt engineering to evaluate
+    and may require more postprocessing
+    """
+
+    def _math_check(
+        self,
+        prompt: str,
+        responses: list[str],
+        ground_truth: None | str = None,
```
allenwang28: why does the judge need access to the ground truth for math check?

src/forge/actors/judge.py
Jack-Khuu: Do Service/Replica keep the non endpoint attributes?

No strong pref, if it does

src/forge/actors/judge.py
Jack-Khuu: If you're using the judge as a grader the ground truth is convenient 
(technically anything can be baked into the prompt, but it'd then be different than the one we provide the model)

src/forge/actors/judge.py
allenwang28: > Do Service/Replica keep the non endpoint attributes?

yes, it'll probably be cleaner than accessing the _method field which could change at any time

src/forge/actors/judge.py
joecummings, line 1: General nit voicing a preference for injection vs. inheritance.

Jack-Khuu: Oh boi this is a fun one. Had a quick chat with @ebsmothers on me opening the flood gates few days ago
cc: @pbontrager too

* **Inheritance** is by far the simplest implementation without leaving the Actor world: This version
* Second Option is also fairly clean by **passing in a hydrated Generation Service** (i.e. Policy) and having Judges be a non-Actor wrapper using it: [iteration 1](https://github.com/meta-pytorch/forge/pull/202/commits/249a6ab7681ff160ee8aedcf8228336a549dbb4f)
  * Potential issue with this one is whether we want to get into the habit of passing around Services as glorified lambda wrappers matching an interface
* Third Option is having **Judges be actors, which internally constructs the Generation** Service: [iteration 2](https://github.com/meta-pytorch/forge/pull/202/commits/e88f58f987a59dc6755ad0bd31611ac9bc00d6f2#diff-4544c229da1b2dafe6fe36604362fe4eb46d2e97030e4a99a3d178ed0c505bf5) 
  * This is the messiest one and enters the territory of nested Services (Judge has a Policy which has PolicyWorkers)

Jack-Khuu: Oh boi this is a fun one. Had a quick chat with @ebsmothers on me opening the flood gates few days ago
cc: @pbontrager too

* **Inheritance** is by far the simplest implementation without leaving the Actor world: This version
* Second Option is also fairly clean by **passing in a hydrated Generation Service** (i.e. Policy) and having Judges be a non-Actor wrapper using it: [iteration 1](https://github.com/meta-pytorch/forge/pull/202/commits/249a6ab7681ff160ee8aedcf8228336a549dbb4f)
  * Potential issue with this one is whether we want to get into the habit of passing around Services as glorified lambda wrappers matching an interface
* Third Option is having **Judges be actors, which internally constructs the Generation** Service: [iteration 2](https://github.com/meta-pytorch/forge/pull/202/commits/e88f58f987a59dc6755ad0bd31611ac9bc00d6f2#diff-4544c229da1b2dafe6fe36604362fe4eb46d2e97030e4a99a3d178ed0c505bf5) 
  * This is the messiest one and enters the territory of nested Services (Judge has a Policy which has PolicyWorkers)

allenwang28: another option is like

```
judge = await Generator.as_service("Qwen/Qwen3-32B", judge_prompt="", ...)
reward_model = await Generator.as_service("Skywork/Skywork-Reward-V2-Llama-3.1-8B", is_reward_model=True)
```

and have everything be kept within Generator? 

imo out of all of the options we've seen, inheritance is the least of all evils. But we should do our best to keep the level of inheritance low, like we certainly shouldn't add in another layer on top of this

Jack-Khuu:
Update (10/10): This is a very much simplified version of the previous iterations of the PRs. It provides just a basic example of LLM Judges for GRPO

There is out of scope future work leveraging structured decoding, but that requires additional investigation on how to configure with CoT models (which need to think)

PR #163, opened 2025-09-16
[icebox][ez-ish] Collect compute_logprobs into util

Jack-Khuu:
Currently `compute_logprobs` is repeated in 3 different locations (2 of which are literally identical). This just puts them all in one place under `util/compute_logprobs.py`

Locations
* `apps/grpo/main.py`
* `apps/rl/main.py` --- Identical to grpo's
* `src/forge/actors/reference_model.py`

---

Note that the third definition in `reference_model` is functionally identical to that of the main.py files

Main.py implementations
``` python
def compute_logprobs(
    logits: Tensor, input_ids: torch.Tensor, temperature: float = 1.0
) -> Tensor:
    context_length = logits.shape[1] - input_ids.shape[1]

    # Truncate request logits and drop last
    logits = logits[:, context_length - 1 : -1]

    # Compute logprobs
    logprobs = torch.log_softmax(logits / temperature, dim=-1)
    logprobs = torch.gather(logprobs, 2, input_ids.unsqueeze(-1)).squeeze(-1)

    return logprobs
```

Util implementation (With assert checks redacted)
``` python
def compute_logprobs(
    logits: torch.Tensor, input_ids: torch.Tensor, temperature: float = 1.0
) -> torch.Tensor:
    context_len = logits.shape[1] - input_ids.shape[1]

    # (bsz, completion_len, vocab_size)
    logits = logits[:, context_len - 1 : -1, :].         # Difference here is a no-op all-slice {:}
    
    token_logprobs = torch.log_softmax(logits / temperature, dim=-1)
    # (bsz, completion_len, 1)
    logprobs = torch.gather(token_logprobs, 2, input_ids.unsqueeze(-1))    # Difference is that it got split to 2 lines
    # (bsz, completion_len)
    logprobs = logprobs.squeeze(-1)

    return logprobs
```

PR #204, opened 2025-09-22
Minor refactor for algorithms in the GRPO app.

JenniferWang:
This diff moves purely mathematical computations to a separate module `algorithms.py`
Specifically remove the advantage computation away from `ForgeActor` since this is a purely functional.

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -281,8 +281,9 @@ async def main(cfg: DictConfig):
             del ref_logits, ref_logprobs, input_ids
 
             # Calculate advantages and add to replay buffer
-            rewards = [episode.reward for episode in group.episodes]
-            advantages = compute_advantages(rewards)
+            advantages = compute_advantages(
+                [episode.reward for episode in group.episodes]
+            )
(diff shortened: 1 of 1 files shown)


JenniferWang:
I think totally understand the desire to keep logics in one file for faster lookup but this makes unit testing functions very heavy lifted -- in the unit test, it will need to import Monarch stuff if the functionality lives in the main app.

As for the `compute_advantages` -- `compute_logprobs` looks way more heavier than `compute_advantages`. Can we talk about optimization when we have more instrumation? Premature and inconsistent optimization confuses the reader.

@joecummings , @allenwang28

Jack-Khuu:
dejavu: https://github.com/meta-pytorch/forge/pull/163

We can mirror the decision we make for @204 (this PR) over there #163

JenniferWang:
Ah I see. I do think this PR is different in the sense that it is only scoped to the GRPO app folder for easier unit testing. 

> dejavu: #163
> 
> We can mirror the decision we make for @204 (this PR) over there #163

PR #207, opened 2025-09-22
Policy uses the generic completion data model

Ritesh1905:
- Policy uses the generic completion data model
- Ran all the apps using the vLLm specific output

### DIFF
--- a/apps/toy_rl/sumdigits.py
+++ b/apps/toy_rl/sumdigits.py
@@ -490,7 +490,7 @@ async def main(cfg: DictConfig):
                 episode.request_tokens = response.prompt_ids
                 episode.response_tokens = response.token_ids
                 episode.response = response.text
-                episode.response_logprobs = response.log_probs
+                episode.response_logprobs = response.logprobs
                 episode.ref_logprobs = await ref_model.forward.choose(episode)
                 episode.reward = await reward_actor.evaluate_response.choose(
                     prompt=prompt, response=response.text, target=target
--- a/apps/vllm/main.py
+++ b/apps/vllm/main.py
@@ -17,8 +17,8 @@ from forge.actors.policy import Policy
 from forge.cli.config import parse
 from forge.controller.provisioner import shutdown
 
+from forge.data_models.completion import Completion
 from omegaconf import DictConfig
-from vllm.outputs import RequestOutput
 
 os.environ["HYPERACTOR_MESSAGE_DELIVERY_TIMEOUT_SECS"] = "600"
 os.environ["HYPERACTOR_CODE_MAX_FRAME_LENGTH"] = "1073741824"
@@ -36,7 +36,9 @@ async def run(cfg: DictConfig):
     try:
         async with policy.session():
             print("Requesting generation...")
-            response_output: RequestOutput = await policy.generate.choose(prompt=prompt)
+            response_output: list[Completion] = await policy.generate.choose(
+                prompt=prompt
+            )
 
             print("\nGeneration Results:")
             print("=" * 80)
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -45,6 +45,7 @@ from forge.controller import ForgeActor, get_proc_mesh, stop_proc_mesh
 from forge.data.sharding import VLLMSharding
 from forge.data_models.completion import Completion
 from forge.data_models.prompt import to_prompt
+
 from forge.interfaces import Policy as PolicyInterface
 from forge.types import ProcessConfig
 
@@ -363,7 +364,7 @@ class Policy(PolicyInterface):
 
             for request_output in processed_outputs.request_outputs:
                 if request_output.finished:
-                    completions = self._collect_completions(request_output)
+                    completions = self._to_completions(request_output)
                     _, fut = self.requests.pop(request_output.request_id)
                     fut.set_result(completions)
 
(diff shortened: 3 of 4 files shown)


apps/toy_rl/sumdigits.py
```diff
@@ -358,7 +358,7 @@ class RewardActor(ForgeActor):
 
     @endpoint
     async def evaluate_response(self, prompt: str, response: str, target: str) -> float:
-        reward = 1.0 if response.strip() == "10" else 0.0
```
Ritesh1905, line 361: It was a leftover from my local debugging.

src/forge/data_models/completion.py
```diff
@@ -29,3 +30,6 @@ class Completion:
 
     # the log probabilities of the target tokens
     log_probs: Optional[torch.Tensor] = None
```
joecummings: nit: looks like standard notation from OpenAI, SGLang, and vLLM is that log_probs should be one word. I know it's not the main point here, but do you mind fixing it? :)

src/forge/actors/policy.py
```diff
@@ -395,6 +398,54 @@ async def get_version(self) -> int:
     async def stop(self):
         self.running = False
 
+    def _collect_completions(self, request_output: RequestOutput) -> list[Completion]:
```
joecummings: nit: _to_completions

src/forge/actors/policy.py
```diff
@@ -395,6 +398,54 @@ async def get_version(self) -> int:
...
+                        for token, top_k_dict in zip(
+                            output.token_ids,
+                            output.logprobs,
+                        )
+                    ]
+                )
+
+            completions.append(
+                Completion(
+                    # TODO: the to_prompt encoding will be different from the original.
```
joecummings, line 410: Can you expand here? Why will it be different?

src/forge/actors/policy.py
```diff
@@ -395,6 +398,54 @@ async def get_version(self) -> int:
...
+                    text=output.text,
+                    prompt_ids=torch.tensor(prompt_token_ids),
+                    token_ids=torch.tensor(output.token_ids),
+                    log_probs=logprobs,
+                )
+            )
+
+        return completions
+
+    def _extract_logprobs(self, one_sample: CompletionOutput) -> torch.Tensor | None:
```
joecummings, line 423: This doesn't seem to be used?

Jack-Khuu, line 423: Where is this being used?

apps/vllm/main.py
```diff
@@ -40,7 +40,7 @@ async def run(cfg: DictConfig):
 
             print("\nGeneration Results:")
             print("=" * 80)
-            for batch, response in enumerate(response_output.outputs):
+            for batch, response in enumerate(response_output):
```
Jack-Khuu, line 45: Can you update the typehint on line 39?

src/forge/actors/policy.py
allenwang28: does it make sense to put this logic with Completion? like

```
@dataclass
class Completion:
    def from_request_output(cls, request_output: RequestOutput) -> "Completion"
        ...
```

Ritesh1905: Sure. I can move it to completion object.

src/forge/actors/policy.py
Ritesh1905: My bad. This will be used in collect_completions method, Will fix.

src/forge/actors/policy.py
Ritesh1905: So, it depends on the the original prompt sent to the `generate` method. In case of GRPO/Sumdigits, you are applying a chat template where as the vllm uses a simple text prompt. 

Unless we are using a strongly types `Prompt`, there would be a mismatch. Using the prompt would come as a next PR.

src/forge/actors/policy.py
Ritesh1905: My bad. This will be used in collect_completions method, Will fix.

src/forge/actors/policy.py
joecummings: Actually, I'd prefer to keep the logic here. If the goal of these interfaces is to provide a minimal contract for what should work with other forge items and that Completions should come from vLLM *or* something else like SGLang, then the logic for transforming into a Completion should live within a specific vLLM implementation of a Policy, which right now means in this file.

If the vLLM specific logic leaks into the core Completion class, then it kinda defeats the purpose of having it.

WDYT @allenwang28 ?

Ritesh1905: Other option is, we could just have a completion util or something which can do these helper conversions, so that we can keep both policy and Completion clean. Thoughts?

joecummings: This would still introduce another location in the code that would have to be updated in order to provide a backend for policy, no? 

My ideal endstate is that we have a vllm_policy.py and a sglang_policy.py with all necessary logic very neatly contained in those files.

Ritesh1905: Ok. Will keep this here.

allenwang28: yeah I understand your point @joecummings - if we expect all of the integration points to be classmethods in Completion, that limits its extensibility to other integrations like SGLang (counterpoint could be "how many inference engines are there actually / should Forge actually support them all?")

but that's not really my point, my point is mostly around just discoverability:
- At some point I discover the Forge specific `Completion` and wonder how it works with e.g. vLLM
- If I click on the `Completion` it would be nice to see `from_vllm` components right there 

otherwise I have to grep the codebase for `Completion` to find that it is in the vLLM specific functionality. This is fine and logical and maybe not that big of an issue? Maybe something we can do easily for now is just have this be a top-level function 

```
def request_output_to_completion(request_output: RequestOutput) -> Completion:
    ...
```

which is admittedly a nit lol

Ritesh1905: I am keeping this here for now and we can clean this up in the next PRs, when we settle this very important debate. lol

Issue #201, opened 2025-09-19
memory leak?


casteryh:
The main GRPO training loop is crashing after 4 steps.
This is at
```
> git log --oneline | head -n 1
605f85f Trainer + reference model in bf16 (#189)
```
<img width="610" height="308" alt="Image" src="https://github.com/user-attachments/assets/edaabe0a-0559-4cbf-87eb-6ef63ee647c2" />

https://meta.wandb.io/torchforge/grpo-training/runs/433bhje2
https://www.internalfb.com/phabricator/paste/view/P1955530803

allenwang28:
It's probably [this guy](https://github.com/meta-pytorch/forge/blob/main/src/forge/actors/trainer.py#L218)

we should be deleting and freeing the memory, but I also wonder if we can just do this computation on CPU

PR #209, opened 2025-09-22
Garbage collect on every train / ref model step

joecummings:
### What is this PR?

As reported in #201, there appeared to be a memory leak in the components using TorchTitan - RLTrainer and ReferenceModel. This PR manually calls the [TorchTitan garbage collection util](https://github.com/pytorch/torchtitan/blob/c9cb3046867ca3cacd6771a60acf65ede424715e/torchtitan/tools/utils.py#L37) on every step, which eliminates the memory leak.

### How was this PR tested?

With ``gc_freq: 100000`` (never running essentially)
Screenshot 2025-09-22 at 1 43 19 PM

Notice the yellow line (reference model) and the ref line (trainer model) consistently going up.

With ``gc_freq: 1`` (run every step)
Screenshot 2025-09-22 at 3 19 55 PM

Notice the yellow line (reference model) and red line (trainer model) stays relatively flat.

### FAQs

1. **Why is this happening?** ~~The theory is that Monarch doesn't know that it is able to free memory after returning from an Endpoint. So every time the Endpoint (either ``forward`` or ``train_step``) is called, new memory is allocated. This theory definitely requires further investigation, but the fix lends credibility.~~ Actually Titan disables garbage collection manually to improve performance. We have to re-enable it with this PR.
2. **How long does garbage collection take? Could this be a bottleneck?** Garbage collection times are logged. In my experiments the longest it takes is 1 ms.

### DIFF
--- a/apps/grpo/qwen3_1_7b.yaml
+++ b/apps/grpo/qwen3_1_7b.yaml
@@ -48,6 +48,7 @@ trainer:
     max_norm: 1.0
     steps: 1000000
     dtype: bfloat16
+    gc_freq: 1
   compile:
     enable: false
   parallelism:
@@ -83,6 +84,7 @@ ref_model:
     hf_assets_path: hf://${model}
   training:
     dtype: bfloat16
+    gc_freq: 1
   compile:
     enable: false
   parallelism:
--- a/src/forge/actors/reference_model.py
+++ b/src/forge/actors/reference_model.py
@@ -37,7 +37,7 @@ class ReferenceModel(ForgeActor):
     compile: Compile = field(default_factory=Compile)
     training: Training = field(
         default_factory=Training
-    )  # Only needed in order to correctly set a lower dtype
+    )  # Needed in order to set attrs like dtype, garbage collection freq, etc.
 
     # Populated in setup
     # TODO: Commented out since engine_config parsing extracts from class members
@@ -61,6 +61,7 @@ class ReferenceModel(ForgeActor):
         """
         self.rank = current_rank().rank
         self.size = math.prod(current_size().values())
+        self.step = 0
 
         env = {
             "RANK": str(self.rank),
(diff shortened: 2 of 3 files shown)


apps/grpo/qwen3_1_7b.yaml
```diff
@@ -48,6 +48,7 @@ trainer:
     max_norm: 1.0
     steps: 1000000
     dtype: bfloat16
+    gc_freq: 1
```
allenwang28, line 51: hmm do we expect users to actually tune this? Should we just assume 1 until we're able to fix this in torchtitan / Monarch / wherever?

allenwang28: hmm nvm i understand now that it's set in the config

src/forge/actors/trainer.py
```diff
@@ -192,10 +193,10 @@ def train_step(
         self.engine.optimizers.zero_grad()
         self.engine.lr_schedulers.step()
 
-        self.current_step += 1
+        self.step += 1
         self.engine.checkpointer.save(
```
Ritesh1905, line 197: question: Why do we do CP save on evert step?

src/forge/actors/reference_model.py
```diff
@@ -83,6 +84,7 @@ async def setup(self):
 
     @endpoint
     async def forward(self, input_ids: torch.Tensor) -> torch.Tensor:
+        self.engine.gc_handler.run(self.step)
```
Ritesh1905, line 87: Can you explain what is this doing internally?

src/forge/actors/trainer.py
joecummings: This is a Titan checkpointer impl detail. What actually happens is that it checks if it _should save_, which is determined by the checkpoint frequency attr found in our config. If it shouldn't checkpoint it just returns. See [here](https://github.com/pytorch/torchtitan/blob/c9cb3046867ca3cacd6771a60acf65ede424715e/torchtitan/components/checkpoint.py#L470).

A much much much better name would be `maybe_save` IMO

src/forge/actors/reference_model.py
joecummings: Code can be found here: https://github.com/pytorch/torchtitan/blob/c9cb3046867ca3cacd6771a60acf65ede424715e/torchtitan/tools/utils.py#L37

src/forge/actors/trainer.py
vidhyav: Curious why isn't it a "finally" step that's done by default?

PR #213, opened 2025-09-22
Perform to_vllm conversions all on CPU

allenwang28:
This may be temporary, but the to_vllm script on Qwen 32B fails at the `to_vllm` step because it's all done on CUDA implicitly. This approach gathers and converts weights to CPUs.

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -259,6 +259,9 @@ def _qwen3_hf_to_vllm(sd: dict[str, Tensor], num_layers: int) -> dict[str, Tenso
         """Unwrap a DTensor to a Tensor."""
         return t.full_tensor() if isinstance(t, torch.distributed.tensor.DTensor) else t
 
+    for key in sd.keys():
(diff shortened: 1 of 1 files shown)


src/forge/actors/trainer.py
```diff
@@ -255,6 +255,10 @@ def _qwen3_hf_to_vllm(sd: dict[str, Tensor], num_layers: int) -> dict[str, Tenso
     """
     load_sd = {}
 
+    def unwrap(t):
+        """Unwrap a DTensor to a Tensor."""
+        return t.full_tensor() if isinstance(t, torch.distributed.tensor.DTensor) else t
+
     # Copy over directly mapped keys
```
casteryh, line 265: nit: wouldn't it be easier if you do the following?
```python
for key in sd.keys():
    sd[key] = unwrap(sd[key]).cpu()
```

allenwang28: haha thanks @casteryh

PR #227, opened 2025-09-24
metric logging - UX and perf tracker (not profiler) (2/N)

felipemello1:
Ddds perf_tracker.py to track time/peak memory. Users can use it as decorator, ctx manager, and timer. Also can enable `use_cuda`. Followed Monarchs implementation (https://github.com/meta-pytorch/monarch/blob/main/python/monarch/timer/execution_timer.py). Main differences are 
a) Monarch has a global variable to track threads, which scared me a bit, while this is per instance, 
b) i set daemon=True (ensure the background polling threads auto terminate if the process exits, avoiding potential hangs)
c) added ability to add steps

Also added unit tests to confirm that async is still async and these functions are not getting in the way

-------

General UX example of what metric logging could look like.

Main files:
apps/toy_metrics/main.py
src/forge/observability/perf_tracker.py

Toy example output:
https://wandb.ai/cabernet-team/my_project/runs/shgaczz7?nw=nwuserfelipemello
image

## Your superpowers:

1. Record metrics anywhere in the codebase
```python
from forge.observability.metrics import record_metric, ReductionType
await record_metric("policy/count_requests", 1, ReductionType.SUM)
```

2. Use context manager or decorator to get time + memory delta and peak. Optionally enable flag to synchronize event (synchronizes single rank, not all of them)

```python
from forge.observability.perf_tracker import record_perf_metrics, record_perf_metrics_ctx,

class TrainActor(ForgeActor):
    @endpoint
    @record_perf_metrics(
        "trainer_perf", track_time=False, track_memory=True, sync_cuda_event=False
    )
    async def train_step(self, step: int):
		...

class GeneratorActor(ForgeActor):
    @endpoint
    async def generate_step(self, step: int):
        async with record_perf_metrics_ctx(
            "policy_perf", track_time=True, track_memory=False, sync_cuda_event=False
        ):
            ...
```

3. Use timer to easily get time deltas

```python
from forge.observability.perf_tracker import StepTimer

timer = StepTimer("trainer_perf/step")
await timer.start()
foo()
await timer.step("foo")
bar()
await timer.step("bar")
await timer.end()

# Metrics logged:
# trainer_perf/step/foo/duration_avg (s)
# trainer_perf/step/foo/duration_max (s)
# 
# trainer_perf/step/bar/duration_avg (s)
# trainer_perf/step/bar/duration_max (s)
# 
# trainer_perf/step/total_duration_avg (s)
# trainer_perf/step/total_duration_max (s)
```

### DIFF
--- a/src/forge/observability/perf_tracker.py
+++ b/src/forge/observability/perf_tracker.py
@@ -9,8 +9,7 @@ import os
 import threading
 import time
 
-from concurrent.futures import ThreadPoolExecutor
-from contextlib import contextmanager
+from concurrent.futures import Future, ThreadPoolExecutor
 from functools import lru_cache, wraps
 from typing import List, Optional, Protocol, Tuple
 
@@ -44,39 +43,42 @@ def _warn_nested_memory_tracking(prefix: str) -> None:
 
 
 """
-==========
-Step Timer
+
+
+class Tracer:
 ==========
 """
 
 
-class Timer:
+class Tracer:
     """
-    Multi-step timer that records average and max timing metrics.
+    Tracer with multi-step timing and optional memory tracking at start/stop boundaries.
+    Steps only affect timing; memory is tracked from start() to stop().
 
-    Supports non-blocking CUDA timing via CUDA events and background polling threads;
-
+    Supports non-blocking CUDA timing via CUDA events and background polling threads.
     Aggregation is handled externally by the metrics system via record_metric.
 
-    User must call start() and end() explicitly.
-    Supports reuse: after calling end(), you may call start() again to begin a new timing session.
+    User must call start() and stop() explicitly.
+    Supports reuse: after calling stop(), you may call start() again to begin a new timing session.
 
     Local env flag DISABLE_PERF_METRICS can be used to skip all timing operations.
     Local env flag METRIC_TIMER_USES_CUDA can be used to set CUDA timing.
 
     Args:
         prefix (str): Prefix for metric names, e.g. "my_prefix" -> "{my_prefix}/{step_name}/duration_avg_s".
-        use_cuda (bool): Sets CUDA timing if True and CUDA is available.
+        track_time (bool): Whether to track execution time. Defaults to True.
+        track_memory (bool): Whether to track CUDA memory usage. Defaults to False.
+        time_with_gpu (bool): Sets CUDA timing if True and CUDA is available.
 
     Example:
-        timer = Timer("my_prefix")
-        timer.start()
+        tracer = Tracer("my_prefix", track_memory=True)
+        tracer.start()  # Memory tracking starts here
         time.sleep(0.1)  # Work for step "a"
-        timer.step("my_step_a")
+        tracer.step("my_step_a")  # Only affects timing
         for i in range(1, 4):  # 3 iterations
             time.sleep(i/10)  # 0.1, 0.2, 0.3 seconds
-            timer.step("my_step_b")
-        timer.end()  # Records metrics, including total time end-start.
+            tracer.step("my_step_b")  # Only affects timing
+        tracer.stop()  # Memory tracking ends here, records metrics
 
         >>> Records:
         >>> my_prefix/my_step_a/duration_avg_s: 0.1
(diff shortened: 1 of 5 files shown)


src/forge/observability/perf_tracker.py
```diff
@@ -0,0 +1,462 @@
...
+    Multi-step timer that records average and max timing metrics.
+
+    Supports non-blocking CUDA timing via CUDA events and background polling threads;
+
+    Aggregation is handled externally by the metrics system via record_metric.
+
+    User must call start() and end() explicitly.
+    Supports reuse: after calling end(), you may call start() again to begin a new timing session.
+
+    Local env flag DISABLE_PERF_METRICS can be used to skip all timing operations.
```
allenwang28, line 64: If we're introducing env variables, we should add a constants file for all env names

src/forge/observability/perf_tracker.py
```diff
@@ -0,0 +1,462 @@
...
+        # Pre-allocate slot with placeholder to preserve ordering when pooling
+        with self._lock:
+            idx = len(self._durations)
+            self._durations.append((name, -1.0))  # Placeholder, -1 indicates pending
+
+        stream = torch.cuda.current_stream()
+        end_event = torch.cuda.Event(enable_timing=True)
+        end_event.record(stream)
+
+        thread = threading.Thread(
```
allenwang28: Logic all looks right to me, but is this implementable with asyncio primitives? It'd probably be better for Forge altogether if we're `asyncio.sleep(...)` etc. because iiuc with threading and `time.sleep()` it will still block the main event loop

src/forge/observability/perf_tracker.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+"""
+=======================================
+Memory+timer as decorator / ctx manager
+=======================================
+"""
+
+
+def record_perf_metrics(
```
allenwang28: we need better names for this and `record_perf_metrics_ctx`. For one I think it should have the same name for both the decorator and the context managed version. 

Can I propose a few changes?
```
# timer => tracer
t: Tracer = tracer.start("train_step")
t.step("fwd")
t.step("bwd")
t.stop()

# decorated / context are both trace
@trace("train_step")
async def train_step():
    pass

# context manager returns a tracer object too
with trace("train_step") as tracer:
    ...
    tracer.step("fwd")
```

src/forge/observability/perf_tracker.py
felipemello1: added

src/forge/observability/perf_tracker.py
felipemello1: i measured it and its pretty minimal, <1ms. I dont think its worth the refactoring at this point. Later we could set "DISABLE_PERF_METRICS" and see if there is a signficant change in end2end. Is that fine? The join is not cuda blocking.

src/forge/observability/perf_tracker.py
felipemello1: done

src/forge/observability/perf_tracker.py
```diff
@@ -0,0 +1,447 @@
...
+
+    def stop(self) -> None:
+        if self._disable:
+            return
+        if not self._active:
+            raise ValueError("Tracer must be started before calling stop")
+
+        # Stop timing
+        if self._timer:
+            self._timer.step("end")  # dropped from steps, included in total
```
allenwang28, line 159: just clarifying, the "end" isn't some special token/key right? It's just that the timer itself drops it in its stop implementation?

src/forge/observability/perf_tracker.py
```diff
@@ -0,0 +1,447 @@
...
+            ) and torch.cuda.is_available()
+            use_cpu = not time_with_gpu_events
+            self._timer = _TimerCPU() if use_cpu else _TimerCUDA()
+            self._timer.start()
+
+        # Start memory tracking
+        if self.track_memory:
+            self._start_memory_tracking()
+
+    def step(self, step_name: str) -> None:
```
allenwang28, line 142: why can `step()` not be used in conjunction with memory tracking?

src/forge/observability/perf_tracker.py
```diff
@@ -0,0 +1,447 @@
...
+            self._timer = None
+
+        # Stop memory tracking
+        if self._memory_started:
+            self._stop_memory_tracking()
+
+        self._active = False
+
+    def _start_memory_tracking(self) -> None:
+        is_outer_scope = not _is_memory_active()
```
allenwang28, line 170: IMO it isn't intuitive to the user that you can't have nested memory tracking. I'm guessing the engineering complexity is that all of the metrics we derive are from `torch.cuda.memory_allocated()` and `torch.cuda.max-memory_allocated()`? That it's a singleton and that nesting it resets it for other timers?

src/forge/observability/perf_tracker.py
```diff
@@ -0,0 +1,447 @@
...
+            _set_memory_active(True)
+            torch.cuda.reset_max_memory_allocated()
+            self._start_mem = torch.cuda.memory_allocated()
+            self._memory_started = True
+
+    def _stop_memory_tracking(self) -> None:
+        if not self._memory_started:
+            return
+
+        end_mem = torch.cuda.memory_allocated()
```
allenwang28, line 189: I guess this automatically picks up the device? Can we add a TODO to consider using the provisioner assigned device ID?

src/forge/observability/perf_tracker.py
allenwang28: if we're going to punt it then we should create an issue for it to make sure we actually get back to it

src/forge/observability/perf_tracker.py
felipemello1: correct. We append the time --> compute total --> drop this last one so we dont record the metric "end". But its not a dict key, so no risk to confuse it with users given keys

src/forge/observability/perf_tracker.py
felipemello1: it can, but lets leave it for a follow up?

src/forge/observability/perf_tracker.py
felipemello1: "That it's a singleton and that nesting it resets it for other timers?" - correct

maybe theres a way? but the above is my understanding

src/forge/observability/perf_tracker.py
allenwang28: ok, to make it clearer can we just add a `def stop()`? even if that logic is the exact same. It looks fragile otherwise. Feel free to add it to a list of follow ups.

src/forge/observability/perf_tracker.py
allenwang28: ok let's add it to a list of follow ups. I like to create an issue tracking these, and embed just `TODO (issues/{issue_number})`

src/forge/observability/perf_tracker.py
allenwang28: ok, I think it's fine for an initial version. Main caveats:
- Let's mark memory tracking capabilities as experimental within the Timer class
- Create a separate issue for re-designing this

since we need it now let's prioritize landing

PR #228, opened 2025-09-24
Fix Policy.launch to be compatible with new ForgeActor.launch API

DNXie:
This PR updates `Policy.launch` to match the new `ForgeActor.launch` API, removing the `process_config` argument and construct it inside of `launch`. 

Specifically fixed the error
```
TypeError: Policy.launch() missing 1 required keyword-only argument: 'process_config'
```

**Test**
```
pytest -s --log-cli-level=INFO  tests/unit_tests/test_service.py
```

cc @casteryh @vidhyav @felipemello1

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -150,7 +150,6 @@ class Policy(PolicyInterface):
     async def launch(  # pyright: ignore[reportIncompatibleMethodOverride]
         cls: type["Policy"],
         *,
-        process_config: ProcessConfig,
         engine_config: EngineConfig | Mapping = EngineConfig(),
         sampling_config: SamplingConfig | Mapping = SamplingConfig(),
         available_devices: str | None = None,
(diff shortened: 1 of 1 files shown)


src/forge/actors/policy.py
```diff
@@ -150,14 +150,18 @@ def __post_init__(self):
...
         *,
-        process_config: ProcessConfig,
         engine_config: EngineConfig | Mapping = EngineConfig(),
         sampling_config: SamplingConfig | Mapping = SamplingConfig(),
         available_devices: str | None = None,
         **kwargs,
     ) -> "Policy":
         # Note - get_proc_mesh will set MASTER_ADDR, MASTER_PORT and CUDA_VISIBLE_DEVICES
         # automatically.
+        process_config: ProcessConfig = ProcessConfig(
```
casteryh, line 160: no action required.
just curious why this wasn't an issue before? is it because we got rid of process_config in the new config format?

DNXie: `Policy.launch` inherited from `ForgeActor.launch`. This wasn’t an issue before because `ForgeActor.launch` required `process_config`. After the PR https://github.com/meta-pytorch/forge/pull/195, `ForgeActor.launch` no longer takes a  `process_config` and constructs a default `process_config`, but `Policy.launch` still required it, which is a mismatch.

PR #217, opened 2025-09-22
integration test for weight sync that actually tests behavior

casteryh:
Previously, the test was bogus because
- it simply repeatead the the logic of the implementation
- it would pass without any problems if you commented out update_weights and push_weights, since all the weights had been the same from the start to the end.

PR #229, opened 2025-09-24
Fix manual vLLM Qwen3 sharding bug when trainer export the weights

JenniferWang:
Thanks to @casteryh 's PR of fixing the integration test https://github.com/meta-pytorch/forge/pull/217, we are able to see the buggy behavior of the manual sharding logic.

This diff basically showcases _why_ the test failed and how easy it is to introduce **silent** data correctness issue if we keep pursuing this route of manual sharding.  

The plan is land these two PRs ASAP
1. Fix the broken test: https://github.com/meta-pytorch/forge/pull/217
2. Replace the manual sharding with vanilla load weight calls: https://github.com/meta-pytorch/forge/pull/184  

At the same time, this PR fixes the existing Qwen3 (non MoE) sharding. 

### Before the fix:
* `test_policy_update` integration test fails when tp > 1
* `grpo` (Qwen 1.7B) loss function super high at the start of the trainer
image
https://meta.wandb.io/jiyue/grpo-training/runs/hh3bht2w?nw=nwuserjiyue

### After the fix:
*  `test_policy_update` integration test passes when tp > 1
* `grpo` (Qwen 1.7B) loss function is much more reasonable??
image
https://meta.wandb.io/jiyue/grpo-training/runs/gnvlw7dc?nw=nwuserjiyue

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -358,7 +358,9 @@ async def main(cfg: DictConfig):
                 loss = await trainer.train_step.route(inputs, targets)
                 training_step += 1
                 mlogger.log("loss/training_step", loss, training_step)
-                await trainer.push_weights.fanout(training_step, vllm_tp_DEPRECATED=policy_tp_size)
+                await trainer.push_weights.fanout(
+                    training_step, vllm_tp_DEPRECATED=policy_tp_size
+                )
                 await policy.update_weights.fanout(training_step)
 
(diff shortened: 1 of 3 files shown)


allenwang28:
@JenniferWang life pro tip, you should be able to do `pre-commit install` (assuming you've `pip install pre-commit`) which will handle the linter for you automatically every time you do a `git commit`

PR #231, opened 2025-09-24
Change services to actors except Policy and drop apps/rl

DNXie:
Within all of our applications, everything is currently a Service. For example, in [apps/grpo/main](https://github.com/meta-pytorch/forge/blob/main/apps/grpo/main.py), trainer and replay_buffer should be actors.

**Summary**:
* Actors: dataloader, trainer, replay_buffer, compute_advantages
* Services: Policy, ref_model, reward_actor

**Changes**:
* Updated `apps/grpo`, `apps/toy_rl` (with dcp off)
* Updated `Policy` to take `use_dcp` from config. 
* Dropped apps/rl since it is deprecated.
* Single actor call `call_one` instead of `choose`. The difference is that [`call_one` makes sure the caller is a singleton](https://github.com/meta-pytorch/monarch/blob/e9aebe22ce5ddbe3181ae4d5fc52b31b491476ff/python/monarch/_src/actor/endpoint.py#L103-L123).

**Test**
```
python -m apps.grpo.main --config apps/grpo/qwen3_8b.yaml
python -m apps.toy_rl.sumdigits --config apps/toy_rl/sumdigits.yaml
```

I didn't run 
```
python -m apps.grpo.main --config apps/grpo/qwen3_multinode.yaml
```
since the config is outdated. Looks like this one is already deprecated.

cc @Ritesh1905

### DIFF
--- a/apps/toy_rl/sumdigits.py
+++ b/apps/toy_rl/sumdigits.py
@@ -16,9 +16,9 @@ from typing import Any
 import torch
 import torch.nn.functional as F
 import torchstore as ts
+from forge.actors._torchstore_utils import get_param_key
 from forge.actors.policy import Policy
 from forge.actors.replay_buffer import ReplayBuffer
-from forge.actors.torchstore_utils import get_param_key
 from forge.actors.trainer import _qwen3_hf_to_vllm
 from forge.cli.config import parse
 from forge.controller.actor import ForgeActor
@@ -552,9 +552,7 @@ async def main(cfg: DictConfig):
                 training_step += 1
                 mlogger.log("loss/training_step", loss, training_step)
                 print(f"loss/training_step: {loss} at training step {training_step}")
-                await trainer.push_weights.call(
-                    training_step, vllm_tp_DEPRECATED=policy_tp_size
-                )
+                await trainer.push_weights.call(training_step)
                 await policy.update_weights.fanout(training_step)
                 # NOTE: hard-coded to be on-policy for faster convergence
                 await replay_buffer.clear.call()
--- a/apps/toy_rl/sumdigits-tp.yaml
+++ b/apps/toy_rl/sumdigits-tp.yaml
@@ -42,22 +42,10 @@ replay_buffer:
   dp_size: 1
 
 services:
-  dataset:
-    procs: 1
-    num_replicas: 1
-    with_gpus: false
   policy:
     procs: 1
     num_replicas: 1
     with_gpus: true
-  trainer:
-    procs: 1
-    num_replicas: 1
-    with_gpus: true
-  replay_buffer:
-    procs: 1
-    num_replicas: 1
-    with_gpus: false
   reward_actor:
     procs: 1
     num_replicas: 1
@@ -66,3 +54,14 @@ services:
     procs: 1
     num_replicas: 1
     with_gpus: true
+
+actors:
+  dataset:
+    procs: 1
+    with_gpus: false
+  trainer:
+    procs: 1
+    with_gpus: true
+  replay_buffer:
+    procs: 1
+    with_gpus: false


casteryh:
Would love some clarification on service vs policy.
Especially why we have this distinction at all.
cc @allenwang28

allenwang28:
> Would love some clarification on service vs policy. Especially why we have this distinction at all.

I am going to assume you mean service vs actor! It's a good question. For context, the main reason we have services in the first place is to exactly handle load balancing and fault tolerance. These are things which Monarch doesn't give you out of the box, but gives you the capabilities to implement. We surely want this for something like vLLM, or if we want the ability to spin up and load balance across multiple execution environments in-band.

We initially had everything just be services for simplicity, but in reality not everything needs to be a service. For replay buffer, trainer, etc., you don't need the routing capabilities. Additionally, the fault tolerance story for those are not as well defined as that of the policy/environments/reference model. 

Since the world will also learn about Monarch, I think the layering is clearer this way - Actors are the base capabilities from Monarch, Service is a distinct abstraction that's built on top of it.

apps/grpo/qwen3_1_7b.yaml
```diff
@@ -100,31 +100,27 @@ ref_model:
...
+    with_gpus: false
   trainer:
     procs: 1
-    num_replicas: 1
     with_gpus: true
   replay_buffer:
     procs: 1
-    num_replicas: 1
     with_gpus: false
   ref_model:
```
allenwang28, line 108: actually I was wrong - I think we only have trainer and replay buffer be actors, the rest are ok to keep as services

DNXie: It makes sense. Why are `dataset`, `compute_advantages`, and `reward_actor` also services? Do they need replicas?

DNXie: If `num_replica=1`, what's the difference between actor and service?

DNXie: Done!

allenwang28: sorry Danning, Dataset is an actor, compute_advantages is an actor, reward_actor is a service

apps/grpo/main.py
```diff
@@ -283,9 +283,9 @@ async def main(cfg: DictConfig):
     # ---- Core RL loops ---- #
     async def continuous_rollouts():
         rollout_count = 0
-        pad_id = await dataloader.pad_token.route()
+        pad_id = await dataloader.pad_token.choose()
```
allenwang28: use `call_one()` here  - it makes it clear that we expect dataloader to be a singleton

apps/grpo/main.py
```diff
@@ -283,9 +283,9 @@ async def main(cfg: DictConfig):
...
         while True:
-            sample = await dataloader.sample.route()
+            sample = await dataloader.sample.choose()
```
allenwang28: `call_one`

apps/grpo/main.py
```diff
@@ -332,17 +332,17 @@ async def continuous_rollouts():
             del ref_logits, ref_logprobs, input_ids
 
             # Calculate advantages and add to replay buffer
-            advantages = await compute_advantages.compute.route(group)
+            advantages = await compute_advantages.compute.choose(group)
```
allenwang28: `call_one()`

apps/grpo/main.py
```diff
@@ -332,17 +332,17 @@ async def continuous_rollouts():
...
             for episode, advantage in zip(group.episodes, advantages):
                 episode.advantage = advantage
-                await replay_buffer.add.route(episode)
+                await replay_buffer.add.choose(episode)
```
allenwang28: `call_one()`

apps/grpo/main.py
```diff
@@ -332,17 +332,17 @@ async def continuous_rollouts():
...
 
             # Log metrics
             avg_response_len = (
                 sum(len(e.response_tokens) for e in group.episodes) / group_size
             )
             mlogger.log("avg_response_len/rollout", avg_response_len, rollout_count)
-            buffer_size = await replay_buffer._numel.route()
+            buffer_size = await replay_buffer._numel.choose()
```
allenwang28: `call_one()`

apps/toy_rl/sumdigits.py
```diff
@@ -16,9 +16,9 @@
 import torch
 import torch.nn.functional as F
 import torchstore as ts
+from forge.actors._torchstore_utils import get_param_key
```
DNXie, line 19: Fixed the typo here. CC @casteryh

casteryh: thanks!

PR #233, opened 2025-09-24
Adds a Qwen 32B config

allenwang28:
Submitting the Qwen 32B config I was using to enable testing


apps/grpo/qwen3_32b.yaml
```diff
@@ -0,0 +1,134 @@
+# Grouped Relative Policy Optimization (GRPO)
+# >>> python -m apps.grpo.main --config apps/grpo/qwen3_8b.yaml
+# NOTE - this runs functionally, but has not been verified
+# for convergence or correctness yet.
+
+# Global configuration
+group_size: 2
```
joecummings, line 7: Why so small??

allenwang28: because it takes a really long time to run a step with a bigger group size

JenniferWang: lol I ran into the same issue for the 32B MoE

PR #234, opened 2025-09-25
Add generator_version into Completion Interface

Jack-Khuu:
GRPO currently uses 2 separate entrypoint call (`get_version`, `generate`) for generator results and version retrieval in the rollout loop. 

Given the replica/distributed nature of services, this can produce potentially inconsistencies if the policy weights were updated in between call (or hit a replica that has already updated). 

This PR just pushes the policy version into the generation payload

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -286,9 +286,12 @@ async def main(cfg: DictConfig):
                 return
             prompt, target = sample["request"], sample["target"]
             responses = await policy.generate.route(prompt)
-            assert len(responses) > 0
-            version = responses[0].generator_version
-            assert version is not None, "Response must indicate a version"
+            assert (
+                len(responses) > 0
+            ), "Sanity check: Responses should NEVER return empty"
+            assert (
+                version := responses[0].generator_version
+            ) is not None, "Response must indicate a version"
             group = Group.new_group(
                 group_id=rollout_count,
                 group_size=group_size,


apps/grpo/main.py
```diff
@@ -286,8 +286,9 @@ async def continuous_rollouts():
                 return
             prompt, target = sample["request"], sample["target"]
             responses = await policy.generate.route(prompt)
-            # TODO: this shall be part of the responses metadata instead of a separate call
-            version = await policy.get_version.route()
+            assert len(responses) > 0
```
joecummings: I'd rather something like this just be a test. No need to expose this logic in the app.

src/forge/data_models/completion.py
```diff
@@ -33,3 +33,6 @@ class Completion:
 
     # the reason for stopping the generation
     stop_reason: str | None = None
+
+    # the version identifier of the model when the generation was performed
+    generator_version: int | None = None
```
joecummings, line 38: The terminology is getting mixed up again haha. Can we stick with generator OR policy? Then should update all related call sites.

apps/grpo/main.py
allenwang28: I think some type of assertion in the main app makes sense - the only caveat is if we're going to do it, it should clearly message what failure is and how the user should avoid it. 

e.g. in what case would the generated responses ever be empty? If we don't expect it to ever be, we don't need an assertion. (I don't think this should ever return an empty response)

Jack-Khuu: It should never return an empty response, so I'll add a message for it

Main reason I have it here is that we're doing a blind index into a list otherwise

PR #236, opened 2025-09-25
Log KL metrics

joecummings:
Not relevant after @felipemello1 but needed for general visibility right now

Screenshot 2025-09-25 at 11 40 38 AM


felipemello1:
if its not super duper urgent, i would say hold off. Hopefully we can land the metric logging PR between tonight and tomorrow, i can have that metric there. How does that sound?

joecummings:
> if its not super duper urgent, i would say hold off. Hopefully we can land the metric logging PR between tonight and tomorrow, i can have that metric there. How does that sound?

Sounds incredibly reasonable how dare you :)

PR #238, opened 2025-09-25
Add the ability to customize threads in GRPO App

Jack-Khuu:
Adds the ability to adjusts threads used for rollout and training via the yaml (`rollout_threads`, `training_threads`) defaulting to 1 when not specified.

Also increases the number of policy replicas to 6 for 8 GPU utilization (6 Policy, 1 Trainer, 1 Reference)

---

Initial testing on Qwen3 1.7B (gs=8, bs=16)

_(We'll have a better tools to analyze configs after https://github.com/meta-pytorch/forge/pull/239)_

| threads | time/train loop (min) | % of base case | Comment|
|-------|------|---|---|
| 1  | 0.7  | 100% | Base Case|
| 2 | 0.45 | 64% | Thread x GS == BS |
| 6 | 0.36 | **51%** | Threads === Replica|
| 8 | 0.47 | 67% | Threads > Replica|

### wandb: torchforge/grpo-training/runs/buwmxe2n
- 1 Rollout Thread, ~7 minutes
- ~20 rollouts => **10 training steps** completed
- Notice that the rollout needs to loop twice per training run
image

### wandb: torchforge/grpo-training/runs/3dwmpt0s
- 2 Rollout Threads, ~ 9 minutes
- ~20 rollouts (per thread) => **20 training steps** completed
- Notice that each version of rollout steps (completed by all threads) correponds to a single training step completion
image

### wandb: torchforge/grpo-training/runs/zi7mogyb
- 6 Rollout Thread, ~ 10 minutes
- ~20 rollouts (per thread) => **28 training steps** completed
- Notice the the replaybuffer over populates (bs=16) meaning over generation by rollouts and resulting in discarded samples
- Note that the number of threads matches the number of replicas (replica=6) resulting in at most 6 concurrent vllm runs
image

### wandb: torchforge/grpo-training/runs/fqcdp31y
- 8 Rollout Thread, ~ 14 minutes
- ~20 rollouts (per thread) => **30 training steps** completed
- Notice the the replaybuffer over populates (bs=16) meaning over generation by rollouts and resulting in discarded samples
- Note that the number of threads is greater than the number of replicas (replica=6) resulting in at most 6 concurrent vllm runs
image

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -366,18 +366,15 @@ async def main(cfg: DictConfig):
     rollout_tasks = [
         asyncio.create_task(continuous_rollouts()) for _ in range(num_rollout_threads)
     ]
-    training_tasks = [
-        asyncio.create_task(continuous_training()) for _ in range(num_training_threads)
-    ]
+    training_task = asyncio.create_task(continuous_training())
 
     try:
-        await asyncio.gather(*rollout_tasks, *training_tasks)
+        await asyncio.gather(*rollout_tasks, training_task)
     except KeyboardInterrupt:
         print("Training interrupted by user")
         for rollout_task in rollout_tasks:
             rollout_task.cancel()
-        for training_task in training_tasks:
-            training_task.cancel()
+        training_task.cancel()
     finally:
         print("Shutting down...")
         await asyncio.gather(
--- a/apps/grpo/qwen3_1_7b.yaml
+++ b/apps/grpo/qwen3_1_7b.yaml
@@ -11,7 +11,6 @@ off_by_n: 1 # Off by one by default
 
 # Main loop configuration
 rollout_threads: 1   # Recommended to set equal to policy.num_replicas
-training_threads: 1  # Recommended to set equal to trainer.num_replicas (usually 1)
 
 # Dataset configuration
 dataset:


felipemello1:
> Can you show us some vroom numbers??

thats on me. It should be easier when logging gets merged, which will happen today

Jack-Khuu:
The config we have now is definitely not optimized, but drawing a summary from initial data:

| threads | time/train loop (min) | % of base case | Comment|
|-------|------|---|---|
| 1  | 0.7  | 100% | Base Case|
| 2 | 0.45 | 64% | Thread x GS == BS |
| 6 | 0.36 | **51%** | Threads === Replica|
| 8 | 0.47 | 67% | Threads > Replica|


When # of Threads == # of Replica, we were training twice as fast in this case

apps/grpo/qwen3_1_7b.yaml
```diff
@@ -9,6 +9,10 @@ max_res_tokens: 512
 model: "Qwen/Qwen3-1.7B"
 off_by_n: 1 # Off by one by default
 
+# Main loop configuration
+rollout_threads: 8
```
joecummings:
```suggestion
rollout_threads: 8  # Recommended to set equal to policy.num_replicas
```

joecummings: Also for this default config, should we keep replicas and rollout threads to 1?

apps/grpo/qwen3_1_7b.yaml
```diff
@@ -9,6 +9,10 @@ max_res_tokens: 512
...
+training_threads: 1
```
joecummings:
```suggestion
training_threads: 1. # Recommended to set equal to trainer.num_replicas (usually 1)
```

apps/grpo/main.py
```diff
@@ -358,17 +358,26 @@ async def continuous_training():
...
-    # TODO: Start multiple rollouts once all serivces support it
-    rollout_task = asyncio.create_task(continuous_rollouts())
-    training_task = asyncio.create_task(continuous_training())
+    num_rollout_threads = cfg.get("rollout_threads", 1)
+    num_training_threads = cfg.get("training_threads", 1)
+    print(
+        f"Starting GRPO with {num_rollout_threads} rollout threads, {num_training_threads} training threads"
+    )
+    rollout_tasks = [
+        asyncio.create_task(continuous_rollouts()) for _ in range(num_rollout_threads)
```
joecummings, line 367: Any way we could get numbers for if we have extra replicas for reference model????

JenniferWang: Is there an easy way for the metric logger to group metrics by meshes? @felipemello1

felipemello1: yes, you can put it in the metric name, i.e. 'my_metric_{mesh_num}'. You can also just do "reduce_across_ranks=False", and every rank will have its own logger, generating N wandb runs.

felipemello1: PR 1/N landed. Addressing some ocmments on 2/N and we should merge it soon

Jack-Khuu: > Any way we could get numbers for if we have extra replicas for reference model????

Definitely doable with some spot testing. Of the ones tested, but the only ones that would benefit would be threads: 1, 2 since it's already overpopulating the buffer for threads >2

apps/grpo/qwen3_1_7b.yaml
```diff
@@ -9,6 +9,10 @@ max_res_tokens: 512
 model: "Qwen/Qwen3-1.7B"
 off_by_n: 1 # Off by one by default
 
+# Main loop configuration
+rollout_threads: 1   # Recommended to set equal to policy.num_replicas
+training_threads: 1  # Recommended to set equal to trainer.num_replicas (usually 1)
```
joecummings: After discussion with @Jack-Khuu , we can just remove this for now since it doesn't make sense to use a separate thread to spawn more training replicas.

joecummings: Same with the above code for threading trainer.

PR #241, opened 2025-09-26
Generalize the policy update integration test to sanity check on any model config (for Titan + vLLM)

JenniferWang:
Th idea is to generalize the existing sanity test is to answer the question "Given a model config `.yaml` file, is the weight transfer sharding correct?"

The PR essentially did the following
1. Generalize the test to accept `--config` parameter
2. Move existing (manually triggered) test case to the `artifacts` folder 

Testing
1. Make sure existing test cases pass as well as `apps/grpo/xx.yaml`
2. Manually changed the sharding logic (to make it incorrect) and verify the sanity check fails.


allenwang28:
> nit: I don't think the folder should be called artifacts

I think there should be a folder there already called `assets`, maybe we can use that?

PR #244, opened 2025-09-26
Add ComputeLobprob Tests

Jack-Khuu:
Moves `compute_logprobs` from GRPO app to `utils/ops.py` and adds basic tests for it

Also updated the logit slicing math, to an equivalent version based on TRL

Old
```
    context_length = logits.shape[1] - input_ids.shape[1]
    logits = logits[:, context_length - 1 : -1].to(input_ids.device)
```
New
```
    logits = logits[:, -input_ids.size(1) - 1 : -1, :].to(input_ids.device)
```

---
` python -m apps.grpo.main --config apps/grpo/qwen3_1_7b.yaml`

**wandb** (main): torchforge/grpo-training/runs/gq9wjime
image

**wandb** (PR): torchforge/grpo-training/runs/ctirsyoa
image

` pytest tests/unit_tests/util/test_compute_logprobs.py`

### DIFF
--- a/src/forge/util/ops.py
+++ b/src/forge/util/ops.py
@@ -66,8 +66,9 @@ def compute_logprobs(
             The temperature value for scaling logits before computing log probabilities.
 
     """
-    context_length = logits.shape[1] - input_ids.shape[1]
-    logits = logits[:, context_length - 1 : -1].to(input_ids.device)
+    # Ignore the last token from logits because it predicts the next token (-1)
+    # And align logits with the input tokens length.
+    logits = logits[:, -input_ids.size(1) - 1 : -1, :].to(input_ids.device)
     scaled_logits = logits / temperature
     logprobs = selective_log_softmax(scaled_logits, input_ids)
     return logprobs


src/forge/util/ops.py
```diff
@@ -49,3 +49,16 @@ def selective_log_softmax(logits: torch.Tensor, index: torch.Tensor) -> torch.Te
             per_token_logps.append(row_per_token_logps)
         per_token_logps = torch.stack(per_token_logps)
     return per_token_logps
+
+
+def compute_logprobs(
+    logits: torch.Tensor, input_ids: torch.Tensor, temperature: float = 1.0
+) -> torch.Tensor:
```
joecummings, line 56: Can you document the params in the docstring as well?

src/forge/util/ops.py
```diff
@@ -49,3 +49,16 @@ def selective_log_softmax(logits: torch.Tensor, index: torch.Tensor) -> torch.Te
...
+    """
+    Computes the log probabilities of the input tokens given the model logits and temperature.
+    """
+    context_length = logits.shape[1] - input_ids.shape[1]
```
joecummings: nit since were here: I think you can do `logits = logits[:, -input_ids.size(1): -1]` and not have to compute the context length above

Jack-Khuu: Uhhhh I'm gonna need to see the math on this, this doesn't smell right

joecummings: https://github.com/huggingface/trl/blob/417915a3e4d3e3bc8d7b196594308b8eabf928be/trl/trainer/grpo_trainer.py#L829

L829 -> L831

Jack-Khuu: Ok the math is looking less sus, but the loss doesn't seem to be happy with it OOTB so I think we make this a different PR instead of squeezing it in

**torchforge/grpo-training/runs/w5i55g7q** (current)
image

**torchforge/grpo-training/runs/b6sh1jnj**(TRL style)
image

casteryh:
Can you revive some of these tests? https://github.com/meta-pytorch/forge/blob/0e1984195c39816082c8d770f8186dfb6f4be764/tests/unit_tests/actors/test_reference_actor.py#L17-L98

src/forge/util/ops.py
```diff
@@ -49,3 +49,25 @@ def selective_log_softmax(logits: torch.Tensor, index: torch.Tensor) -> torch.Te
             per_token_logps.append(row_per_token_logps)
         per_token_logps = torch.stack(per_token_logps)
     return per_token_logps
+
+
+def compute_logprobs(
+    logits: torch.Tensor, input_ids: torch.Tensor, temperature: float = 1.0
+) -> torch.Tensor:
```
casteryh, line 56: can you add a warning saying that the prompt lengths have to be the same in the batch for this to work?

Jack-Khuu:
> Can you revive some of these tests?

Did not realize these tests still existed. yup will revive

src/forge/util/ops.py
Jack-Khuu: Nvn, math is indeed mathing converting to TRL style (little tweak)

PR #246, opened 2025-09-27
fix dcp for new weight update

casteryh:
- Now working for QWen3 8B. Update weights takes ~30 seconds. (~90 seconds total for a single step) https://meta.wandb.io/torchforge/grpo-training/runs/uacquboq?nw=nwuseryuxuanh
- Also added proper clean up of old checkpoints (#194)

Copilot generated summary follows:

This pull request introduces significant improvements to weight management in the GRPO training pipeline, focusing on more efficient handling, saving, and cleanup of model weights using torchstore and distributed checkpoints (DCP). The main changes include adding a mechanism to drop old weights, refactoring how weights are pushed and loaded (favoring DCP whole state dicts), and enhancing robustness with new utility functions and tests.

**Weight Management Improvements**

* Added the `drop_weights` async function in `apps/grpo/main.py` to efficiently delete old model weights after each training step, preventing unnecessary storage growth. This function uses new utilities to locate and drop both DCP and individual parameter keys. ([[1]](diffhunk://#diff-3bdea8452f8678efbb5410b0618ec2942e39e2678db862e9abf4c2fabf507fd8R247-R262), [[2]](diffhunk://#diff-3bdea8452f8678efbb5410b0618ec2942e39e2678db862e9abf4c2fabf507fd8R386-R387))
* Refactored the weight loading logic in `PolicyWorker.update` to prefer loading the entire state dict from a single DCP handle when available, falling back to individual parameters otherwise. This streamlines weight updates and reduces complexity. ([src/forge/actors/policy.pyL563-R582](diffhunk://#diff-831fe9863b096d39b6f902985aaee6e61a43223e546268648df082d801f495e1L563-R582))
* Updated the weight saving logic in `Trainer.push_weights` to save the whole state dict as a single DCP handle when `use_dcp` is enabled, improving performance and consistency. ([src/forge/actors/trainer.pyL347-R369](diffhunk://#diff-fbb948726cb719515c984b7d08bea55ac9c9e16590371b87bcf1501d168d2189L347-R369))

**Utility and Configuration Enhancements**

* Added new utility functions and improved the `DcpHandle` class in `src/forge/actors/_torchstore_utils.py`, including a robust `drop()` method to safely delete checkpoints and handle manifold storage cases. ([[1]](diffhunk://#diff-eee41791ef7899f9fab88c899921e9b958ebfe0c467d24f458a5de0e1e6e2411R6-R49), [[2]](diffhunk://#diff-eee41791ef7899f9fab88c899921e9b958ebfe0c467d24f458a5de0e1e6e2411R69-R72))
* Updated configuration files to enable built-in vllm loading and DCP usage for both policy and trainer components, aligning the pipeline with new weight management strategies. ([[1]](diffhunk://#diff-66b014532dd8c00aa471847f27c989d67c737ddc217b1843eb1818fab7d8e0cdL22-R22), [[2]](diffhunk://#diff-66b014532dd8c00aa471847f27c989d67c737ddc217b1843eb1818fab7d8e0cdL36-R37))

**Testing and Reliability**

* Introduced unit tests for the `DcpHandle.drop()` method to ensure proper deletion and cleanup behavior, including edge cases for manifold storage. ([tests/unit_tests/test_torchstore_utils.pyR1-R61](diffhunk://#diff-a61fe254a2096eb1a84ace6ccf290d7205e0c32d0a289f6763f17ab5475b65b5R1-R61))

These changes collectively make the training pipeline more efficient, reliable, and easier to maintain by improving how model weights are stored, loaded, and cleaned up.

### DIFF
--- a/src/forge/actors/_torchstore_utils.py
+++ b/src/forge/actors/_torchstore_utils.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 import logging
+import shutil
 from dataclasses import dataclass
 
 import torch
@@ -35,8 +36,6 @@ class DcpHandle:
             self.metadata = None
             self.param_names = None
             return
-
-        import shutil
 
         try:
             shutil.rmtree(self.checkpoint_id, ignore_errors=False)


apps/grpo/main.py
```diff
@@ -362,6 +384,8 @@ async def continuous_training():
                 mlogger.log("loss/training_step", loss, training_step)
                 await trainer.push_weights.fanout(training_step)
                 await policy.update_weights.fanout(training_step)
```
vidhyav, line 386: I am wondering if there is a possibility of needing "some" history of the weights. Can the RL loop be still alive after this statement finishes but the policy model goes down?

apps/grpo/main.py
```diff
@@ -239,6 +244,23 @@ async def pad_token(self):
         return self._tokenizer.pad_token_id
 
 
+async def drop_weights(version: int):
```
joecummings, line 247: Great function, but this is actually the kind of logic I don't care to see in the main.py file. Would it be possible to have this be part of torchstore / dcp itself (or a wrapper we write)? That way we can specify here "keep_last_n_weights".

apps/grpo/main.py
casteryh: > I am wondering if there is a possibility of needing "some" history of the weights. Can the RL loop be still alive after this statement finishes but the policy model goes down?

Can you elaborate on what you meant by "the RL loop be still alive"?

apps/grpo/main.py
casteryh: fair

Jack-Khuu:
nit: Can you format the clickable links in the PR description? Copilot borked it

casteryh:
> nit: Can you format the clickable links in the PR description? Copilot borked it

done!

src/forge/actors/_torchstore_utils.py
```diff
@@ -3,19 +3,50 @@
...
+            # Probably don't need to delete the checkpoint if it's on manifold
+            logger.warning(
+                f"Skipping deletion of {self.checkpoint_id} since it's on manifold"
+            )
+            self.checkpoint_id = None
+            self.metadata = None
+            self.param_names = None
+            return
+
+        import shutil
```
joecummings: You can import this at the top

src/forge/actors/_torchstore_utils.py
```diff
@@ -3,19 +3,50 @@
...
+
+        try:
+            shutil.rmtree(self.checkpoint_id, ignore_errors=False)
```
joecummings, line 41: Why do we want to suppress the errors here?

src/forge/actors/policy.py
```diff
@@ -560,28 +558,31 @@ async def update(self, version: int):
...
+            hf_param_names = [extract_param_name(key) for key in matching_keys]
+            # We can't pass a generator since vllm load_weights is not async.
+            # Instead, we just call load_weights with one parameter at a time.
+            for name in hf_param_names:
+                param_key = get_param_key(version, name)
+                param = await ts.get(param_key)
+                loaded = model.load_weights([(name, param)])
+                del param
+                loaded_weights.update(loaded)
         logger.info(
```
joecummings, line 586: The entire weight update timing is already calculated at the policy update level.

src/forge/actors/_torchstore_utils.py
casteryh: Fair. I was just thinking logging the error is fine and don't want to crash everything if delete is not successful. Let me know what you think.

src/forge/actors/policy.py
casteryh: yes, but it's different from each worker's update time. The entire updating time is basically the longest among the workers.

joecummings: Eventually, we might just want to remove the top-level logging time I suppose

joecummings:
we'll need #255 to land

PR #250, opened 2025-09-29
Reduce duplicate logs when spawning service replicas

DNXie:
Previously, `as_service` caused both a high-level service log and per-replica actor logs, which was noisy.
Now, `as_actor` takes a `_log` flag (default `True`) so we can suppress logs when it’s called internally by `Replica.initialize`.

### Before (service with 4 replicas)

```
service = await Counter.options(
        num_replicas=4,
        procs=1,
        health_poll_rate=0.5,
    ).as_service(v=20)
```
```
```

### After (service with 4 replicas)

```
INFO     forge.controller.actor:actor.py:123 Spawning Service for Counter
```

### Direct actor spawn (`as_actor`)

```
actor = await Counter.options(procs=1).as_actor(5)
```

```
INFO     forge.controller.actor:actor.py:208 Spawning single actor Counter
```

### Changes

* Added `_log` flag to `ForgeActor.as_actor` (default `True`).
* Replica initialization now calls `as_actor(..., _log=False)` to suppress redundant logs.
* Keeps logs for direct actor spawns, but ensures services only log once at the service level.

### DIFF
--- a/.github/workflows/unit_test.yaml
+++ b/.github/workflows/unit_test.yaml
@@ -28,13 +28,10 @@ jobs:
       - name: Install monarch
         run: python -m pip install monarch-no-torch==0.1.0.dev20250826 --find-links assets/ci
       - name: Install torchstore
-        run: |
-          eval "$(ssh-agent -s)"
-          ssh-add - <<< '${{ secrets.FORGE_GITHUB_CI_FOR_TORCHSTORE }}'
-          python -m pip install git+ssh://git@github.com/meta-pytorch/torchstore.git
+        run: pip install assets/wheels/torchstore-0.1.0-py3-none-any.whl
       - name: Install torchtitan
         run: |
-          pip install --pre torchtitan==0.1.0.dev20250826+cpu --extra-index-url https://download.pytorch.org/whl/nightly/cpu
+          pip install assets/wheels/torchtitan-0.1.0-py3-none-any.whl
           pip install tyro
       - name: Install dependencies
         run: python -m pip install --no-build-isolation -e ".[dev]"
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -317,11 +317,10 @@ async def main(cfg: DictConfig):
     max_req_tokens = cfg.max_req_tokens
     max_res_tokens = cfg.max_res_tokens
 
-    # TODO: delete this logic after we are confident on the vllm weight sync long term fix PR #184
-    policy_tp_size = cfg.policy.engine_config.tensor_parallel_size
-
     # initialize before spawning services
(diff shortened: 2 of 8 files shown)


src/forge/controller/actor.py
```diff
@@ -120,7 +120,7 @@ async def as_service(
         }
         cfg = ServiceConfig(**cfg_kwargs)
 
-        logger.info("Spawning Service Actor for %s", cls.__name__)
+        logger.info("Spawning Service for %s", cls.__name__)
```
joecummings, line 123: nit: since we have a lowercase "actor" logged elsewhere, we should lowercase "Service" as well.
```suggestion
        logger.info("Spawning service for %s", cls.__name__)
```

src/forge/controller/actor.py
```diff
@@ -196,15 +196,16 @@ async def launch(cls, *args, **kwargs) -> "ForgeActor":
         return actor
 
     @classmethod
-    async def as_actor(cls: Type[T], *args, **actor_kwargs) -> T:
+    async def as_actor(cls: Type[T], *args, _log: bool = True, **actor_kwargs) -> T:
```
joecummings: Just curious - why does this come in between *args, and **kwargs? 

Also can we call it "quiet"? If quiet=True, no log. Else log. The default can be quiet=False

DNXie: >Just curious - why does this come in between *args, and **kwargs?

Good question! I just learnt today from chatgpt: Anything defined between them must be passed as a keyword argument only. If we define like this:
```
async def as_actor(cls, _log: bool = True, *args, **kwargs):
```
then `_log` could be set positionally like
```
await MyActor.as_actor(False, "foo")  # "foo" is *args
```
which is very ambiguous. Although we wouldn't really allow users to pass this, but it would be nice to keep `_log` as keyword argument. 

> Also can we call it "quiet"?

Good suggestion! Updated

src/forge/controller/actor.py
```diff
@@ -196,15 +196,16 @@ async def launch(cls, *args, **kwargs) -> "ForgeActor":
...
+    async def as_actor(cls: Type[T], *args, _quiet: bool = False, **actor_kwargs) -> T:
         """
         Spawns a single actor using the configuration stored in `.options()`, or defaults.
 
         The configuration values stored in the subclass returned by `.options()` (like
         `procs`) are used to construct a ProcessConfig instance.
         If no configuration was stored, defaults to a single process with no GPU.
         """
-        logger.info("Spawning single actor %s", cls.__name__)
+        if not _quiet:
```
allenwang28: we don't need this at all if `replica` just uses `launch()` right?

https://github.com/meta-pytorch/forge/blob/main/src/forge/controller/service/replica.py#L164

DNXie: Right! Done!

PR #256, opened 2025-09-29
Update Float8 to Float8Dense

joecummings:
See https://github.com/pytorch/torchtitan/pull/1770


joecummings:
Unit tests will fail until tomorrow when the new nightly gets built (horrible system, need to fix), but this should land so @Ritesh1905 can kick off a training run.

PR #259, opened 2025-09-30
Fix call in trainer + comment out drop_weights

joecummings:
Trainer was converted back to an actor and can have more than one actor so should be a call.

Commented out drop_weights b/c there's an error and I don't want main to be broken.


vidhyav:
This means we are not gc'ing the weights? (I know this is temporary).

joecummings:
> This means we are not gc'ing the weights? (I know this is temporary).

Correct - @casteryh will fix today.

PR #263, opened 2025-09-30
init backend can happen before service init

felipemello1:


joecummings:
Why?

felipemello1:
> Why?

it was a requirement before, i.e. you spawn the actor and **only then** you have to call init backend.  I fixed it so its robust and you just need to call it once, anywhere, and subsequent spawning will work.

joecummings:
> > Why?
> 
> it was a requirement before, i.e. you spawn the actor and **only then** you have to call init backend. I fixed it so its robust and you just need to call it once, anywhere, and subsequent spawning will work.

Neato

Issue #251, opened 2025-09-29
potential corrupted state in vllm during weight updates


casteryh:
https://github.com/meta-pytorch/forge/blob/516abf0e9270aedd65ac9425eab949eb802aada5/src/forge/actors/policy.py#L391-L396
In this method, only current requests are being awaited. However, while waiting, other generate() requests will keep coming, resulting these requests to be processed *WHILE* the weight update is ongoing.

Another issue is `reset_prefix_cache()` ([vllm doc](https://docs.vllm.ai/en/v0.9.2/api/vllm/v1/core/kv_cache_manager.html#vllm.v1.core.kv_cache_manager.KVCacheManager.reset_prefix_cache) needs to be called after weight update is finished. See similar issuse in TRL (https://github.com/huggingface/trl/issues/3024)

### Block new `generate()` request from being processed when the weight update is initiated.
- option 1: (add a flag) this can be done by adding a `self.halt_requested` attribute to the `Policy` actor. 
  - And we can do something like
  ```python
  async def generate(...):
      while self.halt_requested:
          asyncio.sleep(1)
      pass
  ```
   - risk: there still might be subtle bugs if some operations are not atomic
 - option 2: (rwlock) add a writer first rwlock in Policy, and obtain the write lock before updating weight, obtain the read lock before generate.
 ### Either way, we need to refactor the `generate()` method so that it returns completion along with the policy version, because when calling the `generate()` method, the policy might be at `v0`, and update is initiated, the request won't be processed until the update to `v1` is completed. Therefore, calling `policy_version()` is not a reliable to find which version is the response generated with.

Jack-Khuu:
> refactor the generate() method so that it returns completion along with the policy version

It already does this? https://github.com/meta-pytorch/forge/pull/234
Or do you mean track the version when scheduled (before not processed) and compare that to when it is processed (later is what is returned in Completion)

casteryh:
> It already does this? [#234](https://github.com/meta-pytorch/forge/pull/234)

Ah you are right!
> Or do you mean track the version when scheduled (before not processed) and compare that to when it is processed (later is what is returned in Completion)

No I just missed that it was already added! But some requests are still being process while update is ongoing. (like the first n tokens is in `v0` and the rest is generated with `v1`. And currently this will be incorrectly as marked as being generated by `v1`.

Jack-Khuu:
We solved this a while back with a flag and an auxillary queue, but pulled it out. 

I'll add it back

[...]

PR #264, opened 2025-09-30
Add locking to request queue while weight updates

Jack-Khuu:
Introduces a simple inline locking scheme for Policy Generation and Weight updates
* `request_lock` acts as the guard for checking and updating `accepting_requests`
* `update_lock` acts as the guard to ensure only one update job runs at a time

Also adds missing `self.scheduler.kv_cache_manager.reset_prefix_cache()` call after weight updates

See https://github.com/meta-pytorch/forge/issues/251 for more details

---
Tests

`pytest -s tests/integration_tests/test_policy_update.py::TestWeightSync::test_sanity_check --config apps/grpo/qwen3_8b.yaml`

`python -m apps.grpo.main --config apps/grpo/qwen3_1_7b.yaml`

---
**Wandb**
I wouldn't over analyze the magnitude as there can be drastic swings, but the trend is clearly better

On main: grpo-training/runs/8u6rq5r1
image

On PR: grpo-training/runs/hq33z9su
image


Jack-Khuu:
I assume you're referencing `asyncio.Lock()`? 

It only provides an exclusive mutex, whereas this lets multiple generation requests run without blocking (until a update is involved)

tests/unit_tests/util/test_lock.py
```diff
@@ -0,0 +1,289 @@
...
+
+        await lock.acquire()
+        assert lock._readers == 1
+
+        await lock.release()
+        assert lock._readers == 0
+
+    @pytest.mark.timeout(10)
+    @pytest.mark.asyncio
+    async def test_multiple_readers(self):
```
casteryh: I think testing for behaviors only is better.
It's unnecessary to test implementations (and doing so often gives false confidence). Plus it also makes it harder to make changes in the future.

tests/unit_tests/util/test_lock.py
```diff
@@ -0,0 +1,289 @@
...
+        assert lock._exclusive is False
+
+    @pytest.mark.timeout(10)
+    @pytest.mark.asyncio
+    async def test_exclusive_lock_blocks_readers(self):
+        """Test that exclusive lock blocks new readers."""
+        lock = Lock()
+
+        await lock.acquire_exclusive_lock()
+        assert lock._exclusive is True
```
casteryh: same

tests/unit_tests/util/test_lock.py
```diff
@@ -0,0 +1,289 @@
...
+                await lock.acquire()
+                await lock.release()
+
+        # Run multiple tasks that rapidly acquire/release
+        tasks = [asyncio.create_task(rapid_acquire_release(10)) for _ in range(5)]
+
+        await asyncio.gather(*tasks)
+
+        # Final state should be clean
+        assert lock._readers == 0
```
casteryh: same as above, I don't think testing for implementation details is a good idea.

src/forge/actors/policy.py
```diff
@@ -430,6 +440,9 @@ async def run(self):
 
     @endpoint
     async def update_weights(self, policy_version: int):
+        # Prevent new requests from being queued while letting in-flight requests queue
+        await self.requests_lock.acquire_exclusive_lock()
```
JenniferWang: Can we talk about a potential ideal behavior?

1. There's a notion of blocking requests when the Policy server's current policy version is too old or the weight tensors (GPU) is being swapped to the next version. 

3. There's a separation of "downloading tensors to CPU" and "loading from CPU to GPU". Today these two steps are done in this single `update_weights` function. Downloading tensors to CPU can be done simultaneously as serving requests.

From simple profiler run, "downloading tensors to CPU" takes up > 80% of the time in the `update_weights` step.

JenniferWang: @joecummings , @LucasLLC , @casteryh

src/forge/util/lock.py
```diff
@@ -0,0 +1,48 @@
...
+
+    async def release(self):
+        async with self._cond:
+            if self._readers == 0:
+                raise RuntimeError("Cannot release an unacquired lock")
+            self._readers -= 1
+            if self._readers == 0:
+                self._cond.notify_all()
+
+    async def acquire_exclusive_lock(self):
```
casteryh: I have no problem with this: just curious, why is it called `exclusive` vs `non-exclusive` as apposed to `write` vs `read`.

src/forge/util/lock.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Lock:
+    def __init__(self):
+        self._cond = asyncio.Condition()
+        self._readers = 0
+
+        # Used to indicate that an exclusive lock is held or requested
+        self._exclusive = False
+        self._exclusive_waiters = 0
+
```
casteryh: would be nice to implement an async context manager, so you can
```python
async with lock.read():
    pass
```

src/forge/actors/policy.py
```diff
@@ -243,6 +245,8 @@ async def setup(self):
         self.request_id = 0
         self.policy_version = 0
         self.requests: dict[str, tuple[None | ParentRequest, asyncio.Future]] = {}
```
JenniferWang, line 246: Not directly related to this diff... I actually think we should remove this -- each request is tracked inside the `self.scheduler`

src/forge/util/lock.py
```diff
@@ -0,0 +1,48 @@
...
+    async def acquire(self):
+        async with self._cond:
+            while self._exclusive or self._exclusive_waiters > 0:
```
casteryh: I see you are trying to prevent writer starving here, can you add a test that verifies writers indeed won't be starved.

src/forge/util/lock.py
```diff
@@ -0,0 +1,48 @@
...
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class Lock:
+    def __init__(self):
+        self._cond = asyncio.Condition()
+        self._readers = 0
+
+        # Used to indicate that an exclusive lock is held or requested
```
casteryh: nit on the comment: this actually indicates the lock is *held* right?

src/forge/util/lock.py
```diff
@@ -0,0 +1,48 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import asyncio
+import logging
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class Lock:
```
casteryh: Can you change the name or add a docstring saying this is a write-preferring read-write lock?

allenwang28:
for the locking logic can we do something simpler? I am not convinced we need to introduce our own lock 

```
class Policy:
    def __init__(self):
        self.active_requests = 0
        self.accepting_requests = True
        self.state_lock = asyncio.Lock() # guards accepting_requests
        self.update_lock = asyncio.Lock() # only one update at a time

    async def generate(...):
        while True: # when we get a request, make sure we can accept and spin if not
            async with self.state_lock:   
                if self.accepting_requests:
                    self.active_requests += 1
                    break
             await asyncio.sleep(0.1)

        try:
            # process request
            ...
        finally:
            async with self.state_lock:
                self.active_requests -= 1

    async def update(self):
        # trigger weight download to CPU
        weight_prefetch = asyncio.task(self.policy_workers.prefetch_weights.call())

        # generate should not accepting anything new
        async with self.update_lock:
            async with self.state_lock:
                self.accepting_requests = False

        # spin for all requests to complete
        while True:
            async with self.state_lock:
                if self.active_requests == 0:
                    break
            asyncio.sleep(0.1)

        # finally do the actual weight update
        await weight_prefetch # don't do the update until we're sure that the weights are on CPU
        await self.policy_workers.update.call()

        # afterwards we can resume accepting requests again
        async with self.state_lock:
            self.accepting_requests = True
```

src/forge/actors/policy.py
casteryh: To your point 1: I think these are separate problems we need to solve

> 1. There's a notion of blocking requests when the Policy server's current policy version is too old

This is about request cancellation and gracefully handle cancellations in rollouts.

> or the weight tensors (GPU) is being swapped to the next version.

This is the current PR, trying to prevent corrupted states.



> 2. There's a separation of "downloading tensors to CPU" and "loading from CPU to GPU". Today these two steps are done in this single `update_weights` function. Downloading tensors to CPU can be done simultaneously as serving requests.

It would be good to have downloading tensors to CPU a separate function and only acquire the lock when actually move them to CPU. But that requires downloading the tensors to CPU all at once (which is what we are doing already and probably won't be a problem)

src/forge/actors/policy.py
JenniferWang, line 153: `Policy` is effectively the controller thread for a replica. I don't quite understand the need for a lock. Can we repurpose this state boolean?

src/forge/actors/policy.py
```diff
@@ -430,6 +440,9 @@ async def run(self):
 
     @endpoint
     async def update_weights(self, policy_version: int):
+        # Prevent new requests from being queued while letting in-flight requests queue
+        await self.requests_lock.acquire_exclusive_lock()
+
         # TODO: If generating long sequences, this might be long and will block policy weight updates
         curr_requests = [fut for _, fut in self.requests.values()]
```
casteryh: theoretically, after acquiring the lock, this should always be empty?

casteryh:
> for the locking logic can we do something simpler? I am not convinced we need to introduce our own lock
> 
> ```
> class Policy:
>     def __init__(self):
>         self.active_requests = 0
>         self.accepting_requests = True
>         self.state_lock = asyncio.Lock() # guards accepting_requests
>         self.update_lock = asyncio.Lock() # only one update at a time
> 
>     async def generate(...):
>         while True: # when we get a request, make sure we can accept and spin if not
>             async with self.state_lock:   
>                 if self.accepting_requests:
>                     self.active_requests += 1
>                     break
>              await asyncio.sleep(0.1)
> 
>         try:
>             # process request
>             ...
>         finally:
>             async with self.state_lock:
>                 self.active_requests -= 1
> 
>     async def update(self):
>         # trigger weight download to CPU
>         weight_prefetch = asyncio.task(self.policy_workers.prefetch_weights.call())
> 
>         # generate should not accepting anything new
>         async with self.update_lock:
>             async with self.state_lock:
>                 self.accepting_requests = False
> 
>         # spin for all requests to complete
>         while True:
>             async with self.state_lock:
>                 if self.active_requests == 0:
>                     break
>             asyncio.sleep(0.1)
> 
>         # finally do the actual weight update
>         await weight_prefetch # don't do the update until we're sure that the weights are on CPU
>         await self.policy_workers.update.call()
> 
>         # afterwards we can resume accepting requests again
>         async with self.state_lock:
>             self.accepting_requests = True
> ```

@allenwang28 I think you just implemented a rwlock inside the policy lol.

allenwang28:
> @allenwang28 I think you just implemented a rwlock inside the policy lol

lol fair enough, do we see ourselves needing this more than once though?

src/forge/util/lock.py
Jack-Khuu: Former is held, latter are requests

I'll clarify

src/forge/util/lock.py
Jack-Khuu: Ah so because the way this is written the "read locks" are intentionally instantly released after incrementing the count. This way we we can have multiple "reads" (generation requests)

src/forge/util/lock.py
Jack-Khuu: Because the generation step isn't really "reading", it's managing with the counter

tests/unit_tests/util/test_lock.py
Jack-Khuu: I'll lighten up on some of the checks

casteryh:
> > @allenwang28 I think you just implemented a rwlock inside the policy lol
> 
> lol fair enough, do we see ourselves needing this more than once though?

Maybe not. But I actually like this better simply because of testability. otherwise everything has to be tested in an integration test.

src/forge/actors/policy.py
joecummings: @JenniferWang  To make sure I'm clear, "downloading tensors to CPU" means loading tensors from TorchStore / DCP file(s) into CPU memory?

Jack-Khuu: > blocking requests when the Policy server's current policy version is too old

Handling cancellations is def something we should revisit since we just let the in flight requests (which are over generations for an old policy given we are updating). 

It doesn't affect the correctness, but is inefficient (separate PR/discussion)

> Downloading tensors to CPU can be done simultaneously as serving requests. From simple profiler run, "downloading tensors to CPU" takes up > 80% of the time in the update_weights step.

Oh wow that's a crazy chunk. Definitely something we should split up 


Edit: To clarify, I love everything you brought up. And we should tackle them after we get correctness fixed

Jack-Khuu:
> @allenwang28 I think you just implemented a rwlock inside the policy lol

Lol yuxuan beat me to it. Doing it inline also makes reading the implementation a lot clunkier since the logic is now split between N-functions

> lol fair enough, do we see ourselves needing this more than once though?

If we want it inside `policy.py` we can flatten it (track the lock as a member like in your example) and hide the locking as helper functions. Renaming the functions to be case specific

`acquire_exclusive_lock` => `pause_new_requests()`
`release_exclusive_lock` => `resume_processing()`

src/forge/actors/policy.py
Jack-Khuu: If we inline the Lock we can definitely try reusing `self.running`. 

> need for a lock

Without a lock we can't track when requests that are actively being generated on the old policy have all returned (before we start the upgrade)

src/forge/actors/policy.py
Jack-Khuu: I'm gonna spin this into an issue since this is a good discussion

src/forge/util/lock.py
casteryh: > Because the generation step isn't really "reading", it's managing with the counter

yes. but aren't we protecting the weights instead of the counter? so it is a reader in this sense?

src/forge/util/lock.py
casteryh: > Ah so because the way this is written the "read locks" are intentionally instantly released after incrementing the count. This way we we can have multiple "reads" (generation requests)

Exactly!

But I don't see how this is contradictory to `read()` returns an async context manager?

src/forge/util/lock.py
Jack-Khuu: Functionally yes

src/forge/util/lock.py
Jack-Khuu: Got it, yeah I had a version of this that was context manager based but let's do that after

src/forge/actors/policy.py
Jack-Khuu: Moved before lock; Kept for logging

allenwang28:
tbh I think the implementation being clunky is fine. We're only doing this once, and the logic between `update()` and `generate()` is fundamentally complex - there's no getting around that.

My concern is around long term maintenance. We're not just implementing a lock - we're introducing a coordination primitive with very specific semantics: "pause accepting new work, drain in-flight work, execute weight sync, resume". Getting this right, handling edge cases (cancellation, timeouts, etc.), and maintaining it over time is non-trivial.

There are cases where it makes sense to own custom primitives (like our RL-specific logging patterns), but I don't think this is one of them. And once we have a CustomRWLock class, people will want to add features, and suddenly we're maintaining a mini-framework.

If we **truly** need an rwlock abstraction, I'd prefer using something battle-tested like aiorwlock. Yes, it's an extra dependency, but it prevents us from owning this complexity long-term.

I also understand the testability argument, but the integration is what really matters here. Once we have GPU tests running, I want this specific weight sync tested as an integration test regardless - whether we have smaller unit tests or not.

The reason I prefer inline: you can see exactly what's happening. With an abstraction, you have to jump to another file, understand the API, and mentally map it back to this context. Given the complexity of this coordination, that directness has real value.

I promise I'm not trying to be difficult here - you bring up great, logical points - but this is an area I think we need to be really careful.

src/forge/actors/policy.py
Jack-Khuu: Added as TODO

apps/grpo/qwen3_8b.yaml
```diff
@@ -127,6 +127,10 @@ services:
     procs: 1
     num_replicas: 1
     with_gpus: false
+  trainer:
```
Jack-Khuu: Seems like this got removed by accident?

Jack-Khuu:
Updated to inlined all the locks

apps/grpo/qwen3_8b.yaml
allenwang28: trainer should be an actor now

src/forge/actors/policy.py
Jack-Khuu: https://github.com/meta-pytorch/forge/issues/279

src/forge/actors/policy.py
```diff
@@ -332,6 +342,10 @@ async def generate(self, prompt: str, priority: int = 0) -> list[Completion]:
         )
         t.step("process_inputs")
 
+        # Grab the lock to check if we are accepting requests
+        async with self.request_lock:
+            await self.request_lock.wait_for(lambda: self.accepting_requests)
```
allenwang28, line 349:
```suggestion
            # Wait until we're accepting requests (releases lock while waiting)
            # If accepting_requests is True, continue immediately (holding the lock)
            # If False, release lock, wait for notification, re-acquire and recheck
            await self.request_lock.wait_for(lambda: self.accepting_requests)
```

src/forge/actors/policy.py
```diff
@@ -428,34 +442,56 @@ async def run(self):
                     _, fut = self.requests.pop(request_output.request_id)
                     fut.set_result(completions)
 
+            # Grab the lock to notify potential update requests
```
allenwang28:
```suggestion
            # Notify waiters if queue is drained
```

src/forge/actors/policy.py
```diff
@@ -428,34 +442,56 @@ async def run(self):
...
-                Reduce.MEAN,
-            )
-            record_metric(
-                "policy_perf/update_weights/max_pending_requests",
-                len(curr_requests),
-                Reduce.MAX,
-            )
-            logger.debug(f"Waiting for {len(curr_requests)} pending requests")
-            await asyncio.gather(*curr_requests)
+        # Grab the lock to be the only one updating weights
```
allenwang28:
```suggestion
        # Serialize updates (only one update at a time)
```

src/forge/actors/policy.py
```diff
@@ -428,34 +442,56 @@ async def run(self):
...
+                    record_metric(
+                        "policy_perf/update_weights/max_pending_requests",
+                        len(curr_requests),
+                        Reduce.MAX,
+                    )
+                    logger.debug(f"Waiting for {len(curr_requests)} pending requests")
 
-        # Record weight update metrics
-        record_metric("policy/update_weights/count_weight_updates", 1, Reduce.SUM)
+                # TODO: If generating long sequences, this might be long and will block
```
allenwang28, line 476:
```suggestion
                # Wait until all pending requests have been processed
                # TODO: If generating long sequences, this might be long and will block
```

src/forge/actors/policy.py
```diff
@@ -428,34 +442,56 @@ async def run(self):
...
+            if self.use_vllm_builtin_load:
+                await self.policy_worker.update.call(version=policy_version)
+            else:
+                await self.policy_worker.update_DEPRECATED.call(version=policy_version)
+            self.policy_version = policy_version
+
+            # After updating the weights, we need to reset the KV cache
+            self.scheduler.kv_cache_manager.reset_prefix_cache()
+
+        # Grab the lock to notify resume accepting requests and wake up pending requests
```
allenwang28:
```suggestion
        # Resume accepting requests and wake up any waiting generate() calls
```

src/forge/actors/policy.py
allenwang28: Should the `self.requests[request_id] = ...` add be done within this lock?

A race condition to consider: 
- `generate()` releases the lock before adding to self.requests
- therefore a request could pass `accepting_requests`, then `update` could set `accepting_requests=False` and see an empty queue
- but then `generate()` adds to the queue while the update is already running

Jack-Khuu: Good catch, you're right a new update could kick off otherwise

Jack-Khuu:
Runs outputs (and wandb) for GRPO app are updated in the description

PR #266, opened 2025-09-30
Update make.bat

jspisak:
Added license header to this file


docs/make.bat
```diff
@@ -33,3 +39,4 @@ goto end
 
 :end
 popd
+
```
dcci, line 40: Please remove this line before merging.

PR #271, opened 2025-09-30
Update README.md

init27:
made small updates to top messaging, added links and some caveats

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 #### A PyTorch native agentic library for RL post-training and agentic development that lets you focus on algorithms instead of writing infra code.
 
 ## Overview
-Forge was built with one core principle in mind: researchers should write algorithms, not infrastructure. Forge introduces a “service”-centric architecture that provides the right abstractions for distributed complexity and workloads. When you need fine-grained control over placement, fault handling/redirecting training loads during a run, or communication patterns, the primitives are there. When you don’t, you can focus purely on your RL algorithm.
+The primary purpose of the Forge ecosystem is to delineate infra concerns from model concerns thereby making RL experimentation easier. Forge delivers this by providing clear RL abstractions and one scalable implementation of these abstractions. When you need fine-grained control over placement, fault handling/redirecting training loads during a run, or communication patterns, the primitives are there. When you don’t, you can focus purely on your RL algorithm.
 
 Key features:
 - Usability for rapid research (isolating the RL loop from infrastructure)
--- a/.gitignore
+++ b/.gitignore
@@ -199,3 +199,4 @@ assets/wheels/vllm*.whl
 # DCP artifacts
 model_state_dict/
 forge_dcp_tmp/
+demo_top_down.md


README.md
```diff
@@ -18,15 +20,19 @@ Key features:
 > work. It's recommended that you signal your intention to contribute in the
 > issue tracker, either by filing a new issue or by claiming an existing one.
 
-## 📖 Documentation
+## 📖 Documentation (Coming Soon)
+
+View Forge's hosted documentation [at this link (coming soon)](https://meta-pytorch.org/forge/).
```
dcci: you might remove the link until it's operational

README.md
```diff
@@ -18,15 +20,19 @@ Key features:
...
 
-View Forge's hosted documentation [at this link](https://meta-pytorch.org/forge/).
+You can also find our notebook tutorials below (coming soon)
 
 ## Installation
 
 ### Basic
 
-Forge requires the latest PyTorch nightly with Monarch, vLLM, and torchtitan. For convenience,
+Forge requires the latest PyTorch nightly with [Monarch](https://github.com/meta-pytorch/monarch), [vLLM](https://pytorch.org/blog/pytorch-vllm-♥%EF%B8%8F/), and [torchtitan](https://github.com/pytorch/torchtitan). For convenience,
```
dcci: why a pointer to the blogpost for vLLM rather than the repo?

README.md
```diff
@@ -40,7 +46,7 @@ conda activate forge
 
 Optional: By default, the packages installation uses conda. If user wants to install system packages on the target machine instead of conda, they can pass the `--use-sudo` to the installation script: `./script/install.sh --use-sudo`.
 
-After install, you can run the following command and should see output confirming GRPO training is running (you need a minimum 3 GPU devices).
+After install, you can run the following command and should see output confirming GRPO training is running (you need a minimum 3 GPU devices), right now this will run Full Fine-Tuning:
```
dcci: Not sure about this change, what do you mean?

README.md
```diff
@@ -1,10 +1,12 @@
 # <img width="35" height="35" alt="image" src="https://github.com/user-attachments/assets/2700a971-e5d6-4036-b03f-2f89c9791609" /> Forge
 
 
-#### A PyTorch native agentic library for RL post-training and agentic development
+#### A PyTorch native agentic library for RL post-training and agentic development that lets you focus on algorithms instead of writing infra code.
```
dcci, line 4: not sure you want to make this too long

README.md
init27: I wanted to point out that atm we only support FFT and no LoRA

README.md
init27: I wanted to add nod to infra in headline but will leave it to your call, we can revert if you feel old was better

README.md
ebsmothers: I think it might be a bit confusing to mention full finetuning here. I think I understand what you're getting at (that GRPO can be done with LoRA too but we don't yet support it), but I think readers might not get this point and may instead confuse with SFT.

README.md
```diff
@@ -1,10 +1,12 @@
...
+
+> Note: We also support Supervised Fine-Tuning and what you might expect from [TorchTune](https://github.com/meta-pytorch/torchtune/issues/2883)
```
ebsmothers: Personally I would leave this out. I worry that it will lead folks to expect a 1:1 mapping between the two offerings and open up a ton of questions like "QLoRA/activation offloading/linear cross entropy/some other feature is supported in torchtune, why isn't it here?"

README.md
HamidShojanazeri: maybe "Forge — a PyTorch library that abstracts infrastructure to simplify RL Post-training" or "a PyTorch library for streamlined agentic RL post-training, with clear abstractions and built-in infra." cc @dcci

init27: Thanks Hamid. Personally I want to choose first one, Davide any opinions please?

jspisak: I like the abstract infra language personally - our audience is RL researchers with a goal of enabling them to quickly explore the space. Not necessarily make people infra experts.

jspisak: also note that we will have an agentic API so I don't want to tie it too much to just post training which is why i worded the original language the way I did.

PR #272, opened 2025-09-30
Updates all wheels and removes `await`

allenwang28:
- Moves all the pins
- Monarch breaks now if you try and `await` on an actor spawn. Changes all APIs for that.


scripts/build_wheels.sh
joecummings, line 1: WHY NEW WHEELS FOR TITAN AND STORE IF NOT UPDATED COMMITS HERE?

src/forge/actors/policy.py
```diff
@@ -194,7 +194,7 @@ async def launch(  # pyright: ignore[reportIncompatibleMethodOverride]
         vllm_config = engine_config.create_vllm_config()
         # TODO (felipemello): LocalFetcherActor doesnt spawn with this, so cannot
         # do logging within PolicyWorker
-        workers = await worker_procs.spawn(
+        workers = worker_procs.spawn(
```
joecummings, line 197: Why can you not await here?

allenwang28: new Monarch fun

scripts/build_wheels.sh
allenwang28: reverted

PR #273, opened 2025-09-30
Add Sphinx gallery and fixes to the doc build

svekars:
This pull request updates documentation build system and structure for torchforge
- updates to conf.py
- improved deployment logic for versioned docs, update to in GitHub Actions
- Added template_tutorial.py and config for sphinx-gallery


docs/source/conf.py
```diff
@@ -127,6 +165,88 @@
     "exclude-members": "__weakref__",
 }
 
+# Mock imports for dependencies that aren't available during doc building
```
allenwang28: just for my understanding, once we have wheels that work in CI then we can go ahead and remove all of this right?

allenwang28 (approved):

docs/source/conf.py
svekars: I think it might be problematic for a local installation that doesn't have all these deps more than the CI since we are installing all this. Also, just for API docs these import might not be necessarily needed. Since I'm only adding doc build changes and sphinx gallery in this PR, I will delete this part for now.

PR #239, opened 2025-09-25
metric logger - grpo metrics (3/N)

felipemello1:
Adds metrics to GRPO. Also adds conftest.py to ignore record_metrics for unit testing.

```
WandbBackend: Logged 86 metrics at step 1
=== [global_logger_fF8h_r0] - METRICS STEP 1 ===
  buffer/add/count_episodes_added: 24.0
  buffer/evict/avg_policy_staleness: 0.0
  buffer/evict/max_policy_staleness: 0.0
  buffer/sample/avg_buffer_utilization: 16.0
  buffer/sample/avg_buffer_utilization_pct: 100.0
  buffer/sample/count_sample_requests: 239.0
  buffer_perf/sample/duration_avg_s: 5.8068605490558815e-05
  buffer_perf/sample/duration_max_s: 0.004135082941502333
  dataset/sample/avg_sample_len: 525.25
  dataset/sample/count_samples_generated: 4.0
  main/rollout/avg_advantage: -7.450580596923828e-09
  main/rollout/count_rollout_iterations: 3.0
  main/rollout/std_advantage: 0.9332931327148772
  main/training/count_training_iterations: 1.0
  main_perf/rollout/advantage_computation/duration_avg_s: 0.05832476494833827
  main_perf/rollout/advantage_computation/duration_max_s: 0.06426444882526994
  main_perf/rollout/compute_logprobs/duration_avg_s: 0.20812372273455063
  main_perf/rollout/compute_logprobs/duration_max_s: 0.24763409001752734
  main_perf/rollout/data_loading/duration_avg_s: 0.145997637261947
  main_perf/rollout/data_loading/duration_max_s: 0.43461281480267644
  main_perf/rollout/policy_generation/duration_avg_s: 3.4080680563735464
  main_perf/rollout/policy_generation/duration_max_s: 4.955212695058435
  main_perf/rollout/reference_model_forward/duration_avg_s: 11.607943433026472
  main_perf/rollout/reference_model_forward/duration_max_s: 12.128578653093427
  main_perf/rollout/reward_evaluation/duration_avg_s: 0.010726675235976776
  main_perf/rollout/reward_evaluation/duration_max_s: 0.017631276044994593
  main_perf/rollout/total_duration_avg_s: 15.439195216012498
  main_perf/rollout/total_duration_max_s: 17.847949991934
  policy/generate/avg_tokens_generated: 508.90625
  policy/generate/count_generate_requests: 4.0
  policy/generate/count_requests: 4.0
[...]

PR #275, opened 2025-10-01
hotfix - disable time cuda

felipemello1:
TLDR: 
commit right before metric logging -> No OOM
[metric logging](https://github.com/meta-pytorch/forge/pull/239) --> OOM
Disabling cuda stream for time tracking -> No OOM

Hotfix disables time_with_cuda everywhere in forge until further testing. For now it will track time using only cpu, which is not as precise as cuda.

-----

Currently when running `python -m apps.grpo.main --config apps/grpo/qwen3_8b.yaml` we get OOM on the second train step during [loss computation](https://github.com/meta-pytorch/forge/blob/67cf003a3408d49c82caa53d853b8835cbaee8c4/src/forge/actors/trainer.py#L262).

(1) The [metric logging PR](https://github.com/meta-pytorch/forge/pull/239) introduced this bug. One of the functionalities is to set 'timer="gpu"', which uses cuda streams to capture cuda time without blocking.
(2) Forcing it to always use `timer="cpu"` solved the issue.
(3) Upon investigation, I was not able to detect or produce any sort of memory leakage. However, I realized that we have a possible issue with memory fragmentation:
 
For an 8B model on bf16, we expect:
model = 16GBGB 
model + gradients = 32GB 
model + gradients + optimizer state = 64GB 

We should see 32 extra GB at the init of the 2nd forward because now we have optimizer states.

Indeed we see exactly that
```
| Iter/Phase    | Alloc (GB) | Reserved (GB) | Peak (GB) | Frag Signal? |
|---------------|------------|---------------|-----------|--------------|
| 1st: Pre-fwd | 16.4       | 16.4          | 16.4      | model = 16GBGB |
| 1st: Post-fwd | 32.8      | 82.6          | 63.0      | model + gradients = 32GB |
| 2nd: Pre-fwd | 49.2       | 82.6          | 49.2      | model + optimizer state = 48GB |
| 2nd: Post-fwd | 65.6      | 76.8-87.5     | 95.8      | model + gradients + optimizer state = 64GB  |
``````

However, reserved memory is much higher than allocated. We also see a peak~= 95GB, close to max gpu memory of 100GB.

(4) A common solution for memory fragmentation is to use [`expandable_segments`](https://github.com/meta-pytorch/forge/blob/67cf003a3408d49c82caa53d853b8835cbaee8c4/src/forge/actors/trainer.py#L154), which we do, but we set it after importing torch. AFAIK, it doesnt work to in this order and, instead, we have to use `torch.cuda.memory._set_allocator_settings("expandable_segments:True")`

(5) The cuda stream to capture time apparently allocates a bit of memory, but given the GPU limit + fragmentation, it OOMs. This did not happen for qwen 1.7B, for example after >20 steps of training.

-----

Next steps:
- Check if expandable segments solves it
- See if we can reduce impact from cuda stream
- Maybe we are really at the complete peak gpu utilization, and it cannot even tolerate an extra cuda stream.

### DIFF
--- a/src/forge/observability/perf_tracker.py
+++ b/src/forge/observability/perf_tracker.py
@@ -106,7 +106,7 @@ class Tracer:
 
         self.prefix = prefix
         self.track_memory = track_memory
-        self.time_with_gpu = timer == "gpu"
+        self.time_with_gpu = False  # timer == "gpu"
         self._disable = os.getenv(DISABLE_PERF_METRICS, "false") == "true"
         self._active = False
 

### CODE REVIEW

**Comment by vidhyav** [comment]
Did you run this for a long time?

Also how did you mem profile? Using py-spy?

**Comment by felipemello1** [comment]
1. 3 steps. OOM was peak of second.
2. Just did 'torch.cuda.memory_allocated()', 'torch.cuda.memory_reserved()'. 'torch.cuda.max_memory_allocated()' to get snapshots

File: src/forge/observability/perf_tracker.py [hunk 1]
```diff
@@ -106,7 +106,7 @@ def __init__(
 
         self.prefix = prefix
         self.track_memory = track_memory
-        self.time_with_gpu = timer == "gpu"
```
joecummings, line 109: Can you add a comment here or point to this PR? Otherwise I worry that someone will come by and turn this back on with no context.

PR #281, opened 2025-10-01
Update torchtitan for proper bf16 & new quant APIs

joecummings:
@felipemello1 Correctly diagnosed that our memory was super high for our default Qwen3 8B script. The solution he proposed in #278 was to drop the ``max_req_tokens`` and ``max_res_tokens``, which was very valid. But overall the high memory seems suspect. After looking into it, it was apparent that although we specified "bfloat16" in our Trainer/Ref sections of the configs, it was not being applied - hence the massive memory.  The culprit was that we had an out of date Torchtitan package :/

In this PR, I update the Torchtitan package so that bf16 is applied correctly. 

Memory usage before (with 486 seq len):
Screenshot 2025-10-01 at 4 14 24 PM

Memory usage after (with 512 seq len):
Screenshot 2025-10-01 at 4 00 36 PM

**sidenote**: This has the fortunate side effect of halving the weight sync speed #impact

### DIFF
--- a/apps/grpo/qwen3_8b.yaml
+++ b/apps/grpo/qwen3_8b.yaml
@@ -4,8 +4,8 @@
 # Global configuration
 group_size: 8
 batch_size: 16
-max_req_tokens: 468
-max_res_tokens: 468
+max_req_tokens: 512
+max_res_tokens: 512
 model: "Qwen/Qwen3-8B"
 off_by_n: 1 # Off by one by default
 


src/forge/actors/trainer.py
```diff
@@ -104,7 +104,7 @@ class RLTrainer(ForgeActor):
     )
     use_vllm_builtin_load: bool = True
     compile: Compile = field(default_factory=Compile)
-    float8: Float8Dense = field(default_factory=Float8Dense)
```
joecummings, line 107: This was also a part of the updated Torchtitan package

felipemello1:
nice wandb logging :D

joecummings:
> nice wandb logging :D

🫡

PR #283, opened 2025-10-01
[Perf] make placement for logprob computation flexible

JenniferWang:
I don't know what happened during my rebase; but the PR is done in https://github.com/meta-pytorch/forge/pull/287

-------
This is a fix to the training loop constantly waiting 15s for replay buffer to fill up while the policy generation only takes about 2s. The culprit being the reference model returning raw logits. 

## Before
The reference model returns raw logits to the training loop and then performs log probs calculation. However, with context window being just 512 (req + res = 512 + 512 = 1024), the raw logits that need to be transferred is [8 x 1024 x 151,936] ~ 2.489 GB. 

image

image

https://meta.wandb.io/jiyue/grpo-training/runs/mt4j7q4y?nw=nwuserjiyue

## After 
Fuse the calculation of log probs into the `forward` method which results in the return size being [8 x 512] ~ 8 KB. And the training step is no longer blocked by filling buffer
 
image

image

https://meta.wandb.io/jiyue/grpo-training/runs/xtc8vn7t?nw=nwuserjiyue

## How does it scale?
For larger Qwen3 models or DV3 models, the vocab size stays similar to the Qwen 1.7B model but the *context length* are significantly bigger and the issue with raw logits would be more significant.


apps/grpo/main.py
```diff
@@ -316,6 +306,7 @@ async def main(cfg: DictConfig):
     group_size = cfg.group_size
     max_req_tokens = cfg.max_req_tokens
     max_res_tokens = cfg.max_res_tokens
+    ref_model_return_logprobs = cfg.ref_model_return_logprobs
```
joecummings: No need to give the option here. We can just default to "return_logprobs=True"

apps/grpo/main.py
```diff
@@ -407,13 +398,18 @@ async def continuous_rollouts():
             t.step("reward_evaluation")
 
             # Calculate reference logprobs
-            ref_logits = await ref_model.forward.route(input_ids)
-            t.step("reference_model_forward")
+            if ref_model_return_logprobs:
+                ref_logprobs =  await ref_model.forward.route(input_ids, max_req_tokens, return_logprobs=True)
+                t.step("reference_model_forward_return_logprobs")
+            else:
+                ref_logits = await ref_model.forward.route(input_ids, max_req_tokens, return_logprobs=False)
```
joecummings: Feel free to delete all of this. I just wanted the class itself to be flexible. Within our own demo script, we can keep it nice and clean.

src/forge/actors/reference_model.py
```diff
@@ -90,8 +91,20 @@ async def setup(self):
         self.model.eval()
 
     @endpoint
-    async def forward(self, input_ids: torch.Tensor) -> torch.Tensor:
-
+    async def forward(
+        self, input_ids: torch.Tensor, max_req_tokens: int, return_logprobs: bool
+    ) -> torch.Tensor:
+        """
+        Args:
```
joecummings, line 98: Seems like some args are missing?

src/forge/actors/reference_model.py
```diff
@@ -90,8 +91,20 @@ async def setup(self):
...
+            return_logprobs (bool): whether to return og probabilities instead of raw logits.
+
+            This flag significantly impacts the amount of data transferred to the caller:
+            - When False: Returns logits with shape [group_size, req + res_length, vocab_size].
```
vidhyav, line 104: Omg. Why did we even do this in the first place?

apps/grpo/main.py
```diff
@@ -137,16 +137,6 @@ def collate(batches: list[list[Episode]]):
     return inputs, targets
 
 
-def compute_logprobs(
```
Jack-Khuu, line 140: I don't know if it was intentional, but thanks for matching https://github.com/meta-pytorch/forge/pull/244, it'll make whoever rebases's lives easier

apps/grpo/main.py
Jack-Khuu: Why are we leaving this as an option? Do we have a use case in mind that would want just logits?

PR #285, opened 2025-10-01
Mast setup

Ritesh1905:
- Attempt 2, of the mast setup PR
- Has a read me to help setup the mast env.

### DIFF
--- a/src/forge/controller/launcher.py
+++ b/src/forge/controller/launcher.py
@@ -124,9 +124,6 @@ class BaseLauncher:
 
 
 class Slurmlauncher(BaseLauncher):
-    def __init__(self, cfg: LauncherConfig | None = None):
-        self.cfg = cfg
-
     async def initialize(self) -> None:
         pass
 
@@ -309,11 +306,9 @@ class Mastlauncher(BaseLauncher):
 
 
 def get_launcher(cfg: LauncherConfig | None = None) -> BaseLauncher | None:
-    if not cfg:
-        return None
-    if cfg.launcher == Launcher.MAST:
+    if not cfg or cfg.launcher == Launcher.SLURM:
+        return Slurmlauncher()
+    elif cfg.launcher == Launcher.MAST:
         return Mastlauncher(cfg)
-    elif cfg.launcher == Launcher.SLURM:
-        return Slurmlauncher(cfg)
     else:
         raise ValueError(f"Unsupported config provided, got {cfg}")
--- a/src/forge/controller/provisioner.py
+++ b/src/forge/controller/provisioner.py
@@ -107,6 +107,11 @@ class Provisioner:
     async def create_host_mesh(self, name: str, num_hosts: int) -> HostMesh:
         """Creates a remote server and a HostMesh on it."""
         # no need to lock here because this is already locked behind `get_proc_mesh`
+        if not self.launcher:
+            raise RuntimeError(
+                "You tried to create a remote allocation by specifying the number of hosts on an actor or service, "
+                "but no launcher was specified."
+            )
         logger.debug(f"Creating remote server for alloc {name}")
         alloc, alloc_constraints, server_name = await self.launcher.get_allocator(
             name, num_hosts


apps/mast/main.py
```diff
@@ -0,0 +1,51 @@
...
+from omegaconf import DictConfig
+
+DEFAULT_CHECKPOINT_FOLDER_KEY = "checkpoint_folder"
+DEFAULT_CHECKPOINT_FOLDER = "/mnt/wsfuse/teamforge/forge_runs/"
+
+
+async def main(cfg: DictConfig):
+    """Main module for launching mast jobs for GRPO training."""
+    if cfg.get(SCHEDULER_KEY, Scheduler.MAST.value) != Scheduler.MAST.value:
+        raise ValueError("Schuduler must be MAST.")
```
vidhyav: nit: Scheduler

allenwang28:
Can we change the design a bit?

- Let's retain the Provisioner as original
- We introduce scheduler specific classes like this:

```
class BaseScheduler:
    async def initialize(self):
        pass

    def get_allocator(self):
        pass


class MastScheduler(BaseScheduler):
    async def initialize(self):
        await self._build_appdef(...)
        await self._launch_mast_job(...)
        ...

    async def get_allocator(self):
        allocator = MastAllocator(...)
        return allocator, alloc_constraints


class SlurmScheduler(BaseScheduler):
    async def initialize(self):
        # no op

    async def get_allocator(self):
        appdef = hyperactor.host_mesh(...)
        server_info = commands.get_or_create(...)
        alloc = RemoteAllocator(...)
        return alloc, None
```

There are other things I'd want cleaned up, but I think if we can make this proposed change that is sufficient for the sake of this PR.

src/forge/controller/launcher/mast.py
```diff
@@ -0,0 +1,378 @@
...
+    async def shutdown(self):
+        """Tears down all remaining remote allocations."""
+        async with self._lock:
+            for server_name in self._server_names:
+                commands.kill(server_name)
+
+    async def launch_mast_job(self):
+        handle = self.create_server_handle()
+        server_spec = info(handle)
+        if server_spec and server_spec.state == AppState.RUNNING:
```
LucasLLC: Why do this at all? why not always just return commands.get_or_create... ?

src/forge/controller/launcher/mast.py
```diff
@@ -0,0 +1,378 @@
...
+                remote_allocator_port=26600,  # This is the default monarch port
+            ),
+        )
+        alloc_constraints = AllocConstraints(
+            {MastAllocator.ALLOC_LABEL_TASK_GROUP: task_group}
+        )
+
+        return allocator, alloc_constraints
+
+    async def create_host_mesh(self, name: str, num_hosts: int):
```
LucasLLC: just as an observation, imo we should be able to remove most of this complexity with host mesh. Specifically we should use a single host mesh for the entire job, remove TG specific logic, and go from there.

apps/mast/env_setup.sh
```diff
@@ -0,0 +1,270 @@
...
+log_warn() {
+    echo -e "${YELLOW}[WARN]${NC} $1"
+}
+
+log_error() {
+    echo -e "${RED}[ERROR]${NC} $1"
+}
+
+# Function to mount a single workspace to /mnt/wsfuse
+mount_workspace() {
```
LucasLLC, line 32: Don't need to address here, but in the future we shouldn't make this a requirement for launching mast jobs. By leaving this open it's actually possible the remote job uses some locally passed assets

src/forge/types.py
```diff
@@ -87,13 +88,20 @@ class State:
     metadata: dict[str, Any] = field(default_factory=dict)
 
 
+class Launcher(Enum):
+    MAST = "mast"
+    SLURM = "slurm"
+    LOCAL = "local"
```
allenwang28: since there isn't such thing as a local launcher, can we remove LOCAL?

src/forge/controller/launcher.py
```diff
@@ -0,0 +1,321 @@
...
+        return appdef
+
+    def create_job_name(self):
+        return f"{USER}-forge-{uuid.uuid4().hex[:6]}"
+
+    def create_server_handle(self) -> str:
+        return f"{self.scheduler_name}:///{self.job_name}"
+
+
+def get_launcher(cfg: DictConfig | None = None) -> BaseLauncher:
```
allenwang28: I would like to keep OmegaConf constrained to only the apps as it isn't explicit enough. The pattern we've been following throughout Forge is something like this:

```
@dataclass
class LauncherConfig:
    ...

@dataclass
class SlurmConfig:
    ...

@dataclass
class MastConfig:
    ...
```

Would also like to get rid of the Local launcher:
```
def get_launcher(cfg: LauncherConfig | None = None) ->  BaseLauncher | None:
    """Returns the launcher given the configuration."""
    if not cfg:
        return None

    if isinstance(cfg, MastConfig):
        return MastLauncher(cfg)
    elif isinstance(cfg, SlurmConfig):
        return SlurmLauncher(cfg)
    else:
        raise ValueError(f"Unsupported config provided, got {cfg}") 
```

src/forge/controller/provisioner.py
```diff
@@ -108,44 +94,34 @@ def __init__(self):
         self._host_gpu_map = {
             self._this_host_id: GpuManager(available_local_devices),
         }
+        self.launcher: BaseLauncher = get_launcher(cfg)
```
allenwang28:
```suggestion
        self.launcher: BaseLauncher | None = get_launcher(cfg)
        if not self.launcher:
            logger.warning("Launcher not provided, remote allocations will not work.")
```

src/forge/controller/provisioner.py
```diff
@@ -108,44 +94,34 @@ def __init__(self):
...
+
+    async def initialize(self):
+        """Call this after creating the instance"""
+        await self.launcher.initialize()
```
allenwang28:
```suggestion
        if self.launcher is not None:
            await self.launcher.initialize()
```

src/forge/controller/provisioner.py
```diff
@@ -108,44 +94,34 @@ def __init__(self):
...
 
     async def create_host_mesh(self, name: str, num_hosts: int) -> HostMesh:
         """Creates a remote server and a HostMesh on it."""
         # no need to lock here because this is already locked behind `get_proc_mesh`
         logger.debug(f"Creating remote server for alloc {name}")
```
allenwang28, line 115:
```
if not self.launcher:
    raise RuntimeError("You tried to create a remote allocation by specifying the number of hosts on an actor or service, but no launcher was specified.")
```

src/forge/controller/provisioner.py
Ritesh1905: why? Thoughts on defaulting `get_launcher` to return Slurm, rather than throwing this exception?

allenwang28: because people running on like a dev GPU might set this unknowingly and wonder why they're getting a SLURM error

apps/mast/qwen3_14b_mast.yaml
```diff
@@ -0,0 +1,162 @@
+# Grouped Relative Policy Optimization (GRPO)
+# >>> python -m apps.grpo.main --config apps/grpo/qwen3_1_7b.yaml
```
daniellepintz, line 2: nit: wrong cmd

PR #288, opened 2025-10-02
recreate wheel build PR

ebsmothers:
Need to recreate this PR due to issues after the repo being made public. Only change from the previous version is the addition of the cp command at the end of the post-build script


.github/packaging/pre_build_gpu.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+set -euxo pipefail
+
+# Builds vLLM, Monarch and torchstore
+# This script builds vLLM, Monarch and torchstore and places
+# their wheels into dist/.
+
+MONARCH_COMMIT="265034a29ec3fb35919f4a9c23c65f2f4237190d"
+TORCHTITAN_COMMIT="82f0287b966f1735819a377a9a09e7a303c55faa"
```
joecummings: nightly?

ebsmothers: This is actually not used. I think we really only need to successfully upload Monarch and vLLM here, the rest can be handled elsewhere

felipemello1:
this looks fun

.github/packaging/post_build_script.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -euxo pipefail
+
+FORGE_WHEEL=${GITHUB_WORKSPACE}/${REPOSITORY}/dist/*.whl
+WHL_DIR="${GITHUB_WORKSPACE}/wheels/dist"
+DIST=dist/
+
+echo "Uploading wheels to S3"
```
allenwang28: should we change this to like "Here are the files we're uploading to S3" or remove altogether?

.github/packaging/pre_build_cpu.sh
```diff
@@ -0,0 +1,33 @@
...
+
+    python use_existing_torch.py
+    pip install -r requirements/build.txt
+    export VERBOSE=1
+    export CMAKE_VERBOSE_MAKEFILE=1
+    export FORCE_CMAKE=1
+    pip wheel -v --no-build-isolation --no-deps . -w "$WHL_DIR"
+}
+
+build_vllm
```
allenwang28, line 32: oh I wonder if we need to do the append_date here too. Which forge wheel ultimately gets uploaded?

.github/workflows/build_wheels.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - nightly
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  build:
+    # if: github.repository_owner == 'pytorch'
```
allenwang28:
```suggestion
    if: github.repository_owner == 'pytorch'
```

.github/workflows/build_vllm.yaml
```diff
@@ -0,0 +1,52 @@
+name: Build pinned vLLM against PyTorch nightly and upload
+
+on:
+  pull_request:
```
allenwang28: should we remove `pull_request`? I assume we don't want this to actually run all the builds per PR

ebsmothers: Oh yeah good call, thank you

.github/packaging/post_build_script.sh
ebsmothers: Yeah let me make this a bit clearer

.github/packaging/pre_build_cpu.sh
ebsmothers: append_date is updating the forge wheel's name, right? Do we even care about that? (Actually if we are running this nightly shouldn't we be appending date to monarch and vllm wheels?)

allenwang28: I guess that's true on both accounts. I dunno I'm not too opinionated

PR #290, opened 2025-10-02
Add commands to configs and update readme

daniellepintz:
* Adding commands to top of configs for easy usability
* Fix some small typos in readme

### DIFF
--- a/apps/vllm/qwen2_5_32b.yaml
+++ b/apps/vllm/qwen2_5_32b.yaml
@@ -1,3 +1,5 @@
+# >>> python -m apps.vllm.main --config apps/vllm/qwen2_5_32b.yaml
+
 policy:
   engine_config:
     model: "Qwen/Qwen2.5-32B"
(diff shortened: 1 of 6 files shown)


apps/vllm/qwen2_5_32b.yaml
```diff
@@ -12,7 +14,7 @@ policy:
 services:
   policy:
     procs: 4
-    hosts: 1
+    #hosts: 1
```
joecummings: Why is this commented out?

joecummings (approved):
One question, but otherwise looks good!

apps/vllm/qwen2_5_32b.yaml
allenwang28: yeah we can just delete it

daniellepintz: sg will delete it. Needed to remove it bc otherwise I get a slurm error: P1972107650 which is not great to happen out of the box

PR #292, opened 2025-10-02
factor out weight cleanup to separate file

casteryh:


src/forge/util/weight_sync.py
joecummings, line 1: Why not put this in _torchstore_utils?

It seems that's where all the other weight sync information is. If that's not supposed to be the end location, I'd almost rather all _torchstore_utils be moved out into a weight_sync.py file.

src/forge/util/weight_sync.py
```diff
@@ -0,0 +1,31 @@
...
+
+import torchstore as ts
+
+from forge.actors._torchstore_utils import (
+    get_dcp_whole_state_dict_key,
+    get_param_prefix,
+)
+
+
+async def drop_weights(version: int):
```
joecummings: Now that we're here, I'd prefer a name like "delete_old_weights" 

And instead of version, something like "oldest_version_to_keep"

src/forge/util/weight_sync.py
```diff
@@ -0,0 +1,31 @@
...
+    print(f"Dropping weights @ version {version}")
```
joecummings: Remove this

src/forge/util/weight_sync.py
```diff
@@ -0,0 +1,31 @@
...
+    # TODO: once we have something like `get_meta()` in torchstore, we can just
+    # query the type of the object instead of relying on keys.
+    dcp_key = get_dcp_whole_state_dict_key(version)
+    if dcp_key in matching_keys:
+        dcp_handle = await ts.get(dcp_key)
+        dcp_handle.drop()
+    for key in matching_keys:
+        await ts.delete(key)
+    elapsed = time.perf_counter() - start_time
+    print(f"Dropped weights @ version {version}, took {elapsed:.2f} seconds")
```
joecummings: Log instead of print

apps/grpo/main.py
```diff
@@ -455,9 +435,9 @@ async def continuous_training():
                 await policy.update_weights.fanout(training_step)
                 t.step("update_weights")
 
-                # if training_step >= 2:
-                #     await drop_weights(training_step - 1)
-                #     t.step("drop_weights")
+                if training_step >= 2:
+                    await drop_weights(training_step - 1)
```
joecummings: Can this be truly async or does it have to be blocking like this?

casteryh: this can be truly async, if we just create a task and not await on it.

src/forge/util/weight_sync.py
casteryh: > rather all _torchstore_utils be moved out into a weight_sync.py

Maybe I will do this. Let me know wyt. Do you want me to make it _weight_sync.py instead?

casteryh:
image

Ended up doing a whole refactor. It does the weight clean up in the background in a non-blocking way now.
tested: unit tests and running grpo main (https://meta.wandb.io/torchforge/grpo-training/runs/r8q88uie)
ptal @joecummings

src/forge/util/_torchstore.py
```diff
@@ -69,3 +73,54 @@ def extract_param_name(key: str) -> str:
...
+    async def wait(self):
+        """Wait for all scheduled deletion tasks to complete."""
+        await asyncio.gather(*self._tasks)
+
+
+async def drop_weights(version: int):
+    start_time = time.perf_counter()
+    prefix = get_param_prefix(version)
+    matching_keys = await ts.keys(prefix)
+    # TODO: once we have something like `get_meta()` in torchstore, we can just
```
LucasLLC, line 117: we do have a 'get_meta' in torchstore (although it's lacking a proper object).

src/forge/util/_torchstore.py
```diff
@@ -69,3 +73,54 @@ def extract_param_name(key: str) -> str:
...
+    # query the type of the object instead of relying on keys.
+    dcp_key = get_dcp_whole_state_dict_key(version)
```
LucasLLC, line 119: Is this implementation specific to DCP?

Do we need something like `(ts.delete(r"key.*")` support in torchstore?

LucasLLC (commented):
Asyncio looks reasonable to me. A good way to measure this would be to run this code for a while and always synchronize (finish your tasks) on weighsync step, and measure the delta between doing that and waiting until the next step

apps/grpo/main.py
```diff
@@ -463,9 +444,9 @@ async def continuous_training():
                 await policy.update_weights.fanout(training_step)
                 t.step("update_weights")
 
-                if training_step >= 2:
-                    await drop_weights(training_step - 1)
-                    t.step("drop_weights")
+                # weight cleanup is non-blocking, the task is executed in the background
```
joecummings, line 433: How are you confirming this all finishes before adding more weights?

Also in typical async form this step would just be an async method that you'd await now or later. Why is there an extra method called "wait"?

src/forge/util/_torchstore.py
joecummings, line 1: Move to core app/

apps/grpo/main.py
casteryh: > How are you confirming this all finishes before adding more weights?

I thought the point is you don't, if you just need the weight to be eventually deleted. when you do `step()`, the task is scheduled in the background and everything else proceeds as normal.

> Also in typical async form this step would just be an async method that you'd await now or later.

Yes but in that case, if we want to schedule the task in the background and not await for it, we need to manage the task in `main.py`, which we supposedly don't want to do. This essentially hides the task scheduling logic in the WeightCleaner class.

> Why is there an extra method called "wait"?

If you want to make sure all the scheduled tasks are indeed completed (i.e. all old weights are deleted. like you mentioned earliner), you can `await weight_cleaner.wait()`. Presumably this can be named better, let me know what you think.

src/forge/util/_torchstore.py
casteryh: > Is this implementation specific to DCP?

Yes

> Do we need something like `(ts.delete(r"key.*")` support in torchstore?

It would be good if we can have it. Although currently it is not a bottleneck to simply call delete on every key.

apps/grpo/main.py
casteryh: > Also in typical async form this step would just be an async method that you'd await now or later. Why is there an extra method called "wait"?

My understanding is, in typical async code, if you don't explicitly create a task, then it will never get executed unless you await on it? I think we can also always schedule the task and return a join handle.

src/forge/util/_torchstore.py
casteryh: Everything or only the WeightCleaner? trainer and policy both need functions in this file.

casteryh:
ptal @joecummings

PR #293, opened 2025-10-02
Update ``RLTrainer`` init variables to match TorchTitan ``JobConfig``

joecummings:
Needs https://github.com/pytorch/torchtitan/pull/1788 to land first, then we can do this

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -93,6 +93,7 @@ def cleanup_old_weight_versions(
 
 @dataclass
 class RLTrainer(ForgeActor):
+    job: Job = field(default_factory=Job)
     model: Model = field(default_factory=Model)
     optimizer: Optimizer = field(default_factory=Optimizer)
     lr_scheduler: LRScheduler = field(default_factory=LRScheduler)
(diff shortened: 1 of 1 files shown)


src/forge/actors/trainer.py
```diff
@@ -102,15 +105,17 @@ class RLTrainer(ForgeActor):
...
-    float8: Float8Linear = field(default_factory=Float8Linear)
+    quantize: Quantize = field(default_factory=Quantize)
     comm: Comm = field(default_factory=Comm)
+    memory_estimation: MemoryEstimation = field(default_factory=MemoryEstimation)
+    # Non JobConfig-related fields
     loss: Callable = lambda logits, **targets: logits
     state_dict_key: str = "model_state_dict"
     use_dcp: bool = True
     dcp_path: str = "forge_dcp_tmp"
     vllm_tp_DEPRECATED: int = 1  # noqa: N815
```
ebsmothers, line 117: random q but why are we keeping around deprecated fields like this? we don't really need to care about BC rn right?

joecummings: https://github.com/meta-pytorch/forge/issues/295

PR #294, opened 2025-10-02
Add basic issue template

Jack-Khuu:
Based on the template used by torchchat and executorch

**Example of what this looks like in ET**
image


casteryh:
Can we test it before merging? If not maybe just go ahead before anyone is actually watching this repo.

casteryh (approved):

Jack-Khuu:
Test in prod (But yeah it's pretty harmless for us to land this then just tweak as we want)

PR #300, opened 2025-10-03
Zero to Forge Tutorials

init27:
Hi team, 

As promised, here is the three part tutorial series covering (written in MD):

1. RL and Forge concepts (with conceptual training loops)
2. Forge Internals (With training loop borrowed from GRPO)
3. Monarch 101: Sharing the overview of how Monarch Internals behave and come together

### DIFF
--- a/docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
+++ b/docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
@@ -85,7 +85,6 @@ graph LR
     end
     
     subgraph Services["Forge Services (Real Classes)"]
-    
         S1["DatasetActor"]
         S2["Policy"]
         S3["RewardActor"]
@@ -108,6 +107,8 @@ graph LR
 ```
 
 ### RL Step with Forge Services
+
+Let's look at the example from above again, but this time we would use the names from Forge:
 
 ```python
 # Conceptual Example
@@ -144,6 +145,8 @@ async def conceptual_forge_rl_step(services, step):
 ```
 
 **Key difference**: Same RL logic, but each component is now a distributed, fault-tolerant, auto-scaling service.
+
+Did you realise-we are not worrying about any Infra code here! Forge Automagically handles the details behind the scenes and you can focus on writing your RL Algorthms!
 
 
 ## Why This Matters: Traditional ML Infrastructure Fails
--- a/docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
+++ b/docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
@@ -275,14 +275,14 @@ async def real_rl_training_step(services, step):
 ```python
 responses = await policy.generate.route(prompt=question)
 answer = responses[0].text  # responses is list[Completion]
-
-# Forge handles behind the scenes:
-# - Routing to least loaded replica
-# - GPU memory management  
-# - Batch optimization
-# - Failure recovery
-# - Auto-scaling based on demand
 ```
+
+Forge handles behind the scenes:
+- Routing to least loaded replica
+- GPU memory management  
+- Batch optimization
+- Failure recovery
+- Auto-scaling based on demand
 
 ### Independent Scaling
 ```python
@@ -361,12 +361,13 @@ group_size = 1
             reward_functions=[MathReward(), ThinkingReward()],
         )
     )
-
-# Production scaling - multiply num_replicas:
-# Policy: num_replicas=8 for high inference demand
-# RewardActor: num_replicas=16 for parallel evaluation
-# Trainer: num_replicas=4 for distributed training
 ```
+
+Production scaling - multiply num_replicas:
+- Policy: num_replicas=8 for high inference demand
+- RewardActor: num_replicas=16 for parallel evaluation
+- Trainer: num_replicas=4 for distributed training
+
 
 ### Fault Tolerance
 ```python
@@ -381,9 +382,12 @@ answer = responses[0].text
 score = await reward_actor.evaluate_response.route(
     prompt=question, response=answer, target=target
 ) 
-# -> Retries on different replica automatically
-# -> Graceful degradation if all replicas fail
-# -> System continues (may need application-level handling)
 ```
 
-This is fundamentally different from monolithic RL implementations where any component failure stops everything.
+- Retries on different replica automatically
+- Graceful degradation if all replicas fail
+- System continues (may need application-level handling)
+
+This is fundamentally different from monolithic RL implementations where any component failure stops everything!
+
+In the next Section, we will go a layer deeper and learn how ForgeServices work. Continue to [Part 2 here](./2_Forge_Internals.MD)
--- a/docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
+++ b/docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
@@ -1,0 +1,385 @@
+# Part 1: RL Fundamentals - Using Forge Terminology
+
+## Core RL Components in Forge
+
+Let's start with a simple math tutoring example to understand RL concepts with the exact names Forge uses:
+
+### The Toy Example: Teaching Math
+
+```mermaid
+graph TD
+    subgraph Example["Math Tutoring RL Example"]
+        Dataset["Dataset<br/>math problems<br/>'What is 2+2?'"]
+        Policy["Policy<br/>student AI<br/>generates: 'The answer is 4'"]
+        Reward["Reward Model<br/>Evaluation Exam<br/>scores: 0.95 (excellent)"]
+        Reference["Reference Model<br/>original student<br/>baseline comparison"]
+        ReplayBuffer["Replay Buffer<br/>notebook<br/>stores experiences"]
+        Trainer["Trainer<br/>tutor<br/>improves student"]
+    end
+    
+    Dataset --> Policy
+    Policy --> Reward
+    Policy --> Reference
+    Reward --> ReplayBuffer
+    Reference --> ReplayBuffer
+    ReplayBuffer --> Trainer
+    Trainer --> Policy
+    
+    style Policy fill:#99ff99
+    style Reward fill:#ffcc99
+    style Trainer fill:#ff99cc
+```
+
+### RL Components Defined (Forge Names)
+
+1. **Dataset**: Provides questions/prompts (like "What is 2+2?")
+2. **Policy**: The AI being trained (generates answers like "The answer is 4")
+3. **Reward Model**: Evaluates answer quality (gives scores like 0.95)
+4. **Reference Model**: Original policy copy (prevents drift from baseline)
+5. **Replay Buffer**: Stores experiences (question + answer + score)
+6. **Trainer**: Updates the policy weights based on experiences
+
+### The RL Learning Flow
+
+```python
+# CONCEPTUAL EXAMPLE - see apps/grpo/main.py for GRPO Code
+
+def conceptual_rl_step():
+    # 1. Get a math problem
+    question = dataset.sample()  # "What is 2+2?"
+    
+    # 2. Student generates answer  
+    answer = policy.generate(question)  # "The answer is 4"
+    
+    # 3. Teacher grades it
+    score = reward_model.evaluate(question, answer)  # 0.95
+    
+    # 4. Compare to original student
+    baseline = reference_model.compute_logprobs(question, answer)
+    
+    # 5. Store the experience
+    experience = Episode(question, answer, score, baseline)
+    replay_buffer.add(experience)
+    
+    # 6. When enough experiences collected, improve student
+    batch = replay_buffer.sample(curr_policy_version=0)
+    if batch is not None:
+        trainer.train_step(batch)  # Student gets better!
+        
+# 🔄 See complete working example below with actual Forge service calls
+```
+
+## From Concepts to Forge Services
+
+Here's the key insight: **Each RL component becomes a Forge service**. The toy example above maps directly to Forge:
+
+```mermaid
+graph LR
+    subgraph Concepts["RL Concepts"]
+        C1["Dataset"]
+        C2["Policy"]
+        C3["Reward Model"]
+        C4["Reference Model"]
+        C5["Replay Buffer"]
+        C6["Trainer"]
+    end
+    
+    subgraph Services["Forge Services (Real Classes)"]
+        S1["DatasetActor"]
+        S2["Policy"]
+        S3["RewardActor"]
+        S4["ReferenceModel"]
+        S5["ReplayBuffer"]
+        S6["RLTrainer"]
+    end
+    
+    C1 --> S1
+    C2 --> S2
+    C3 --> S3
+    C4 --> S4
+    C5 --> S5
+    C6 --> S6
+    
+    style C2 fill:#99ff99
+    style S2 fill:#99ff99
+    style C3 fill:#ffcc99
+    style S3 fill:#ffcc99
+```
+
+### RL Step with Forge Services
+
+```python
+# Conceptual Example
+
+async def conceptual_forge_rl_step(services, step):
+    # 1. Get a math problem - CONCEPTUAL API
+    sample = await services['dataloader'].get_sample()
+    question, target = sample["question"], sample["answer"]
+    
+    # 2. Student generates answer - CONCEPTUAL API
+    # Actual method names vary by implementation
+    responses = await services['policy'].generate(prompt=question)
+    answer = responses[0].text  
+    
+    # 3. Teacher grades it - CONCEPTUAL API  
+    # Actual reward evaluation varies by implementation
+    score = await services['reward_actor'].evaluate(
+        prompt=question, response=answer, target=target
+    )
+    
+    # 4. Compare to baseline - CONCEPTUAL API
+    ref_logprobs = await services['ref_model'].compute_baseline(responses[0].token_ids)
+    
+    # 5. Store experience - CONCEPTUAL Episode structure
+    # Real Episode structure in src/forge/data_models/episode.py
+    episode = create_episode(responses[0], score, ref_logprobs, step)
+    await services['replay_buffer'].store(episode)
+    
+    # 6. Improve student - CONCEPTUAL API
+    batch = await services['replay_buffer'].get_batch(policy_version=step)
+    if batch is not None:
+        loss = await services['trainer'].update_policy(batch)
+        return loss
+```
+
+**Key difference**: Same RL logic, but each component is now a distributed, fault-tolerant, auto-scaling service.
+
+
+## Why This Matters: Traditional ML Infrastructure Fails
+
+### The Infrastructure Challenge
+
+Our simple RL loop above has complex requirements:
+
+#### Problem 1: Different Resource Needs
+
+```mermaid
+graph TD
+    subgraph Components["Each Component Needs Different Resources"]
+        Policy["Policy (Student AI)<br/>Generates: 'The answer is 4'<br/>Needs: Large GPU memory<br/>Scaling: Multiple replicas for speed"]
+        
+        Reward["Reward Model (Teacher)<br/>Scores answers: 0.95<br/>Needs: Moderate compute<br/>Scaling: CPU or small GPU"]
+        
+        Trainer["Trainer (Tutor)<br/>Improves student weights<br/>Needs: Massive GPU compute<br/>Scaling: Distributed training"]
+        
+        Dataset["Dataset (Question Bank)<br/>Provides: 'What is 2+2?'<br/>Needs: CPU intensive I/O<br/>Scaling: High memory bandwidth"]
+    end
+    
+    style Policy fill:#99ff99
+    style Reward fill:#ffcc99
+    style Trainer fill:#ff99cc
+    style Dataset fill:#ccccff
+```
+
+### Problem 2: Complex Interdependencies
+
+```mermaid
+graph LR
+    A["Policy: Student AI<br/>'What is 2+2?' → 'The answer is 4'"]
+    B["Reward: Teacher<br/>Scores answer: 0.95"]
+    C["Reference: Original Student<br/>Provides baseline comparison"]
+    D["Replay Buffer: Notebook<br/>Stores: question + answer + score"]
+    E["Trainer: Tutor<br/>Improves student using experiences"]
+    
+    A --> B
+    A --> C
+    B --> D
+    C --> D
+    D --> E
+    E --> A
+    
+    style A fill:#99ff99
+    style B fill:#ffcc99
+    style C fill:#99ccff
+    style D fill:#ccff99
+    style E fill:#ff99cc
+```
+
+Each step has different:
+- **Latency requirements**: Policy inference needs low latency, training can batch
+- **Scaling patterns**: Reward evaluation scales with response count, training with model size
+- **Failure modes**: Policy failure stops generation, reward failure affects learning quality
+- **Resource utilization**: GPUs for inference/training, CPUs for data processing
+
+### Problem 3: The Coordination Challenge
+
+Unlike supervised learning where you process independent batches, RL requires coordination:
+
+```python
+# This won't work - creates bottlenecks and resource waste
+def naive_rl_step():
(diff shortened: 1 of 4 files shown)


docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+    style A fill:#99ff99
+    style B fill:#ffcc99
+    style C fill:#99ccff
+    style D fill:#ccff99
+    style E fill:#ff99cc
+```
+
+Each step has different:
+- **Latency requirements**: Policy inference needs low latency, training can batch
+- **Scaling patterns**: Reward evaluation scales with response count, training with model size
```
allenwang28: for scaling patterns, I guess the point is technically more like "we might want N generators to keep the trainer busy + the generators and trainers may have different sharding strategies"?

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+- **Failure modes**: Policy failure stops generation, reward failure affects learning quality
```
allenwang28: > reward failure affects learning quality

this is true, but I feel like the way this is phrased is more of a "if your reward model failed in its function" (like giving you the WRONG reward)

But Forge's failure handling is more about "if a component fails, all of the other components fail because it's a complex web of dependencies" so the system grinds to a halt - Forge is trying to handle that as much as possbile

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+- **Resource utilization**: GPUs for inference/training, CPUs for data processing
+
+### Problem 3: The Coordination Challenge
+
+Unlike supervised learning where you process independent batches, RL requires coordination:
+
+```python
+# This won't work - creates bottlenecks and resource waste
```
allenwang28: nit: but I would say more like "while this does work, it creates bottlenecks and resource waste"

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+    loss = compute_loss(response, reward)  # Batch size = 1, inefficient
+    
+    # Everything stops if any component fails
+    if policy_fails or reward_fails or trainer_fails:
+        entire_system_stops()
+```
+
+## Enter Forge: RL-Native Architecture
+
+Forge solves these problems by treating each RL component as an **independent, scalable service**
```
allenwang28: > by treating each RL component as an **independent, scalable service**

@init27 - we have actually made pieces like the Trainer a regular actor, since the recovery semantics are unclear, whereas handling a vLLM generator going down is pretty straightforward. We *could* have torchft style recovery but it's unclear how well this affects RL numerics (and we just haven't had time to add it yet)

It's also why you see some `call_one()`, `call()` (actor APIs) mixed in with `route()` and `fanout()` which are service APIs in the real RL training step. It's a logical choice, but hard to tell the story at the 101 level. I'm wondering if this is something we should mention at all, or if you have any ideas for how to simplify.

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+        RewardActor.options(procs=1, num_replicas=1).as_service(
+            reward_functions=[MathReward(), ThinkingReward()]
+        )
+    )
+```
+
+Production scaling - multiply num_replicas for services or spawn multiple actors:
+- Policy: num_replicas=8 for high inference demand
+- RewardActor: num_replicas=16 for parallel evaluation
+- Trainer: Multiple actors for distributed training (RLTrainer handles this internally)
```
allenwang28: nit: multiple processes

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,676 @@
...
+Here's the actual ServiceConfig from Forge source code:
+
+```python
+# Configuration pattern from apps/grpo/main.py:
+Policy.options(
+    procs=1,           # Processes per replica
+    num_replicas=4,    # Number of replicas  
+    with_gpus=True     # Allocate GPUs
+    # Other available options:
+    # hosts=None
```
allenwang28:
```suggestion
    # hosts=None   #  the number of remote hosts used per replica
```

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,676 @@
...
+)
+```
+
+### 2. Real Service Creation
+
+Services are created using the `spawn_service` function:
```
allenwang28: delete this line?

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,676 @@
...
+responses = await policy.generate.route(prompt=prompt)
+```
+
+But behind the scenes:
+- `ServiceInterface` selects healthy replica
+- Routes message to that replica's `Policy.generate()` endpoint
+- Handles failures and retries automatically
+- Returns list[Completion] from the selected replica
+
+### 3. Different Service Types and Their Characteristics
```
allenwang28: I think this section is a state we will strive for for Forge, but we're not there yet frankly 😅

can we remove this for now?

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,676 @@
...
+**When to use**: Normal request routing where any replica can handle the request.
+
+```python
+responses = await policy.generate.route(prompt=question)
+answer = responses[0].text  # Extract text from Completion object
+```
+
+Behind the scenes:
+1. Health check eliminates failed replicas
+2. Load balancer picks least loaded healthy replica
```
allenwang28: it's technically round robin here, but we're working towards "you can specify your own load balancer"

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,676 @@
...
+```
+
+**Performance characteristics**:
+- **Latency**: Process first result immediately  
+- **Throughput**: Pipeline parallelism (much higher than sequential)
+- **Fault tolerance**: Continues if some replicas fail
+
+**Critical insight**: This is essential for high-throughput RL where you can't wait for batches.
+
+### 4. Fire-and-Forget Operations
```
allenwang28: I'm not sure I understand this section - I guess we don't have fire-and-forget operations built into services. I would remove this part entirely for now

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,676 @@
...
+### Replay Buffer Consistency
+
+**The challenge**: Multiple trainers and experience collectors reading/writing concurrently.
+
+**Real Forge approach**: The ReplayBuffer actor handles concurrency internally:
+
+```python
+# Forge ReplayBuffer endpoints (verified from source code)
+# Add episodes (thread-safe by actor model)
+await replay_buffer.add.call_one(episode)  # Note: .call_one() not .choose()
```
allenwang28: if it helps, `.choose()` would work too, but we're doing `.call_one()` to sanity check that the `replay_buffer` is a singleton actor rather than an `ActorMesh`

docs/Tutorials/3_Monarch_101.MD
```diff
@@ -0,0 +1,439 @@
...
+        ServiceInterface -.->|"'The answer is 4'"| UserCall
+    end
+
+    style UserCall fill:#99ff99
+    style ServiceActor fill:#ffcc99
+    style PolicyActor fill:#cc99ff
+    style GPU fill:#ffcccc
+```
+
+## Multiple Services Sharing Infrastructure
```
allenwang28, line 260: We are definitely shooting for this, but it sadly isn't there yet! You do have multiple services sharing now, but only in context of single node

allenwang28 (commented):
Thanks Sanyam, this is awesome! Had a bunch of comments but I do generally like the flow and where we're at with this

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+    subgraph Concepts["RL Concepts"]
+        C1["Dataset"]
+        C2["Policy"]
+        C3["Reward Model"]
+        C4["Reference Model"]
+        C5["Replay Buffer"]
+        C6["Trainer"]
+    end
+    
+    subgraph Services["Forge Services (Real Classes)"]
```
ebsmothers, line 87: nit but at least for me this part does not render right: the text gets split across two lines and "Classes)" ends up buried under DatasetActor

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+
+### The Infrastructure Challenge
+
+Our simple RL loop above has complex requirements:
+
+#### Problem 1: Different Resource Needs
+
+```mermaid
+graph TD
+    subgraph Components["Each Component Needs Different Resources"]
```
ebsmothers: Similar comment here around multi-line split formatting

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+        Policy["Policy (Student AI)<br/>Generates: 'The answer is 4'<br/>Needs: Large GPU memory<br/>Scaling: Multiple replicas for speed"]
+        
+        Reward["Reward Model (Teacher)<br/>Scores answers: 0.95<br/>Needs: Moderate compute<br/>Scaling: CPU or small GPU"]
```
ebsmothers: Nit but it doesn't really have to be a "model", right? Can e.g. just be some heuristic. (As written I was confused why we would run a model on CPU)

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+    
+    style A fill:#99ff99
+    style B fill:#ffcc99
+    style C fill:#99ccff
+    style D fill:#ccff99
+    style E fill:#ff99cc
+```
+
+Each step has different:
+- **Latency requirements**: Policy inference needs low latency, training can batch
```
ebsmothers: I don't understand the "training can batch" point here. Is this in reference to different episodes that we're sampling?

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+            training={"dtype": "bfloat16"}
+        ),
+        # Reward actor (CPU)
+        RewardActor.options(procs=1, num_replicas=1).as_service(
+            reward_functions=[MathReward(), ThinkingReward()]
+        )
+    )
+```
+
+Production scaling - multiply num_replicas for services or spawn multiple actors:
```
ebsmothers: At this point (and in the code snippet above) the concept of an actor has not yet been introduced to the reader, right? May be worth at least giving some basic idea (and how it differs from services)

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,676 @@
...
+        # Process batch immediately
+        loss = await trainer.train_step.call_one(batch)
+        print(f"Training loss: {loss}")
+    else:
+        await asyncio.sleep(0.1)  # Wait for more data
+```
+
+**Performance characteristics**:
+- **Latency**: Process first result immediately  
+- **Throughput**: Pipeline parallelism (much higher than sequential)
```
ebsmothers: Is it accurate to call this "pipeline parallelism"? I thought this is basically just demonstrating non-blocking operations (maybe I'm missing the point though)

init27:
Thanks so much Allen! Will wait for Evan's review and then will address all the comments to land this. Thanks for your time both!

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
+# Part 1: RL Fundamentals - Using Forge Terminology
+
+## Core RL Components in Forge
+
+Let's start with a simple math tutoring example to understand RL concepts with the exact names Forge uses:
+
+### The Toy Example: Teaching Math
+
+```mermaid
```
felipemello1, line 9: this is how it displays to me. Can we pick better collors? :P
image

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+Did you realise-we are not worrying about any Infra code here! Forge Automagically handles the details behind the scenes and you can focus on writing your RL Algorthms!
+
+
+## Why This Matters: Traditional ML Infrastructure Fails
+
+### The Infrastructure Challenge
+
+Our simple RL loop above has complex requirements:
+
+#### Problem 1: Different Resource Needs
```
felipemello1, line 161: IMO, using boxes here is a bit of an overkill. Plain text or a table may be better?

image

init27: Thanks, I will change to table

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
```diff
@@ -0,0 +1,395 @@
...
+```python
+# Conceptual Example
+
+async def conceptual_forge_rl_step(services, step):
+    # 1. Get a math problem - Using actual DatasetActor API
+    sample = await services['dataloader'].sample.call_one()
+    question, target = sample["request"], sample["target"]
+    
+    # 2. Student generates answer - Using actual Policy API
+    responses = await services['policy'].generate.route(prompt=question)
```
felipemello1, line 122: when i first used monarch it was not 100% clear what call_one, route, call meant. Perhaps its worth introducing it very briefly a block before this one?

svekars:
Can we move the files under https://github.com/meta-pytorch/forge/tree/main/docs/source/tutorial_sources directory and  list them in the https://github.com/meta-pytorch/forge/blob/main/docs/source/tutorials.md toctree? 

Also, if you want to make them executable and have a link to a Google Colab. You could convert them into .py files using this template: https://github.com/meta-pytorch/forge/blob/main/docs/source/tutorial_sources/template_tutorial.py

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
init27: Thanks yes it was quite hard to get this right, for now I mentioned some high level details and acknowledge we cover them properly in part 2. 

Open to any other suggestions

docs/Tutorials/1_RL_and_Forge_Fundamentals.MD
init27: Thanks-yes, corrected it

docs/Tutorials/2_Forge_Internals.MD
init27: Sorry, this was my own confusion which I had to re read

init27:
@allenwang28 @ebsmothers-Sorry for being slow because of Flu, just addressed all comments, I think we are good to merge. 

Once we merge, I will work with @svekars to maybe create excalidraw diagrams for docs

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,622 @@
...
+```python
+# Your code sees this:
+responses = await policy.generate.route(prompt=prompt)
+```
+
+But behind the scenes:
+- `ServiceInterface` selects healthy replica
+- Routes message to that replica's `Policy.generate()` endpoint
+- Handles failures and retries automatically
+- Returns list[Completion] from the selected replica
```
felipemello1: fine to keep as is, just sharing that the first paragraph seems quite similar to the second, including "behind the scenes" mentioned twice. Maybe some minor rewording? but thats just a nit, i can understand and follow the content

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,622 @@
...
+```python
+responses = await policy.generate.route(prompt=question)
+answer = responses[0].text  # Extract text from Completion object
+```
+
+Behind the scenes:
+1. Health check eliminates failed replicas
+2. Load balancer picks replica (currently round robin, configurable balancers coming soon)
+3. Request routes to that specific replica
+4. Automatic retry on different replica if failure
```
felipemello1, line 177: seems similar to the above paragraphs, since route was used as the example before the deep dive

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,622 @@
...
+
+prompt = "What is 3 + 5?"
+responses = await policy.generate.route(prompt)
+print(f"Response: {responses[0].text}")
+
+# Cleanup when done
+await policy.shutdown()
+```
+
+### 3. How Services Actually Work
```
felipemello1, line 107: I think it would be nice to add here the same summary you put in part 1 about fanout, call_one, etc. Then the user can have a quick TLDR mental model, and then deep dive. An imagine might be worth a thousand words here too.

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,622 @@
...
+        self.value = 0
+
+counter_service = await ForgeCounter.options(
+    procs=1, num_replicas=4
+).as_service(initial_value=0)
+
+# Test basic operations
+await counter_service.increment.route()
+results = await counter_service.increment.fanout()  # Get from all replicas
+print(f"All replica values: {results}")
```
felipemello1, line 273: Can you add the output here as a comment?

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,622 @@
...
+
+# STICKY SESSIONS
+print("\nUsing sticky sessions:")
+async with counter_service.session():
```
felipemello1: it is the first time that "sticky session" is mentioned. Whats the difference between this and just calling `await counter_service.increment.route()` without a ctx manager?

docs/Tutorials/2_Forge_Internals.MD
felipemello1: Perhaps you could break it down into 2 coding blocks to display this output vs sticky session output

docs/Tutorials/2_Forge_Internals.MD
```diff
@@ -0,0 +1,622 @@
...
+```python
+# This breaks KV cache optimization:
+async def naive_multi_turn():
+    # Each call might go to different replica = cache miss
+    response1 = await policy_service.generate.choose(question1)
+    response2 = await policy_service.generate.choose(question1 + response1) # Cache miss!
+    response3 = await policy_service.generate.choose(conversation_so_far)   # Cache miss!
+```
+
+**The solution**: Sticky sessions ensure all calls go to same replica.
```
felipemello1, line 328: cool, now i understand sticky sessions. Not sure how to address that they are first introduced without explanation. Perhaps we could delete the first example above in favor of this?

docs/Tutorials/2_Forge_Internals.MD
init27: Thanks, I agree-made some minor re-wroding. Open to more suggestions!

docs/Tutorials/2_Forge_Internals.MD
init27: Removed, thanks

docs/Tutorials/2_Forge_Internals.MD
init27: added a TLDR and a quick image

docs/Tutorials/3_Monarch_101.MD
```diff
@@ -0,0 +1,439 @@
...
+    style ServiceActor fill:#FF9800
+    style ActorMesh fill:#9C27B0
+    style ProcMesh fill:#2196F3
+```
+
+## Deep Dive: ProcMesh - The Foundation
+
+**ProcMesh** is Monarch's core abstraction for organizing processes across hardware. Think of it as a multi-dimensional grid that maps directly to your cluster topology.
+
+### Single Host ProcMesh
```
felipemello1, line 51: I am not sure what the learning from this image is. It gives me a sense that GPUs are linearly connected. Should we replace it with a simple phrase? 
image

docs/Tutorials/3_Monarch_101.MD
```diff
@@ -0,0 +1,439 @@
...
+    style P1 fill:#F44336
+    style P2 fill:#F44336
+    style P3 fill:#F44336
+    style P4 fill:#F44336
+    style P5 fill:#F44336
+    style P6 fill:#F44336
+    style P7 fill:#F44336
+```
+
+### Multi-Host ProcMesh
```
felipemello1: Same here, i am not 100% sure what the learning is. Could you describe what you are trying to communicate?

image

docs/Tutorials/3_Monarch_101.MD
```diff
@@ -0,0 +1,439 @@
...
+value = await counters.get_value.call_one()
+print(f"One counter: {value}")
+
+# choose() - random single actor (actors only, not services)
+value = await counters.get_value.choose()
+print(f"Random counter: {value}")
+
+# call() - all actors, collect results
+values = await counters.get_value.call()
+print(f"All counters: {values}")
```
felipemello1: nit: i think it would be helpful to add as a comment what it prints

docs/Tutorials/3_Monarch_101.MD
```diff
@@ -0,0 +1,439 @@
...
+
+# broadcast() - fire and forget
+await counters.increment.broadcast()
```
felipemello1: nit: should we add a print here too? Or are you just trying to show that this is also an option, but the user should have understood what it does at this point?

docs/Tutorials/3_Monarch_101.MD
```diff
@@ -0,0 +1,439 @@
...
+
+# Cleanup
+await procs.stop()
+```
+
+## Actor Meshes: Your Code Running Distributed
```
felipemello1, line 153: i am not sure if the lines here are too helpful. Its fine to keep, but maybe a few bullet points and one example with text might be enough 
image

docs/Tutorials/3_Monarch_101.MD
```diff
@@ -0,0 +1,439 @@
...
+    H1P1 -.->|"InfiniBand"| H2P1
+    H2P0 -.->|"InfiniBand"| H3P0
+    H2P1 -.->|"InfiniBand"| H3P1
+
+    style PM1 fill:#F44336
+    style PM2 fill:#4CAF50
+    style PM3 fill:#2196F3
+```
+
+```python
```
felipemello1, line 94: nit about this example. Its fine to keep as is. When we run `this_proc().spawn`, i dont think that it triggers any of the forge things like in provisioner.py. So i wonder if users might try to add this type of API to their code, and it not work in forge, since its not a ForgeActor. So maybe its worth to make a note that Forge is using this under the hood, but if they want to build on Forge, they should leverage forge APIs, e.g. ForgeActor, .as_service, etc.

docs/Tutorials/3_Monarch_101.MD
init27: Actually yes I assumed so but agree, will add it

docs/Tutorials/3_Monarch_101.MD
init27: Replaced diagram with bullet points

docs/Tutorials/3_Monarch_101.MD
init27: Thanks-added comments to nuance this in line

init27:
Thanks so much for the detailed review and awesome guidance @allenwang28, @ebsmothers and @felipemello1. I've addressed all comments and merging now. 

I'll trouble you for examples soon now :)

PR #302, opened 2025-10-03
Rebranding Policy Implementation as a Generator Implementation

Jack-Khuu:
Everything boils down to these simple changes
* `policy.py` => `generator.py` 
* `Policy` => `Generator`

#### But why are there still mentions of `policy` in the repo?

That's because this PR intentionally does **NOT** change the assigned variables themselves. 
A Policy is the concrete concept, and in  this case it's just backed by a particular implementation. The yaml files are untouched for the same reason (it's configuring the Policy, which happens to be this `Generator` under the hood)

``` python
policy: Generator = Generator(...)
```

The `Generator` class has no mentions of policy because it's just a vllm generator (it can be used in other components like a Judge)

#### But what if we want other Generator implementations?

Great, we can rename this class when we actually need to (`VllmGenerator`, `Vllm` doesn't sound terrible)

#### "I don't like the name"

`Generator` > `Policy`, so a different name is not a blocker. 
If anyone gets more than 3 votes on a different name, we'll rename

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -21,7 +21,7 @@ from forge.actors._torchstore_utils import (
     get_dcp_whole_state_dict_key,
     get_param_prefix,
 )
-from forge.actors.policy import Policy
+from forge.actors.generator import Generator
 from forge.actors.reference_model import ReferenceModel
 from forge.actors.replay_buffer import ReplayBuffer
 from forge.actors.trainer import RLTrainer
@@ -329,7 +329,7 @@ async def main(cfg: DictConfig):
         reward_actor,
     ) = await asyncio.gather(
         DatasetActor.options(**cfg.actors.dataset).as_actor(**cfg.dataset),
-        Policy.options(**cfg.services.policy).as_service(**cfg.policy),
+        Generator.options(**cfg.services.policy).as_service(**cfg.policy),
         RLTrainer.options(**cfg.actors.trainer).as_actor(
             **cfg.trainer, loss=simple_grpo_loss
         ),
--- a/apps/toy_rl/sumdigits.py
+++ b/apps/toy_rl/sumdigits.py
@@ -17,7 +17,7 @@ import torch
 import torch.nn.functional as F
 import torchstore as ts
 from forge.actors._torchstore_utils import get_param_key
-from forge.actors.policy import Policy
+from forge.actors.generator import Generator
 from forge.actors.replay_buffer import ReplayBuffer
 from forge.actors.trainer import _qwen3_hf_to_vllm
 from forge.cli.config import parse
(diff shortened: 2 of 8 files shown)


allenwang28:
not a hill I'm going to die on, but if we wanted to introduce a Judge that re-uses our vLLM implementation, then we would do this:

```
class Judge(Generator):
    ...
```

does that make sense to others? Is there a world where we have a `vLLM` base actor and we have `Generator`, `Judge`, `RewardModel` based on that?

felipemello1:
regarding `VllmGenerator` . Is it realistic that someone in the community at some point will contribute with a `SGLangGenerator`?

Not for this PR, but if/when this happens, it makes sense to me to keep everything as `Generator`, untouched, and have it as a factory for `vLLMGenerator`, `SGLangGenerator`, etc.

allenwang28:
i'm not sure we should support any and all generators in Forge, but my point was slightly different. Does it make sense logically for a Judge to inherit from a Generator? It doesn't really to me - it makes more sense if they all inherit from some InferenceEngine or something like that

Jack-Khuu:
> if we wanted to introduce a Judge that re-uses our vLLM implementation, then we would do this: `class Judge(Generator):
`

Would we subclass? This might be a different question, but I thought forge/tune were intentionally avoiding subclassing
I would think Judge is 
```python
class Judge(ForgeActor):
  generator: GeneratorInterface  # Member (whether it be an instance ServiceInterface, self instantiated generator, etc.)
```

Edit: Discussed offline. We're gonna try sublassing

Jack-Khuu:
**Update (10/9):**
* Delta since last is [just a rebase] (tm)
* Regarding relation to other classes: `class Judge(Generator)` actually doesn't seem terribly bizarre at this time
  * Not in this PR, but we might even consider pushing it further with moving weight updates to `Policy(Generator)`

---

> Is there a world where we have a vLLM base actor and we have Generator, Judge, RewardModel based on that

> it makes sense to me to keep everything as Generator, untouched, and have it as a factory for vLLMGenerator, SGLangGenerator, etc.

Generator can definitely be converted to a factory under the hood when we need to support them (we could even have the generated instance from the factory be a `vLLM base actor` when we start offering different generation options

apps/grpo/main.py
```diff
@@ -345,7 +345,7 @@ async def main(cfg: DictConfig):
         reward_actor,
     ) = await asyncio.gather(
         DatasetActor.options(**cfg.actors.dataset).as_actor(**cfg.dataset),
-        Policy.options(**cfg.services.policy).as_service(**cfg.policy),
+        Generator.options(**cfg.services.policy).as_service(**cfg.policy),
```
pbontrager: I think this should still be Policy here, but it should be derived from Generator to make it clearer how to do judges. Probably just add a two line "class Policy(Generator):" like in my judges demo.

PR #303, opened 2025-10-03
[Logging] add time stamp logging + test

felipemello1:
### How to review this PR:
- I dont think you should review every file. Too many small changes.
- Main feedback should be on `logging_mode` and specially on the API for `backend.log_batch` and `backend.log_stream` 
--> I added comments for both in the PR

-------

### Summary of changes:
1) Added timestamp logging system for metrics tracking

This means we now enable zero reduce. The metrics are pushed immediately to the backend with a timestamp and step, and its up to the backend to buffer if they want to.

image

-------

2) Changed metric logging configs
Before:
```
reduce_across_ranks: bool
share_run_id: bool
```

After:
```
logging_mode: Enum[GLOBAL_REDUCE, PER_RANK_REDUCE, PER_RANK_NO_REDUCE]
per_rank_share_run: bool
```

It should be more intuitive what the options mean and we avoid adding a 3rd flag for no_reduce.

-------

3) unit tests for metrics.py and metric_actors.py

-------

4) Better actor naming. Now we get this:
image


src/forge/observability/metrics.py
```diff
@@ -395,68 +411,116 @@ def __init__(self):
...
-            self.accumulators[key] = reduction.accumulator_class(reduction)
+            self.accumulators[key] = metric.reduction.accumulator_class(
+                metric.reduction
+            )
+        self.accumulators[key].append(metric.value)
 
-        self.accumulators[key].append(value)
+        # For PER_RANK_NO_REDUCE backends: log immediately
+        for backend in self.per_rank_no_reduce_backends:
+            backend.log_immediate(metric=metric, step=self.step)
```
felipemello1: This is the major chunk that needs review, IMO.

1) We do **async** `log_batch` flush, i.e
```
for train_step in range(10):
	do_something()
    await mlogger.flush.call_one()
```
mlogger.flush calls MetricLogger.flush, which flushes all metrics accumulated during the train step.

-----

2) However, when `logging_mode=PER_RANK_NO_REDUCE`, there is no reduce, and MetricCollector flushes immediately on every push, i.e.
```
record_metric(key, value, reduce_type) # reduce is ignored
```

`record_metric` then calls `MetricCollector.push`, which calls backend.log_stream(key, value). Notice that this is **sync**.

------

3) I am not sure if this is the right API

async batch flush that calls backend.log
sync single metric flush that calls backend.log_stream

What happens when we have backend.log_table. Do we need .log_table_stream too?

src/forge/observability/metrics.py
```diff
@@ -5,17 +5,55 @@
...
+        and logs aggregated values on flush(). No cross-rank reduction.
+
+    - PER_RANK_NO_REDUCE: Backend is instantiated per-rank. Each rank logs raw metric values
+        immediately on each record_metric() call. Reduce type is **ignored**. Great alternative for
+        analyzing metrics per time stamp instead of per step.
+    """
+
+    GLOBAL_REDUCE = "global_reduce"
+    PER_RANK_REDUCE = "per_rank_reduce"
+    PER_RANK_NO_REDUCE = "per_rank_no_reduce"
```
felipemello1, line 54: is this intuitive?

src/forge/observability/utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+import logging
+from typing import Optional
+
+from monarch.actor import context, current_rank
+
+logger = logging.getLogger(__name__)
+
+
+def detect_actor_name_from_call_stack() -> str:
```
felipemello1, line 15: @allenwang28 , mind taking a look at this file?

I used inspect to get the class name from the call stack. I didnt find a good way to get the ActorName from the context(). I tried it in `get_actor_name_with_rank`, but it returns the local_fetcher, and not the actor.

felipemello1: I am getting this currently 
image

src/forge/observability/metrics.py
```diff
@@ -5,17 +5,60 @@
...
+
+    PER_RANK_NO_REDUCE = "per_rank_no_reduce"
+        Best for: Real-time streaming, time-series debugging
+        Behavior: Raw values logged immediately on record_metric() calls
+        Example use: See what every rank is doing in real time.
+    """
+
+    GLOBAL_REDUCE = "global_reduce"
+    PER_RANK_REDUCE = "per_rank_reduce"
+    PER_RANK_NO_REDUCE = "per_rank_no_reduce"
```
felipemello1, line 57: is this intuitive?

apps/grpo/main.py
```diff
@@ -363,6 +362,9 @@ async def main(cfg: DictConfig):
         ),
     )
 
+    # Call after services are initialized
```
JenniferWang: Would you maybe explain in the comment, why the `init_backends` should be called after services are initialized?

felipemello1: Calling **before** works for every mode except when 'per_rank_share_run=True'. Then it hangs. wandb says its experimental, and it didnt investigate it more deeply to see if i need to wait for something to finish. But i agree, i will add a note! Edit: done

src/forge/observability/metrics.py
```diff
@@ -499,13 +498,21 @@ def push(self, metric: Metric) -> None:
...
         if not self._is_initialized:
-            raise ValueError(
-                "MetricCollector was not initialized. This happens when you try to use `record_metric` "
-                "before you have initialized any logging backends. Please call in your main file:\n"
-                "`mlogger = await get_or_create_metric_logger(actor_name='Controller')`\n"
-                "`await mlogger.init_backends.call_one(logging_config)`\n"
-                "or, to disable metric logging globally, set env variable `FORGE_DISABLE_METRICS=True`"
+            from forge.util.logging import log_once
+
+            log_once(
```
felipemello1, line 502: @joecummings

apps/grpo/qwen3_32b.yaml
```diff
@@ -17,13 +17,15 @@ provisioner:
 rollout_threads: 1   # Recommended to set equal to policy.num_replicas
 
 # Observability configuration
+# logging_mode: global_reduce, per_rank_reduce, per_rank_no_reduce
```
joecummings: nit: Just put these besides the logging_mode option

apps/grpo/qwen3_32b.yaml
```diff
@@ -17,13 +17,15 @@ provisioner:
...
 metric_logging:
   wandb:
     project: "grpo-training"
     group: "grpo_exp_${oc.env:USER}"
-    reduce_across_ranks: True
+    logging_mode: "global_reduce"
```
joecummings:
```suggestion
    logging_mode: "global_reduce"  # "per_rank_reduce", "per_rank_no_reduce"
```

Per my above comment. Also, do they have to be in quotes??

apps/toy_rl/toy_metrics/main.py
```diff
@@ -82,31 +83,35 @@ async def main():
...
-            "share_run_id": False,  # Share run ID across ranks -- Not recommended.
+            "logging_mode": "per_rank_no_reduce",
+            "per_rank_share_run": False,
         },
     }
 
     service_config = {"procs": 2, "num_replicas": 2, "with_gpus": False}
-    mlogger = await get_or_create_metric_logger()
-    await mlogger.init_backends.call_one(config)
+    mlogger = await get_or_create_metric_logger(actor_name="Controller")
```
joecummings: What happens if you leave actor_name blank? For instance, this isn't actually an actor... feels a bit weird.

src/forge/env_constants.py
```diff
@@ -11,7 +11,7 @@
 
 # Force all timing methods in forge.observability.perf_tracker.py to use
 # CPU timer if False or GPU timer if True. If unset, defaults to the assigned value to the function.
-METRIC_TIMER_USES_CUDA = "METRIC_TIMER_USES_CUDA"
+METRIC_TIMER_USES_GPU = "METRIC_TIMER_USES_GPU"
```
joecummings, line 14: ?

src/forge/observability/metrics.py
```diff
@@ -5,17 +5,59 @@
...
+import pytz
+from monarch.actor import current_rank
+
+from forge.observability.utils import get_actor_name_with_rank
 
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class Role:
```
joecummings: Role is being used for Data - mauybe call this one MetricRole? Or anything else?

src/forge/observability/metrics.py
joecummings: hot damn, thank you

src/forge/observability/metrics.py
joecummings: Yeah I think this is about as good as were gunna get here. Thanks!

apps/toy_rl/toy_metrics/main.py
felipemello1: maybe i should rename it to "process_name" then

apps/grpo/main.py
allenwang28: can we debug this further instead of checking in this workaround?

src/forge/observability/metrics.py
felipemello1: I could rename it to BackendRole or maybe something better. Let me make it more obvious

src/forge/env_constants.py
felipemello1: making it future proof when we support other backends besides cuda

apps/toy_rl/toy_metrics/main.py
joecummings: I like `process_name`

src/forge/env_constants.py
joecummings: TPUs here we come

felipemello1: OH NO

felipemello1: "METRIC_TIMER_USES_ACCELERATOR"? Is this what torch uses? geez, we refactor it when the time comes

src/forge/observability/metrics.py
```diff
@@ -652,34 +768,66 @@ async def _init_shared_global(self):
...
+        # to connect to non-existent services. Clearing forces fresh service connection.
+        service_token.clear_service_in_env()
+
         settings = wandb.Settings(mode="shared", x_primary=False, x_label=self.name)
         self.run = wandb.init(
-            id=shared_id,
-            project=self.project,
-            group=self.group,
-            settings=settings,
+            id=shared_id, project=self.project, group=self.group, settings=settings
```
felipemello1, line 796: @allenwang28 @JenniferWang This fixes the wandb hang. Now we can initialize before or after services, no difference. 

tldr: it seems that spawning new processes copies environment variables used by wandb for shared mode, but they are not true for new runs. When spawning everything AFTER services, we didnt have this issue, because every new proc would have their own fresh vars.

apps/grpo/qwen3_8b.yaml
```diff
@@ -9,14 +9,16 @@ max_res_tokens: 512
 model: "Qwen/Qwen3-8B"
 off_by_n: 1 # Off by one by default
 
+# Observability configuration
```
joecummings: doble?

joecummings (approved):

src/forge/observability/metrics.py
```diff
@@ -594,44 +711,43 @@ def __init__(self, logger_backend_config: Dict[str, Any]):
...
-        elif role == "global" and self.share_run_id:
+        # Per-rank modes based on per_rank_share_run bool
+        elif role == BackendRole.GLOBAL and self.per_rank_share_run:
             await self._init_shared_global()
 
-        elif role == "local":
-            if self.share_run_id:
+        elif role == BackendRole.LOCAL:
+            if self.per_rank_share_run:
                 await self._init_shared_local(primary_logger_metadata)
```
DNXie, line 751: wouldn't this call simply overwrite `self.run` that is initialized in `self._init_shared_global()` and set the `x_primary` to False?

felipemello1: I am not sure what you mean by "overwrite". Are you saying because the backend would be called once with `backend.init(role=global, ...)` and then again `(role=local)`?

If so, then no. This should never happen. A backend is initialized only once per process.

What happens is that each process has a different **instance** of this class. So controller has a wandb backend, trainActor has another, etc.

global happens here: https://github.com/felipemello1/forge/blob/ece12d72d39120392ac679dc951125772157bfa6/src/forge/observability/metric_actors.py#L262

Local happens here: https://github.com/felipemello1/forge/blob/ece12d72d39120392ac679dc951125772157bfa6/src/forge/observability/metrics.py#L471

Let me know if thats not clear

DNXie: But here in GlobalLoggingActor.init_backends, it would call 
```
await backend.init(role=BackendRole.GLOBAL)
```
as you pointed out. And also 
```
fetcher.init_backends.call(
                    self.metadata_per_primary_backend, self.config
                )
```
https://github.com/felipemello1/forge/blob/ece12d72d39120392ac679dc951125772157bfa6/src/forge/observability/metric_actors.py#L276

In `LocalFetcherActor.init_backends` it would call `MetricCollector.init_backends`
https://github.com/felipemello1/forge/blob/ece12d72d39120392ac679dc951125772157bfa6/src/forge/observability/metric_actors.py#L167

Are they not the same backend instances? But with grpo/main, there is only one rank involved (rank0). Could you help me understand?

felipemello1: sure! We have:

1) Single `GlobalLoggingActor`.
2) Then, when every process is spawned, we spawn a `LocalFetcherActor` with it in [provisioner.py](https://github.com/felipemello1/forge/blob/ece12d72d39120392ac679dc951125772157bfa6/src/forge/controller/provisioner.py#L265).
So, if `TrainActor` has 2 replicas with 2 gpus each, we have 4 `LocalFetcherActor`.
3) Each process also has its own `MetricCollector`.
4) Each `MetricCollector` has its own backend. So in the example above, we have 4 backends, 1 per process, if you set `logging_mode="per_rank_reduce"|"per_rank_no_reduce"`. If you set "global_reduce", then no backend is instantiated in the ranks.
5) When we call `GlobalLoggingActor.init_backends` --> calls `LocalFetcherActor.init_backends`, --> calls `MetricCollector.init_backends` --> `local_backend.init`, tagged as "local".

----

So where is the "global" backend? This one is instantiated inside of the GlobalLoggingActor, and its the only "global" one: https://github.com/felipemello1/forge/blob/timestamp_logging/src/forge/observability/metric_actors.py#L262

So to recap:
- Each process has its own backend. N processes == N Backends instances
- And we have an extra one in global

felipemello1: If it makes it easier to visualize, imagine that instead of backend, it was just a `.json` that we write to. Every process writes to its own file. Global has its own file too.

global.json
r0.json
r1.json
Etc...

DNXie: Thanks, Felipe, for the explanation! It is much clear now.

felipemello1:
will break it down into 4 PRs

PR #304, opened 2025-10-03
Add Batch routing support via `@service_endpoint` with configurable batch size and timeout

DNXie:
Migrated from https://github.com/meta-pytorch/forge/pull/177
Context: https://github.com/meta-pytorch/forge/issues/160

Add batch routing to `Service` to improve request throughput and maintain session-aware routing.

-  Added new `@service_endpoint` decorator that supports routing configuration (`router`, `batch_size`, `batch_timeout`).
- Introduced `ServiceEndpointProperty` to distinguish between `@endpoint` and `@service_endpoint`.
- Centralized endpoint-to-router mapping in `Service` (`self.routers`) with support for both plain routers and batchers.
- Updated `ServiceInterface` to register endpoints through `_set_router`, ensuring consistent setup for both standard and service endpoints.
-  Extended `_call` and `_get_replica` to handle batch routing, session routing, and fallback routing in a unified way.
- Enhanced `Service.stop` to gracefully shut down any active batchers in addition to replicas.
- Added integration tests to validate:

  * Round-robin distribution with and without batching
  * Correct batch flushing when `batch_size` is reached
  * Independent coexistence of multiple endpoints with different batch sizes/routers

**Test**

```bash
pytest tests/unit_tests/test_router.py
```

### DIFF
--- a/src/forge/controller/service/endpoint.py
+++ b/src/forge/controller/service/endpoint.py
@@ -182,7 +182,7 @@ def service_endpoint(
 
     Example:
         class MyForgeActor(ForgeActor):
-            @service_endpoint(router=RoundRobinRouter(), batch_size=16, batch_timeout=0.05)
+            @service_endpoint(router=RoundRobinRouter, batch_size=16, batch_timeout=0.05)
             async def predict(self, x): ...
     """
 
--- a/src/forge/controller/service/router.py
+++ b/src/forge/controller/service/router.py
@@ -138,10 +138,8 @@ class Batcher:
             batch_timeout=0.01,
         )
 
-        request = ServiceRequest(...)
-
-        # Enqueue a request to be sent to a replica
-        await batcher.enqueue(request)
+        # Enqueue a endpoint call to be sent to a replica
+        results = await batcher.route(function, args, kwargs)
     """
 
     def __init__(
@@ -213,12 +211,6 @@ class Batcher:
                 except asyncio.TimeoutError:
                     break
 
-            session_map = self.get_session_map()
-            healthy_replicas = self.get_healthy_replicas()
-
-            # One routing decision for the whole batch
-            replica = self.inner_router.get_replica(healthy_replicas, None, session_map)
-
             # Merge args for batched call
             if batch and len(batch[0][1]) > 0:
                 # Normal case: endpoints expect positional arguments
@@ -228,29 +220,34 @@ class Batcher:
                 # No-arg case: just one batched call, no inputs to merge
                 args = ()
 
-            batch_req = ServiceRequest(
-                session_id=None,
-                function=self.function,
-                args=args,
-                kwargs={},
-                future=asyncio.Future(),
-            )
-
             for attempt in range(self._num_attempts):
+                session_map = self.get_session_map()
+                healthy_replicas = self.get_healthy_replicas()
+
+                # One routing decision for the whole batch
+                replica = self.inner_router.get_replica(
+                    healthy_replicas, None, session_map
+                )
+
+                batch_req = ServiceRequest(
+                    session_id=None,
+                    function=self.function,
+                    args=args,
+                    kwargs={},
+                    future=asyncio.Future(),
+                )
+
                 try:
                     # Send whole batch to replica
-                    logger.debug(
-                        f"[Batcher] enqueue request executing function={batch_req.function}, args={batch_req.args}, kwargs={batch_req.kwargs}"
-                    )
                     await replica.enqueue_request(batch_req)
                     results = await batch_req.future
-                    logger.debug(f"[Batcher] results: {results}")
                     # Normalize result shape
                     if isinstance(results, (list, tuple)):
                         results = list(results)
                         if len(results) != len(batch):
                             results = [results] * len(batch)
                     else:
+                        # scalar result, broadcast to batch size
                         results = [results] * len(batch)
 
                     # Fulfill each individual Future from the batch
@@ -263,13 +260,6 @@ class Batcher:
                     if attempt < self._num_attempts - 1 and not replica.healthy:
                         logger.debug(
                             f"Replica {replica.idx} failed during request, retrying on healthy replica. Exception: {e}"
-                        )
-                        healthy_replicas = self.get_healthy_replicas()
-                        session_map = self.get_session_map()
-                        if not healthy_replicas:
-                            raise RuntimeError("No healthy replicas available") from e
-                        replica = self.inner_router.get_replica(
-                            healthy_replicas, None, session_map
                         )
                         continue
                     else:
--- a/src/forge/controller/service/replica.py
+++ b/src/forge/controller/service/replica.py
@@ -9,7 +9,7 @@ import asyncio
 import logging
 import time
 from collections import deque
-from dataclasses import asdict, dataclass, field
+from dataclasses import dataclass, field
 from enum import Enum
 from typing import Optional
 


src/forge/controller/service/router.py
```diff
@@ -88,3 +105,178 @@ def get_replica(
...
         return replica
+
+
+class Batcher:
+    """
+    Asynchronous batching wrapper around a Router.
+
+    Instead of selecting a replica immediately, incoming requests are enqueued
+    and grouped into batches. Once a batch is ready (either reaching the maximum
+    size or exceeding the maximum wait time), the batcher makes a single routing
```
pbontrager, line 116: Is there a way to base the wait time on the status of the replica instead of a fixed time? For example, if the replica is still busy, we can let the batch grow larger, but if the replica is free for some minimum time interval, then we can send the batch.

DNXie: Definitely. We could make the batch timeout adaptive based on replica load (e.g., wait longer when replicas are busy and flush earlier when they’re idle). I’d prefer to land this current version first, then explore that as a follow-up improvement once the base batching logic is stable. Just added a TODO in the while loop.
```python
# TODO: make timeout adaptive based on replica load.
```

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+                # One routing decision for the whole batch
+                replica = self.inner_router.get_replica(
+                    healthy_replicas, None, session_map
+                )
+
+                batch_req = ServiceRequest(
+                    session_id=None,
+                    function=self.function,
+                    args=args,
+                    kwargs={},
```
DNXie, line 236: @allenwang28 Do we want to support kwargs here?

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+                    await replica.enqueue_request(batch_req)
+                    results = await batch_req.future
+                    # Normalize result shape
+                    if isinstance(results, (list, tuple)):
+                        results = list(results)
+                        if len(results) != len(batch):
+                            results = [results] * len(batch)
+                    else:
+                        # scalar result, broadcast to batch size
+                        results = [results] * len(batch)
```
DNXie, line 251: @allenwang28 Do we want to handle when the returned results have different length or a scalar?

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def __init__(
+        self,
+        service,
+        endpoint_name: str,
+    ):
+        self.service = service
+        self.endpoint_name = endpoint_name
+
+    async def route(self, *args: P.args, **kwargs: P.kwargs) -> R:
```
casteryh, line 41: Do we want to schedule the task to start immediately by making this return a future?
In particular we can fire and forget without `await`ing it - this is actually what happens with monarch native actor api iiuc.

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,199 @@
...
+            "Services only support route() and fanout()."
+        )
+
+    async def broadcast(self, *args: P.args, **kwargs: P.kwargs) -> List[R]:
+        raise NotImplementedError(
+            "You tried to use broadcast() on a service, not an actor. "
+            "Services only support route() and fanout()."
+        )
+
+    async def generate(self, *args: P.args, **kwargs: P.kwargs):
```
casteryh, line 139: ?

src/forge/controller/service/endpoint.py
```diff
@@ -0,0 +1,199 @@
...
```
casteryh, line 76: actor also doesn't have `generate`?

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+
+    def __init__(
+        self,
+        function: str,
+        inner_router: Router,
+        get_healthy_replicas: Callable[[], List["Replica"]],
+        get_session_map: Callable[[], Dict[str, int]],
+        batch_size: int = 16,
+        batch_timeout: float = 0.01,
+    ):
```
casteryh, line 153:
```suggestion
    def __init__(
        self,
        function: str,
        inner_router: Router,
        *,
        get_healthy_replicas: Callable[[], List["Replica"]],
        get_session_map: Callable[[], Dict[str, int]],
        batch_size: int = 16,
        batch_timeout: float = 0.01,
    ):
```

src/forge/controller/service/router.py
```diff
@@ -4,16 +4,33 @@
...
 
+class Router(ABC):
+    """Abstract base class for routing logic."""
+
+    @abstractmethod
+    def get_replica(
+        self,
+        healthy_replicas: List[Replica],
+        sess_id: str | None = None,
+        session_map: Dict[str, int] | None = None,
```
casteryh, line 28:
```suggestion
    def get_replica(
        self,
        healthy_replicas: List[Replica],
        *,
        sess_id: str | None = None,
        session_map: Dict[str, int] | None = None,
```

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+                    )
+                    nxt = await asyncio.wait_for(
+                        self._queue.get(), timeout
+                    )  # wait for timeout or until self._queue.get() finishes
+                    batch.append(nxt)
+
+                    if len(batch) >= self.batch_size:
+                        break
+                except asyncio.TimeoutError:
+                    break
```
casteryh, line 212:
```suggestion
            try:
                with asyncio.timeout(self.batch_timeout):
                    while len(batch) < self.batch_size:
                        item = await self._queue.get()
                        batch.append(item)
            except TimeoutError:
                pass
```

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+    Instead of selecting a replica immediately, incoming requests are enqueued
+    and grouped into batches. Once a batch is ready (either reaching the maximum
+    size or exceeding the maximum wait time), the batcher makes a single routing
+    decision using the inner router. All requests in that batch are then resolved
+    with the same replica.
+
+    This reduces router overhead by amortizing multiple requests into one decision.
+
+    Args:
+        inner_router: The underlying Router used to pick a replica.
```
casteryh, line 123: why do we want an inner_router to be part of the batcher?
it makes more sense for a batcher to be chained with a router instead of a batcher wraps a router.

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+                    )  # wait for timeout or until self._queue.get() finishes
+                    batch.append(nxt)
+
+                    if len(batch) >= self.batch_size:
+                        break
+                except asyncio.TimeoutError:
+                    break
+
+            # Merge args for batched call
+            if batch and len(batch[0][1]) > 0:
```
casteryh, line 215: I have no idea what's going on here. I assume you're trying to combine the args. but not clear how you are doing it.

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+                except Exception as e:
+                    if attempt < self._num_attempts - 1 and not replica.healthy:
+                        logger.debug(
+                            f"Replica {replica.idx} failed during request, retrying on healthy replica. Exception: {e}"
+                        )
+                        continue
+                    else:
+                        for _, _, _, f in batch:
+                            if not f.done():
+                                f.set_exception(e)
```
casteryh, line 268: can you factor out the retry/error handling logic to a separate function?

src/forge/controller/service/router.py
casteryh: I think both the batching logic and the unbatching logic should be factored out.
The batcher should accept two functions
`batch_fn :: [Request] => BatchRequest | Request`
`unbatch_fn :: BatchResponse | Response | => [Response]`

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+
+    async def route(self, function: str, args: tuple, kwargs: dict) -> Any:
+        """Add (args, kwargs) pair to queue, return a Future resolved when batch completes."""
+        # Queue the request for batching
+        fut = asyncio.Future()
+        self._queue.put_nowait((function, args, kwargs, fut))
```
casteryh, line 274: why `put_nowait`? if there is no capacity limit for the queue, there isn't much of a difference. if there is, this will throw an Error?

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+    size or exceeding the maximum wait time), the batcher makes a single routing
+    decision using the inner router. All requests in that batch are then resolved
+    with the same replica.
+
+    This reduces router overhead by amortizing multiple requests into one decision.
+
+    Args:
+        inner_router: The underlying Router used to pick a replica.
+        get_healthy_replicas: Callable that returns the current list of healthy replicas.
+        get_session_map: Callable that returns the session-to-replica mapping.
```
casteryh, line 125: @allenwang28 this is a question for you. why doesn't the SessionRouter own the session_map in the first place? Instead, we have to pass this down layer by layer?

src/forge/controller/service/router.py
```diff
@@ -88,3 +105,181 @@ def get_replica(
...
+                args = ()
+
+            for attempt in range(self._num_attempts):
+                session_map = self.get_session_map()
+                healthy_replicas = self.get_healthy_replicas()
+
+                # One routing decision for the whole batch
+                replica = self.inner_router.get_replica(
+                    healthy_replicas, None, session_map
+                )
```
casteryh, line 230: Since you only have one (batched) request in the first place, why do you need to keep track of the replica?

src/forge/controller/service/service.py
```diff
@@ -108,7 +110,10 @@ async def __initialize__(self):
 
         # Initialize the routers
         self._default_router = RoundRobinRouter()
-        self._session_router = SessionRouter(fallback_router=LeastLoadedRouter())
+        self._session_router = SessionRouter(fallback_router=self._default_router)
+
+        # This keeps the map between the registered endpoints and the routers
+        self.routers: dict[str, Router | Batcher] = {}
```
casteryh, line 116: I don't think Batcher should be considered a router. See above.

casteryh:
```suggestion
        self._routers: dict[str, Router] = {}
        self._batchers: dict[str, Batcher] = {}
```

tests/unit_tests/test_router.py
```diff
@@ -0,0 +1,401 @@
...
+    router = SessionRouter(fallback)
+
+    r1 = router.get_replica(replicas, sess_id="sess1", session_map=session_map)
+    r2 = router.get_replica(replicas, sess_id="sess2", session_map=session_map)
+
+    assert r1.idx != r2.idx
+    assert set(session_map.values()) == {0, 1}
+
+    # If LeastLoadedRouter as fallback, r1 and r2 should be assigned to same replicas
+    replicas = [make_replica(0, load=0), make_replica(1, load=5)]
```
casteryh, line 184: please test for behaviors.
e.g. if there is a bug in the code so that `self.active_requests` never gets incremented, this won't catch it.
also, if the implementation of LeastLoadedRouter gets changed, this test will unnecessarily break.

PR #307, opened 2025-10-03
Update spawn_actors API in sft v2 main.py

daniellepintz:
`spawn_actors` has been removed, so updating to use `as_actor` API.

Also some updates to the config to make it work.

Test Plan:
```
python -m apps.sft_v2.main --config apps/sft_v2/llama3_8b.yaml
```
Output:
```
[3] [ForgeSFTRecipe-3/8] 2025-10-03 13:14:24 INFO 91 / 1000|Loss: 1.11570405960083
[1] [ForgeSFTRecipe-1/8] 2025-10-03 13:14:24 INFO 92 / 1000|Loss: 0.9416455030441284
[2] [ForgeSFTRecipe-2/8] 2025-10-03 13:14:24 INFO 92 / 1000|Loss: 1.0203598737716675
[3] [ForgeSFTRecipe-3/8] 2025-10-03 13:14:24 INFO 92 / 1000|Loss: 0.8977944254875183
[6] [ForgeSFTRecipe-6/8] 2025-10-03 13:14:24 INFO 92 / 1000|Loss: 1.0252584218978882
```

### DIFF
--- a/apps/sft_v2/llama3_8b.yaml
+++ b/apps/sft_v2/llama3_8b.yaml
@@ -19,8 +19,6 @@ model:
   hf_assets_path: /tmp/Meta-Llama-3.1-8B-Instruct
 
 processes:
-  # scheduler: local # local | mast (not supported yet)
-  # hosts: 1
   procs: 8
   with_gpus: true
 
--- a/apps/sft_v2/main.py
+++ b/apps/sft_v2/main.py
@@ -280,16 +280,16 @@ class ForgeSFTRecipe(ForgeActor, ForgeEngine):
 async def run(cfg: DictConfig) -> None:
     logging.info("Spawing recipe...")
     process_cfg = cfg.pop("processes")
-    recipe = await ForgeSFTRecipe.options(**process_cfg).as_service(cfg)
+    recipe = await ForgeSFTRecipe.options(**process_cfg).as_actor(cfg)
 
     logging.info("Created recipe, running setup.")
-    await recipe.setup.fanout()
+    await recipe.setup.call()
 
     logging.info("Recipe has been setup. Training now.")
-    await recipe.train.fanout()
+    await recipe.train.call()
 
     logging.info("Done training. Clean up")
-    await recipe.cleanup.fanout()
+    await recipe.cleanup.call()
     await recipe.mesh.stop()
     logging.info("All done!")
 


apps/sft_v2/llama3_8b.yaml
```diff
@@ -14,11 +16,11 @@ comm:
 model:
   name: llama3
   flavor: 8B
-  tokenizer_path: /tmp/Llama-3.1-8B-Instruct
+  hf_assets_path: /tmp/Meta-Llama-3.1-8B-Instruct
 
 processes:
-  scheduler: local # local | mast (not supported yet)
-  hosts: 1
+  # scheduler: local # local | mast (not supported yet)
```
daniellepintz: does the `scheduler` api still exist? doesnt seem like it? i can also remove this

apps/sft_v2/llama3_8b.yaml
```diff
@@ -14,11 +16,11 @@ comm:
...
+  # hosts: 1
```
daniellepintz: is `hosts:1` also needed/valid for running locally?

apps/sft_v2/llama3_8b.yaml
```diff
@@ -1,9 +1,11 @@
+# >>> python -m apps.sft_v2.main --config apps/sft_v2/llama3_8b.yaml
+
 # Config for supervised full finetuning using a Llama3.1 8B Instruct model
 #
 # This config assumes that you've run the following command before launching
 # this run:
 #   export HF_HUB_DISABLE_XET=1
-#   uv run forge download meta-llama/Meta-Llama-3.1-8B-Instruct
+#   forge download meta-llama/Meta-Llama-3.1-8B-Instruct --output-dir /tmp/Meta-Llama-3.1-8B-Instruct
```
daniellepintz, line 8: I didn't have uv installed so this cmd didn't work. is `uv` necessary?

also, do we still need to do `export HF_HUB_DISABLE_XET=1`?

felipemello1: i tried 'HF_HUB_DISABLE_XET=1 about ~1 week ago and it broke. I dont think it works well with our devmachines. @joecummings  do you know?

apps/sft_v2/main.py
```diff
@@ -280,13 +280,7 @@ def __repr__(self) -> str:
...
     logging.info("Spawing recipe...")
     process_cfg = cfg.pop("processes")
-    recipe = await spawn_actors(
-        "sft",
-        ForgeSFTRecipe,
-        {"config": cfg},
-        process_cfg,
-        set_address=True,
-    )
+    recipe = await ForgeSFTRecipe.options(**process_cfg).as_service(cfg)
```
allenwang28: I think you can just use `.as_actor()` in this case

apps/sft_v2/llama3_8b.yaml
allenwang28: yeah let's remove it

apps/sft_v2/llama3_8b.yaml
allenwang28: not setting hosts will run it locally, setting it to 1 will try and run it remotely

apps/sft_v2/main.py
daniellepintz: thanks! does the `fanout` api below have to change then?

allenwang28: yeah you could just do .call()

PR #311, opened 2025-10-03
Provisioner cleanup 

allenwang28:
This PR implements a few cleanups after https://github.com/meta-pytorch/forge/pull/285

Changes:
- Introduces a utility for just `get_remote_info()` - given a host mesh, spawn a proc and a setup actor, get the host port and address of the hostmesh, and kill the proc. This happens regardless of the underlying launcher. Slight refactoring in the bootstrap
  - Launchers no longer return this remote info
- Monarch now complains (with AssertionError) if you try and create an empty workspace. This doesn't really apply for SLURM, so we just create a tempfile and pass it to the workspace 
- Gets rid of `controller.proc_mesh.get_proc_mesh` etc. and redirects everything to just use `provisioner.get_proc_mesh`
- Add host mesh <> proc tracking. I basically thought that I could do multiple `spawn_procs(...)` but I see that the prior procs die still. The intention was to enable the colocation of the vLLM controller and worker. We do this tracking because Monarch doesn't yet let you get the `host_mesh` from the `proc_mesh` (but it will eventually)
- Also adds the capability to provide a host mesh to `get_proc_mesh` - this way you can colocate things

### DIFF
--- a/src/forge/controller/provisioner.py
+++ b/src/forge/controller/provisioner.py
@@ -229,7 +229,6 @@ class Provisioner:
                 env_vars["MASTER_ADDR"] = addr
                 env_vars["MASTER_PORT"] = port
                 env_vars["CUDA_VISIBLE_DEVICES"] = ",".join(gpu_ids)
-                env_vars["HYPERACTOR_MESSAGE_DELIVER_TIMEOUT_SECS"] = "600"
                 env_vars["HYPERACTOR_MESSAGE_DELIVERY_TIMEOUT_SECS"] = "600"
                 env_vars["HYPERACTOR_CODE_MAX_FRAME_LENGTH"] = "1073741824"
 


src/forge/controller/launcher.py
```diff
@@ -32,28 +33,14 @@
     from torchx.specs import AppState
     from torchx.specs.fb.component_helpers import Packages
 except ImportError as e:
-    print(f"Warning: Monarch meta/fb inetrnal imports failed: {e}")
-    print("Monarch functionality will be limited")
+    # This means there is an erorr with MAST
```
ebsmothers: very useful comment 
```suggestion
    # This means there is an error with MAST
```

src/forge/controller/launcher.py
```diff
@@ -157,8 +145,7 @@ async def get_allocator(self, name: str, num_hosts: int) -> tuple[Any, Any, str]
         return alloc, None, server_name  # (Allocator, AllocConstraints, SeverName)
 
     async def remote_setup(self, procs: ProcMesh) -> tuple[str, int]:
-        setup = procs.spawn(f"setup-{uuid.uuid1()}", SetupActor)
-        return await setup.get_info.choose()
+        return
```
ebsmothers, line 152: noob q: why does this become a no-op?

src/forge/controller/launcher.py
vidhyav: nit: erorr -> error

why not log an error?

src/forge/controller/launcher.py
```diff
@@ -157,8 +145,7 @@ async def get_allocator(self, name: str, num_hosts: int) -> tuple[Any, Any, str]
         return alloc, None, server_name  # (Allocator, AllocConstraints, SeverName)
 
     async def remote_setup(self, procs: ProcMesh) -> tuple[str, int]:
```
vidhyav, line 151: Why do you need this at all?

src/forge/controller/provisioner.py
```diff
@@ -131,81 +158,87 @@ async def get_proc_mesh(
...
+                    host_mesh, server_name = await self.create_host_mesh(
+                        name=mesh_name,
+                        num_hosts=num_hosts,
+                    )
+                    host_id = uuid.uuid1()
+                    gpu_manager = GpuManager()
+                    self._host_gpu_map[host_id] = gpu_manager
+                    host_mesh._host_id = host_id
+                else:
+                    host_id = host_mesh._host_id
```
ebsmothers, line 208: Another noob q: where is this field defined? Took a quick look in Monarch and couldn't find it

src/forge/controller/provisioner.py
```diff
@@ -28,6 +28,32 @@
...
 
+def _get_port() -> str:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("localhost", 0))
+        addr = s.getsockname()
+        port = addr[1]
+        return str(port)
+
+
+class _SetupActor(Actor):
```
vidhyav: Why is this called setup actor when it is just getting host and port info?

vidhyav: Why do you need this class at all?

ebsmothers (approved):

src/forge/controller/provisioner.py
```diff
@@ -131,81 +158,87 @@ async def get_proc_mesh(
...
+        env_vars: dict[str, str] | None = None,
     ):
         """Gets a proc mesh.
 
         num_hosts = None implies that you want a local allocation, this may change.
 
         """
+        if env_vars is None:
+            env_vars = {}
+
```
vidhyav, line 189: Put this as the default value while passing in then?

src/forge/controller/launcher.py
allenwang28: ah so the remote setup is because MAST has a bunch of setup steps (like mount oilfs etc.) 

SLURM doesn't really need that so it's a no op. Originally it was just implemented as "get port/addr" but I've pulled that out to happen for anything that needs a GPU

src/forge/controller/launcher.py
allenwang28: because on SLURM runs this just becomes spammy

src/forge/controller/provisioner.py
allenwang28: I changed it to `_RemoteInfoFetcher`

this class is needed because we don't have a way to run a Python function on proc meshes

This all exists because I need to get remote information from the host like addr/port in order to initialize PyTorch distributed

src/forge/controller/provisioner.py
allenwang28: i did and linter yelled at me lol

src/forge/controller/provisioner.py
allenwang28: It's not - I'm just attaching this to the host mesh object. Not a real field

src/forge/controller/launcher.py
```diff
@@ -32,28 +33,14 @@
     from torchx.specs import AppState
     from torchx.specs.fb.component_helpers import Packages
 except ImportError as e:
-    print(f"Warning: Monarch meta/fb inetrnal imports failed: {e}")
-    print("Monarch functionality will be limited")
+    # This means there is an error with MAST
```
felipemello1, line 40: should it silently fail?

src/forge/controller/provisioner.py
```diff
@@ -131,81 +158,113 @@ async def get_proc_mesh(
...
+                This behavior may change in the future.
+            host_mesh: The host mesh to allocate the process on.
+                If None, a new host mesh will be created.
+            port: The distributed port to use.
+                If None, a port will be detected.
+            addr: The distributed address to use.
+                If None, an address will be detected.
+
+        Returns:
+            A proc mesh.
```
felipemello1: can we use the actual class, e.g. ProcMesh?

also add a typehint?

src/forge/controller/provisioner.py
ebsmothers: Sorry maybe I'm being dense.. won't L188 fail then? In the previous version you have host_mesh._host_id on the LHS of assignment statements, in that case your comment makes sense. But here you have it on the RHS and it's not a real attribute

src/forge/controller/launcher.py
Ritesh1905: For now, yes. (However some sort of logging would be idea.) until we figure out what would be a right way to segregate meta-only code blocks. These dependencies are meta internal and are installed via an internal fbpkg monarch build. The env setup for mast requires a separate installation script.

allenwang28: I've just added in a block that says "if imports failed and you're trying to use MAST, print an error messaging say imports failed and that you should check your build was correct"

src/forge/controller/provisioner.py
```diff
@@ -131,81 +158,113 @@ async def get_proc_mesh(
...
-                procs._port = port
-                procs._gpu_ids = gpu_ids
-            else:
-                procs = host_mesh.spawn_procs(per_host={"gpus": num_procs})
 
+                env_vars["MASTER_ADDR"] = addr
+                env_vars["MASTER_PORT"] = port
+                env_vars["CUDA_VISIBLE_DEVICES"] = ",".join(gpu_ids)
+                env_vars["HYPERACTOR_MESSAGE_DELIVER_TIMEOUT_SECS"] = "600"
+                env_vars["HYPERACTOR_MESSAGE_DELIVERY_TIMEOUT_SECS"] = "600"
```
Ritesh1905: Do you need these both? what's the difference between these two?

allenwang28: great catch! Removed the DELIVER which was a typo

PR #313, opened 2025-10-04
[Debugging] Create a stand alone trainer app

JenniferWang:
Very similar to the stand-alone vllm app, this trainer app is introduced to make investigating trainer OOM faster. This could be very useful for single-node trainer because you can run it locally and the system metrics are much easier to obtain.

## Test
Change the activation checkpointing config in `apps/grpo/qwen3_32b.yaml` and reproduce the OOM.  

```
  activation_checkpoint:
    mode: selective
    selective_ac_option: op
```
The repro is blazingly fast :) 

https://meta.wandb.io/jiyue/grpo-training/runs/8qe73q1b?nw=nwuserjiyue

### DIFF
--- a/apps/rl_trainer/main.py
+++ b/apps/rl_trainer/main.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-# Usage: python -m apps.trainer.main --config apps/grpo/qwen3_32b.yaml
+# Usage: python -m apps.rl_trainer.main --config apps/grpo/qwen3_32b.yaml
 
 import asyncio
 


apps/trainer/main.py
```diff
@@ -0,0 +1,237 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Usage: python -m apps.trainer.main --config apps/grpo/qwen3_32b.yaml
```
allenwang28: nit, can we re-name this to rl_trainer? The reason is because we have sft, sft_v2 and then if we have a separate trainer it may confuse even further lol

PR #314, opened 2025-10-05
[WIP] enable rdma for weight sync

casteryh:
this needs relevant torchstore update to be landed first for the monarch api change, and we also have to target a newer version of monarch.

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -112,7 +112,7 @@ class RLTrainer(ForgeActor):
     # Non JobConfig-related fields
     loss: Callable = lambda logits, **targets: logits
     state_dict_key: str = "model_state_dict"
-    use_dcp: bool = True
+    use_dcp: bool = False
     dcp_path: str = "forge_dcp_tmp"
 
     def __post_init__(self):


src/forge/actors/trainer.py
```diff
@@ -403,7 +403,8 @@ async def push_weights(self, policy_version: int) -> None:
         else:
             for name, param in hf_state_dict.items():
                 key = get_param_key(policy_version, name)
-                await ts.put(key, param)
+                # RDMA is still broken on GPU, so we need to copy to CPU
+                await ts.put(key, param.detach().cpu())
```
allenwang28, line 340: is this a bad thing? I thought we were writing it to CPU memory anyways on the trainer put side

casteryh: theoretically we don't need this extra copy if RDMA on GPU is working. yes we are writing to cpu memory, but currently it is local gpu-> local cpu-> remote cpu.

vidhyav: 1) Why is the GPU RDMA not working?

2) What the perf penalty is with the GPU-CPU copy?

3) Also, what's the corresponding access on the read side?

casteryh: > 1. Why is the GPU RDMA not working?

Memory registration error. This could be a build issue but I doubt we have enough time to debug. CPU should work now.

> 2. What the perf penalty is with the GPU-CPU copy?

Not sure, need profiling. I'd guess anywhere between 30% - 100% increased latency.

> 3. Also, what's the corresponding access on the read side?

On the read side, it's basically remote cpu -> local cpu -> gpu (vllm worker).

PR #315, opened 2025-10-05
Add a check to not build docs in forks

mirceamironenco:
Just a small QOL change to not run `build-docs` in forks - should prevent `deploy-docs` from running and sending an email on failure whenever main is synced.


joecummings (approved):
@mirceamironenco !!!! Hello again !!!!

Thanks for the PR :)

PR #319, opened 2025-10-06
Moves various apps to tests/sandbox

allenwang28:
As mentioned in chat, also makes toy_rl run again. Several of these should also be turned into smoke tests, mast should probably be kept internal only.

Want approval from:
- @Ritesh1905  for toy_rl
- @Jack-Khuu @pbontrager for vllm
- @JenniferWang for rl_trainer


JenniferWang (approved):
LGTM

PR #323, opened 2025-10-07
Log KL Divergence in GRPO Loss function

### LINKED ISSUE #298: Log KL Divergence in GRPO Loss function

krammnic:
Resolves #298

### DIFF
--- a/src/forge/losses/grpo_loss.py
+++ b/src/forge/losses/grpo_loss.py
@@ -20,7 +20,9 @@ class SimpleGRPOLoss(nn.Module):
         super().__init__()
         self.beta = beta
 
-    def forward(self, logprobs, ref_logprobs, advantages, padding_mask):
+    def forward(
+        self, logprobs, ref_logprobs, advantages, padding_mask
+    ) -> tuple[torch.tensor, LossMetrics]:
         kl = torch.exp(ref_logprobs - logprobs) - (ref_logprobs - logprobs) - 1
 
         per_token_policy_loss = torch.exp(logprobs - logprobs.detach()) * advantages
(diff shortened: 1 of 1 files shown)


src/forge/actors/trainer.py
```diff
@@ -179,7 +181,7 @@ async def setup(self):
 
     def forward_backward(
         self, inputs: dict[str, Tensor], targets: dict[str, Tensor]
-    ) -> Tensor:
+    ) -> Tensor | LossMetrics:
```
casteryh, line 184: shouldn't this be `tuple[..,..]`?

krammnic: It should!

krammnic:
@casteryh let's merge

src/forge/losses/grpo_loss.py
```diff
@@ -7,23 +7,26 @@
...
         https://github.com/huggingface/trl/blob/417915a3e4d3e3bc8d7b196594308b8eabf928be/trl/trainer/grpo_trainer.py#L1624.
     """
 
-    def __init__(self, beta: float = 0.1):
+    def __init__(self, beta: float = 0.1) -> tuple[torch.Tensor, LossMetrics]:
         super().__init__()
         self.beta = beta
 
     def forward(self, logprobs, ref_logprobs, advantages, padding_mask):
         kl = torch.exp(ref_logprobs - logprobs) - (ref_logprobs - logprobs) - 1
```
joecummings, line 26: Can we log the KL divergence minus padding tokens? May have to move that op up in the loss function.

src/forge/losses/grpo_loss.py
```diff
@@ -7,23 +7,26 @@
 import torch
 from torch import nn
 
+from forge.data_models.loss_metrics import LossMetrics
```
joecummings, line 10: I'm not sure this is a fully fleshed out data model we want to use. 

For now could we just define a loose type in this file and shove the metrics in that?

krammnic: I've done it with data_model, because we might want to log some other things from different losses in future (margins from DPO loss for instance).

src/forge/losses/grpo_loss.py
krammnic: Yep good idea

krammnic: Adressed

joecummings:
We landed this !

PR #333, opened 2025-10-07
minor bug fix and config changes

Ritesh1905:
- updates mast configs 
- minor bugfix in the launcher

### DIFF
--- a/src/forge/controller/launcher.py
+++ b/src/forge/controller/launcher.py
@@ -299,9 +299,9 @@ class Mastlauncher(BaseLauncher):
 def get_launcher(cfg: LauncherConfig | None = None) -> BaseLauncher | None:
     if not cfg:
         return None
-    if cfg.launcher == Launcher.SLURM.value:
+    if cfg.launcher == Launcher.SLURM:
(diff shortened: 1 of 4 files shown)


ebsmothers (approved):

src/forge/controller/launcher.py
```diff
@@ -299,9 +299,9 @@ def create_server_handle(self) -> str:
 def get_launcher(cfg: LauncherConfig | None = None) -> BaseLauncher | None:
     if not cfg:
         return None
-    if cfg.launcher == Launcher.SLURM.value:
+    if cfg.launcher == Launcher.SLURM:
```
allenwang28, line 302: what error message do you get with `.value`?  I was seeing an error message this proposed way

Ritesh1905: I did not see any error message. When running for mast, this will throw a `ValueError` since cfg.launcher is of type `Launcher`, this would never be true, since we are compareing string with a `Launcher`

the other way was just use `cfg.launcher.value == Launcher.SLURM.value:`

PR #334, opened 2025-10-07
GPU CI

ebsmothers:
this PR is the culmination of a long and sad story. but we can now run forge tests on GPU runners

Still requires some follow-ups, but it works:

Screenshot 2025-10-14 at 9 42 43 AM


joecummings:
hey

ebsmothers:
> hey

~don't look at me~ actually though this is ready for review

.github/workflows/gpu_test.yaml
```diff
@@ -0,0 +1,67 @@
...
+          activate-environment: test
+          python-version: ${{ matrix.python-version }}
+      - name: Update pip
+        run: python -m pip install --upgrade pip
+      - name: Install pinned torch nightly
+        run: python -m pip install --pre torch==2.9.0.dev20250905 --no-cache-dir --index-url https://download.pytorch.org/whl/nightly/cu129
+      - name: Download and install vLLM and its dependencies
+        # TODO: this honestly could not be hackier if I tried
+        run: |
+          python -m pip install -r .github/packaging/vllm_reqs.txt
```
joecummings, line 48: this seems stable

ebsmothers:
OK at least in the run https://github.com/meta-pytorch/forge/actions/runs/18421419301?pr=334 all unit tests but 3 pass. Two look like pretty normal failures, the only reasonably suspicious one is the linker-related error in `TestProvisionerCudaVisibleDevices::test_get_proc_mesh_respects_cuda_visible_devices`. Tagging @allenwang28 as the author of that test in case you have any ideas on the cause

.github/workflows/gpu_test.yaml
```diff
@@ -0,0 +1,67 @@
...
+          python -m pip install vllm==0.10.1.dev0+g6d8d0a24c.d20251009.cu129 --no-cache-dir --index-url https://download.pytorch.org/whl/preview/forge
```
allenwang28, line 49: if we were to put these into folders by day and have a latest, then could we just do something like this?
```
python -m pip install vllm --no-cache-dir --index-url https://download.pytorch.org/whl/preview/forge/latest
```

pbontrager, line 49: Could these dependencies go in the pyproject toml?

ebsmothers: Ideally yes but need to check on the ability to add new subdirectories from devinfra's point of view. cc @atalman on the best approach here

ebsmothers: leaving cleanup as a follow-up. most important thing here is to get this working imo. after that i will open some other smaller refactor PRs (e.g. @allenwang28 was talking about consolidation with the install script as well).

PR #337, opened 2025-10-07
[Memory optm] loss using torch + compile

felipemello1:
Memory freebies
image

------
I dont think that loss/reward is a good way to check correctness here. But i compared the functions locally and they provide the same output.

image

image

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -160,6 +160,10 @@ class RLTrainer(ForgeActor):
         }
         os.environ.update(env)
 
+        # compile loss
+        logger.info("Compiling loss")
+        self.loss = torch.compile(self.loss)
+
     @endpoint
     async def setup(self):
         # TODO: update ForgeEngine to not use ForgeJobConfig
--- a/src/forge/util/ops.py
+++ b/src/forge/util/ops.py
@@ -56,6 +56,7 @@ def compute_logprobs(
 ) -> torch.Tensor:
     """
     Computes the log probabilities of the input tokens given the model logits and temperature.
+    Always converts inputs to fp32 for numerical stability
 
     Args:
         logits (`torch.Tensor`):
@@ -65,10 +66,23 @@ def compute_logprobs(
         temperature (`float`, *optional*, defaults to 1.0):
             The temperature value for scaling logits before computing log probabilities.
 
+    Returns:
+        logprobs: [batch, seq_len] log probabilities for each token
     """
     # Ignore the last token from logits because it predicts the next token (-1)
     # And align logits with the input tokens length.
     logits = logits[:, -input_ids.size(1) - 1 : -1, :].to(input_ids.device)
     scaled_logits = logits / temperature
-    logprobs = selective_log_softmax(scaled_logits, input_ids)
-    return logprobs
+
+    # Convert to fp32 for numerical stability
+    scaled_logits_fp32 = scaled_logits.float()
+
+    # get per-token log probs
+    batch_size, seq_len, vocab_size = scaled_logits_fp32.shape
+    log_probs = -F.cross_entropy(
+        scaled_logits_fp32.reshape(-1, vocab_size),
+        input_ids.reshape(-1),
+        reduction="none",
+    )
+
+    return log_probs.reshape(batch_size, seq_len)


src/forge/util/ops.py
```diff
@@ -65,10 +66,23 @@ def compute_logprobs(
...
     """
     # Ignore the last token from logits because it predicts the next token (-1)
     # And align logits with the input tokens length.
     logits = logits[:, -input_ids.size(1) - 1 : -1, :].to(input_ids.device)
     scaled_logits = logits / temperature
-    logprobs = selective_log_softmax(scaled_logits, input_ids)
-    return logprobs
+
+    # Convert to fp32 for numerical stability
+    scaled_logits_fp32 = scaled_logits.float()
```
JenniferWang, line 78: Noob question: what's the dtype for `scaled_logits`?

felipemello1: float becomes torch.float32

felipemello1:
@ebsmothers @Jack-Khuu @joecummings @pbontrager can some of you confirm that i dont need to do the all_gather that was happening in `selective_log_softmax`? Maybe its necessary if `trainer.parallelism.disable_loss_parallel=False`? But this is off for all of our configs.

src/forge/util/ops.py
```diff
@@ -65,10 +66,23 @@ def compute_logprobs(
...
+        logprobs: [batch, seq_len] log probabilities for each token
     """
     # Ignore the last token from logits because it predicts the next token (-1)
     # And align logits with the input tokens length.
     logits = logits[:, -input_ids.size(1) - 1 : -1, :].to(input_ids.device)
     scaled_logits = logits / temperature
-    logprobs = selective_log_softmax(scaled_logits, input_ids)
-    return logprobs
+
+    # Convert to fp32 for numerical stability
```
JenniferWang:
```suggestion
    # Cast up to fp32 for numerical stability
```

src/forge/util/ops.py
JenniferWang, line 11: Should we also delete this function?

casteryh:
<del> just curious, if we do torch.compile on the textbook implementation, do we still need manual optimization like `selective_log_softmax`? </del>
never mind, just  remove the `selective_log_softmax` all together.

src/forge/actors/trainer.py
```diff
@@ -160,6 +160,10 @@ def __post_init__(self):
         }
         os.environ.update(env)
 
+        # compile loss
+        logger.info("Compiling loss")
+        self.loss = torch.compile(self.loss)
```
joecummings, line 165: Is there any circumstance under which this command would fail?

src/forge/util/ops.py
```diff
@@ -65,10 +66,23 @@ def compute_logprobs(
...
     scaled_logits = logits / temperature
-    logprobs = selective_log_softmax(scaled_logits, input_ids)
-    return logprobs
+
+    # Cast up to fp32 for numerical stability
+    scaled_logits_fp32 = scaled_logits.float()
+
+    # get per-token log probs
+    batch_size, seq_len, vocab_size = scaled_logits_fp32.shape
+    log_probs = -F.cross_entropy(
```
joecummings:
```suggestion
    logprobs = -F.cross_entropy(
```

I will die on this hill.

src/forge/util/ops.py
```diff
@@ -65,10 +66,23 @@ def compute_logprobs(
...
+        logprobs: [batch, seq_len] log probabilities for each token
     """
     # Ignore the last token from logits because it predicts the next token (-1)
     # And align logits with the input tokens length.
     logits = logits[:, -input_ids.size(1) - 1 : -1, :].to(input_ids.device)
     scaled_logits = logits / temperature
-    logprobs = selective_log_softmax(scaled_logits, input_ids)
-    return logprobs
+
+    # Cast up to fp32 for numerical stability
```
joecummings, line 77: nit: I would change this to something like "ensure logits are in fp32" b/c they actually could already be in fp32 and no need for "Casting up"

src/forge/actors/trainer.py
felipemello1: cant think of one in our scenario, but if/when this happens, we can fix it

src/forge/util/ops.py
felipemello1: its used in 3 other places. I will leave it there for now. We will prob need some larger refactoring later to clean up / organize losses

PR #338, opened 2025-10-07
[DEMO][Do NOT LAND] Example Judge Implementation

pbontrager:
This shows how you can add a judge into the workflow. This is a very basic example that just replaces the RewardActor with the Judge and the logic around the judge is very basic. The primary goal of this PR is as an example.


Jack-Khuu:
I'll spin this into test/sandbox with some additional context going into PTC

(Guided Decoding doesn't work OOTB for CoT since it neuters the prob distribution so it doesn't think)
- Supposedly vllm supports this in v0; Will need to dig into how well supported it is in vllm v1

PR #345, opened 2025-10-08
Metric Logging updates 1/4

felipemello1:
1. Env var `FORGE_DISABLE_METRICS` now disables spawning `LocalFetcherActor` when spawning processes
2. Before, when calling `record_metrics(...)`, if `GlobalLoggingActor.init_backends` was not initialized, it would raise errors. Now it just prints a warning. This enables stand alone components to be tested without metric logging.
3. In the mode where wandb runs are shared, it would hang when `init_backend` was called AFTER initializing the other services. Calling `service_token.clear_service_in_env()` in shared mode fixes this issue.
4. Add unit tests for these changes and also general tests for metrics

### DIFF
--- a/src/forge/controller/provisioner.py
+++ b/src/forge/controller/provisioner.py
@@ -21,7 +21,6 @@ from monarch.tools import commands
 from forge.controller.launcher import BaseLauncher, get_launcher
 
 from forge.env_constants import FORGE_DISABLE_METRICS
-from forge.observability.metric_actors import get_or_create_metric_logger
 
 from forge.types import ProcessConfig, ProvisionerConfig
 
@@ -265,6 +264,8 @@ class Provisioner:
 
         # Spawn local fetcher actor on each process and register with global logger
         if os.getenv(FORGE_DISABLE_METRICS, "false").lower() != "true":
+            from forge.observability.metric_actors import get_or_create_metric_logger
+
             _ = await get_or_create_metric_logger(procs)
         return procs
 
@@ -286,6 +287,10 @@ class Provisioner:
         async with self._lock:
             # Deregister local logger from global logger
             if hasattr(proc_mesh, "_local_fetcher"):
+                from forge.observability.metric_actors import (
+                    get_or_create_metric_logger,
+                )
+
                 global_logger = await get_or_create_metric_logger(proc_mesh)
                 await global_logger.deregister_fetcher.call_one(proc_mesh)
 


src/forge/observability/metrics.py
```diff
@@ -437,7 +437,21 @@ async def init_backends(
 
     def push(self, key: str, value: Any, reduction: Reduce = Reduce.MEAN) -> None:
         if not self._is_initialized:
-            raise ValueError("Collector not initialized—call init first")
+            from forge.util.logging import log_once
```
allenwang28: should we import at the toplevel?

allenwang28 (commented):
thanks Felipe!

src/forge/observability/metrics.py
```diff
@@ -662,6 +682,15 @@ async def _init_shared_local(self, primary_metadata: Dict[str, Any]):
             raise ValueError(
                 f"Shared ID required but not provided for {self.name} backend init"
             )
+
+        # Clear any stale service tokens that might be pointing to dead processes
+        # In multiprocessing environments, WandB service tokens can become stale and point
+        # to dead service processes. This causes wandb.init() to hang indefinitely trying
+        # to connect to non-existent services. Clearing forces fresh service connection.
+        from wandb.sdk.lib.service import service_token
```
allenwang28, line 690: should this also be imported at the top?

felipemello1: I dont think so. My opinion is that backends like wandb should be protected, otherwise user is required to have wandb even if they dont use it. Extrapolate that to mlflow, scuba, etc. Let me know if you disagree.

src/forge/controller/provisioner.py
```diff
@@ -262,8 +262,11 @@ def bootstrap(env: dict[str, str]):
 
             self._proc_host_map[procs] = host_mesh
 
-        # Spawn local logging actor on each process and register with global logger
-        _ = await get_or_create_metric_logger(procs)
+        # Spawn local fetcher actor on each process and register with global logger
+        if os.getenv(FORGE_DISABLE_METRICS, "false").lower() != "true":
+            from forge.observability.metric_actors import get_or_create_metric_logger
```
felipemello1, line 267: had to also protect import here to avoid circular import issues

```
ImportError while loading conftest '/home/felipemello/forge/tests/unit_tests/observability/conftest.py'.
tests/unit_tests/observability/conftest.py:12: in <module>
    from forge.observability.metrics import LoggerBackend, MetricCollector
src/forge/observability/__init__.py:7: in <module>
    from .metric_actors import (
src/forge/observability/metric_actors.py:15: in <module>
    from forge.observability.metrics import (
src/forge/observability/metrics.py:16: in <module>
    from forge.util.logging import log_once
src/forge/util/__init__.py:8: in <module>
    from .metric_logging import get_metric_logger
src/forge/util/metric_logging.py:11: in <module>
    from forge.interfaces import MetricLogger
src/forge/interfaces.py:12: in <module>
src/forge/controller/__init__.py:6: in <module>
    from .actor import ForgeActor
src/forge/controller/actor.py:15: in <module>
    from forge.controller.provisioner import get_proc_mesh, stop_proc_mesh
src/forge/controller/provisioner.py:24: in <module>
    from forge.observability.metric_actors import get_or_create_metric_logger
E   ImportError: cannot import name 'get_or_create_metric_logger' from partially initialized module 'forge.observability.metric_actors' (most likely due to a circular import) (/home/felipemello/forge/src/forge/observability/metric_actors.py)
```

allenwang28: kk

PR #346, opened 2025-10-08
Test vLLM prefix caching correctness

joecummings:


tests/integration_tests/test_vllm_policy_correctness.py
```diff
@@ -0,0 +1,247 @@
...
+
+    Namely we want to check two things:
+    1. KV cache is populated correctly.
+    2. KV cache is cleared correctly.
+
+    Our main tool to inspect the KV cache is the `num_cached_tokens` field in the request output.
+    According to the vLLM docs (https://docs.vllm.ai/en/v0.9.0/api/vllm/outputs.html#vllm.outputs.RequestOutput),
+    this is the number of tokens with a prefix cache hit. So, the logic is that if we run one generation,
+    then run another generation with the same start, we should see the number of cached tokens == the length of the prefix.
+
```
JenniferWang, line 119: TIL

src/forge/actors/policy.py
```diff
@@ -500,6 +500,10 @@ async def update_weights(self, policy_version: int):
 
         logger.info(f"Weight update completed (now v{self.policy_version})")
 
+    @endpoint
+    async def _reset_prefix_cache(self):
```
JenniferWang, line 504: Maybe we can have this convention of 


```suggestion
    async def _test_reset_prefix_cache(self):
```

Jack-Khuu, line 504: Another example of needing
https://github.com/meta-pytorch/monarch/issues/1455

tests/integration_tests/test_vllm_policy_correctness.py
```diff
@@ -0,0 +1,247 @@
...
+        vllm_outputs = []
+        policy_outputs = []
+
+        # Exactly 16 tokens to fill up 1 block
+        first_prompt = (
+            "The paged prefix caching mechanism in vLLM is an interesting approach."
+        )
+        expected_cached_tokens = 0
+        async for res in vllm_model.generate(
+            first_prompt, sampling_params, request_id="first_16"
```
Jack-Khuu, line 181: Does the request_id matter here?

joecummings: Nah, just for logging.

src/forge/actors/policy.py
joecummings: In this case, I actually don't mind exposing to the end user! It could be used if someone wants to do something custom with their Policy setup.

PR #348, opened 2025-10-08
build with CUDA 12.6

ebsmothers:


scripts/install.sh
```diff
@@ -261,7 +261,7 @@ main() {
     download_vllm_wheel
 
     log_info "Installing PyTorch nightly..."
-    pip install torch==$PYTORCH_VERSION --index-url https://download.pytorch.org/whl/nightly/cu129
+    pip install torch==$PYTORCH_VERSION --index-url https://download.pytorch.org/whl/nightly/cu126
```
allenwang28: leave this as is for now

PR #357, opened 2025-10-09
Auto-track and globally shut down all Forge actors and services

DNXie:
Context: https://github.com/meta-pytorch/forge/issues/360

This PR introduces **centralized allocation tracking** and **unified shutdown** for all `ForgeActor` and `ServiceInterface` instances managed by the global `Provisioner`.

Actors and services are now **automatically registered** with the provisioner when spawned via `.as_actor()` or `.as_service()`.
This enables a single, global shutdown sequence (`await shutdown()`) that gracefully terminates all registered allocations in reverse order, no manual teardown required.

#### **Automatic Registration**

* Added `register_service()` / `register_actor()` to the `Provisioner` with type checks.
* Added matching top-level helpers for API consistency (`get_proc_mesh`, `stop_proc_mesh`).
* `ForgeActor.as_service()` and `.as_actor()` now automatically register their proxies after initialization.

#### **Centralized Shutdown**

* Added `Provisioner.shutdown_all_allocations()` for unified teardown.
* Gracefully shuts down all tracked `ServiceInterface` and `ForgeActor` instances in reverse allocation order.
* Move metric logger shutdown into the centralized function `shutdown()`

### Test:
```bash
```
The log after ctrl+C (cleaned version):
```bash
Shutting down...
Shutting down metric logger...
Shutting down provisioner..
Shutting down 3 service(s) and 4 actor(s)...
Health loop stopped gracefully.
Health loop stopped gracefully.
Shutdown completed successfully
```

### DIFF
--- a/src/forge/controller/actor.py
+++ b/src/forge/controller/actor.py
@@ -8,9 +8,12 @@ import logging
 
 import math
 import sys
-from typing import Any, Type, TypeVar
+from typing import Any, Type, TYPE_CHECKING, TypeVar
 
 from monarch.actor import Actor, current_rank, current_size, endpoint
+
+if TYPE_CHECKING:
+    from monarch._src.actor.actor_mesh import ActorMesh
 
 from forge.controller.provisioner import (
     get_proc_mesh,


src/forge/controller/actor.py
```diff
@@ -185,13 +198,16 @@ async def as_actor(cls: Type[T], *args, **actor_kwargs) -> T:
...
         return actor
 
     @classmethod
-    async def shutdown(cls, actor: "ForgeActor"):
+    async def shutdown(cls, actor: "ForgeActor", quiet: bool = False):
         """Shuts down an actor.
         This method is used by `Service` to teardown a replica.
         """
+        if not quiet:
+            logger.info(f"Shutting down actor {getattr(actor, 'name', cls.__name__)}")
```
DNXie: Adding this quiet check because otherwise when we shutdown a service, it would call actor.shutdown and print the log twice.

src/forge/controller/provisioner.py
```diff
@@ -302,8 +306,40 @@ async def stop_proc_mesh(self, proc_mesh: ProcMesh):
                 commands.kill(server_name)
             del self._proc_host_map[proc_mesh]
 
+    async def track_allocation(self, alloc: Any):
+        """Tracks an allocation for cleanup."""
+        self._allocations.append(alloc)
```
allenwang28: hmmm, I think an even simpler approach is to just track the proc meshes right? We can just do `await proc_mesh.stop()` and I think everything inside of it should shut down neatly. Let me know if that doesn't work though

DNXie: I think it would work for actor since[ shutting down actor is essentially stopping the `proc_mesh`](https://github.com/meta-pytorch/forge/blob/main/src/forge/controller/actor.py#L191-L197).
But for [service](https://github.com/meta-pytorch/forge/blob/main/src/forge/controller/service/service.py#L487-L510), it involves some other operations such as stopping the replicas and healthy loop.

DNXie:
> ok, I understand - here are a few suggestions:
> 
> * In provisioner.py, let's add in a `register_service` and `register_actor` on the Provisioner
> * Also add top level `register_service` and `register_actor` functions (like `get_proc_mesh`)
> * Within `.as_service(...)` and `.as_actor(...)` we call `register_service` and `register_actor` respectively

@allenwang28 Done!

src/forge/controller/actor.py
```diff
@@ -127,7 +132,9 @@ async def as_service(
         logger.info("Spawning Service for %s", cls.__name__)
         service = Service(cfg, cls, actor_args, actor_kwargs)
         await service.__initialize__()
-        return ServiceInterface(service, cls)
+        service_interface = ServiceInterface(service, cls)
+        await register_service(service_interface)
```
allenwang28, line 177: can we add a comment here?

```
# Register this service with the provisioner so it can cleanly shut this down
```

src/forge/controller/actor.py
```diff
@@ -185,13 +192,16 @@ async def as_actor(cls: Type[T], *args, **actor_kwargs) -> T:
         """
         logger.info("Spawning single actor %s", cls.__name__)
         actor = await cls.launch(*args, **actor_kwargs)
+        await register_actor(actor)
```
allenwang28, line 237:
```
# Register this actor with the provisioner so it can cleanly shut this down
```

src/forge/controller/provisioner.py
```diff
@@ -302,8 +305,52 @@ async def stop_proc_mesh(self, proc_mesh: ProcMesh):
                 commands.kill(server_name)
             del self._proc_host_map[proc_mesh]
 
+    def register_service(self, service: "ServiceInterface") -> None:
+        """Registers a service allocation for cleanup."""
+        from forge.controller.service import ServiceInterface
```
allenwang28, line 348: let's add a comment saying 
```
# Import ServiceInterface here instead of at top-level to avoid circular import
```

src/forge/controller/provisioner.py
```diff
@@ -302,8 +305,52 @@ async def stop_proc_mesh(self, proc_mesh: ProcMesh):
...
+        if not isinstance(service, ServiceInterface):
+            raise TypeError(
+                f"register_service expected ServiceInterface, got {type(service)}"
+            )
+
+        self._registered_services.append(service)
+
+    def register_actor(self, actor: "ForgeActor") -> None:
+        """Registers a single actor allocation for cleanup."""
+        from monarch._src.actor.actor_mesh import ActorMesh
```
allenwang28: can we do this at the toplevel? And use
```
from monarch.actor import ActorMesh
```

src/forge/controller/service/interface.py
```diff
@@ -200,6 +200,7 @@ async def shutdown(self) -> None:
         """
         Shut down the underlying Service.
         """
+        logger.info(f"Shutting down service {self.actor_def.__name__}")
```
allenwang28: since we're quietly shutting down actors/replicas, in this log here can we mention how many actors we're shutting down?

src/forge/controller/provisioner.py
DNXie: I got 
```
ImportError: cannot import name 'ActorMesh' from 'monarch.actor' (/home/dxie/.fbpkg_conda_envs/forge-e146614/lib/python3.10/site-packages/monarch/actor/__init__.py)
```
Sure I can do it on top-level.

src/forge/controller/service/interface.py
DNXie: Done. Now the log only says 
```
Shutting down 3 service(s) and 4 actor(s)...
```

apps/grpo/main.py
```diff
@@ -516,23 +516,10 @@ async def continuous_training():
         training_task.cancel()
     finally:
         print("Shutting down...")
-
         # give mlogger time to shutdown backends, otherwise they can stay running.
         # TODO (felipemello) find more elegant solution
         await mlogger.shutdown.call_one()
```
allenwang28: @DNXie @felipemello1 maybe we can just move the mlogger shutdown into the global shutdown as well?

DNXie: I moved it into `shutdown()`.

src/forge/controller/provisioner.py
```diff
@@ -429,12 +481,43 @@ async def host_mesh_from_proc(proc_mesh: ProcMesh):
...
+
+
 async def stop_proc_mesh(proc_mesh: ProcMesh):
     provisioner = await _get_provisioner()
     return await provisioner.stop_proc_mesh(proc_mesh=proc_mesh)
 
 
+async def shutdown_metric_logger():
+    """Shutdown the global metric logger and all its backends."""
+    from forge.observability.metric_actors import get_or_create_metric_logger
```
allenwang28, line 481: if this doesn't need to be inline imported, can we do this at the toplevel? This should be a general thing

DNXie: This will cause a circular dependency issue. This is also what Felipe did: https://github.com/meta-pytorch/forge/blob/main/src/forge/controller/provisioner.py#L325-L327

allenwang28: ok sg! I will just assume this to be the case moving forward

PR #363, opened 2025-10-09
Metric Logging updates 5/N - enable streaming

felipemello1:
1) Changes logging mode, so its clearer:

Before:
```
reduce_across_ranks: bool
share_run_id: bool
```

After:
```
logging_mode: Enum[GLOBAL_REDUCE, PER_RANK_REDUCE, PER_RANK_NO_REDUCE]
per_rank_share_run: bool
```

2) Adds class:
```
class LoggingMode(Enum):
    GLOBAL_REDUCE = "global_reduce"
    PER_RANK_REDUCE = "per_rank_reduce"
    PER_RANK_NO_REDUCE = "per_rank_no_reduce"
```

3) Introduces the "PER_RANK_NO_REDUCE" mode, aka streaming. This means we call `backend.log(metric)` as soon as we get it, without any reduction.

Before, MetricLogger.push(metric) would just collect the metric. Now, it also logs.
```
def push(self, metric: Metric) -> None:
      # flush in "PER_RANK_NO_REDUCE" mode
      for backend in self.per_rank_no_reduce_backends:
            backend.log_stream(metric=metric, global_step=self.global_step)

      # Always accumulate for reduction and state return
        key = metric.key
        if key not in self.accumulators:
            self.accumulators[key] = metric.reduction.accumulator_class(
                metric.reduction
            )
        self.accumulators[key].append(metric.value)
```

Notice how x-axis is timestamp:
image

4) Main design change: logger backends now have `async def log_batch` and `def log_stream`. It not totally clear to me if both should be async/sync or if i should try to unify them.

```
class LoggerBackend(ABC):
    """Abstract logger_backend for metric logging, e.g. wandb, jsonl, etc."""

    def __init__(self, logger_backend_config: dict[str, Any]) -> None:
        self.logger_backend_config = logger_backend_config

    @abstractmethod
    async def init(
        self,
        role: BackendRole,
        primary_logger_metadata: dict[str, Any] | None = None,
        process_name: str | None = None,
    ) -> None:
        """Initializes backend, e.g. wandb.run.init()."""
        pass

    @abstractmethod
    async def log_batch(
        self, metrics: list[Metric], global_step: int, *args, **kwargs
    ) -> None:
        """Log batch of accumulated metrics to backend"""
        pass

    def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
        """Stream single metric to backend immediately."""
        pass

    async def finish(self) -> None:
        pass

    def get_metadata_for_secondary_ranks(self) -> dict[str, Any] | None:
        """Return sharable state after primary init (e.g., for shared modes). Called only on globals."""
        return None
```

### DIFF
--- a/src/forge/observability/metric_actors.py
+++ b/src/forge/observability/metric_actors.py
@@ -217,11 +217,14 @@ class GlobalLoggingActor(ForgeActor):
     ) -> dict[str, Any]:
         """Validate and normalize backend configuration."""
         if "logging_mode" not in config:
-            logger.debug(
-                f"logging_mode not provided for backend {backend_name}. Defaulting to global_reduce."
+            raise ValueError(
+                f"logging_mode is required for backend '{backend_name}' but was not provided. "
+                f"Please specify a logging_mode in your config. "
+                f"See forge.observability.metrics.LoggingMode for available options: "
+                f"{', '.join([mode.value for mode in LoggingMode])}."
             )
 
-        mode_str = config.get("logging_mode", "global_reduce")
+        mode_str = config["logging_mode"]
         mode = LoggingMode(mode_str)
 
         # Validate per_rank_share_run configuration


apps/grpo/qwen3_1_7b.yaml
```diff
@@ -16,11 +16,12 @@ rollout_threads: 1   # Recommended to set equal to policy.num_replicas
...
-    project: "grpo-training"
-    group: "grpo_exp_${oc.env:USER}"
-    reduce_across_ranks: True
+    project: grpo-training
+    group: grpo_exp_${oc.env:USER}
+    logging_mode: global_reduce # global_reduce, per_rank_reduce, per_rank_no_reduce
+    per_rank_share_run: False
   console:
-    reduce_across_ranks: True
+    logging_mode: global_reduce
```
ebsmothers, line 23: Why do we need to duplicate logging_mode across different configs like this? feels like clunky UX to me

src/forge/observability/metrics.py
```diff
@@ -352,22 +378,23 @@ def reset(self) -> None:
 class MetricCollector:
     """Per-rank singleton for accumulating, retrieving and flushing metrics to backends.
 
-    A logger is represented by a backend, i.e. wandb backend. If reduce_across_ranks=False,
-    the backend is instantiated per-rank, in the MetricCollector, otherwise it is instantiated once globally,
-    in the GlobalLoggingActor.
+    Supports multiple logging backends, each with different logging modes.
```
ebsmothers, line 403: (I know this is the design from prior to this PR but) just wondering why we use a singleton across multiple logger backends. In my mind it's more idiomatic to instantiate different classes for different logger backends, and also that usage pattern shouldn't be super common anyways. (Personally I would prefer e.g. a simple console logger + generalized logging backend for wandb etc as two separate entities.) Basically I worry that we are putting extra onus on people who just wanna log to wandb with all these nested dict configs etc for something that is a bit of a niche use case

src/forge/observability/metrics.py
```diff
@@ -614,24 +693,27 @@ async def log(self, metrics: list[Metric], global_step: int) -> None:
...
-    Track all processes to a single run: reduce_across_ranks=False, share_run_id=True
+    More details on wandb distributed logging here: https://docs.wandb.ai/guides/track/log/distributed-training/
 
     Configuration:
-        reduce_across_ranks (bool, default True): If True, log reduced metrics only from controller (global mode).
-            If False, enables per-rank logging; then use share_run_id to pick mode.
-        share_run_id (bool, default False): Only used if reduce_across_ranks=False.
-            True -> shared run across ranks; False -> separate runs per rank.
+        logging_mode (LoggingMode): Determines logging behavior
+        per_rank_share_run (bool, default False): For per-rank modes, whether to share run ID across ranks.
```
ebsmothers, line 748: I know we chatted about it a bit already, but personally I still don't understand this one. It seems to me like we should always use a single run; if e.g. there are name collisions of metrics within the run we should sort that out ourselves.

src/forge/observability/metrics.py
```diff
@@ -642,39 +724,35 @@ def __init__(self, logger_backend_config: dict[str, Any]) -> None:
...
-                    f"Skipped init for global mode (reduce_across_ranks=True) and {role} role."
-                )
+                logger.warning(f"Skipped init for GLOBAL_REDUCE mode and {role} role.")
                 return
             await self._init_global()
 
-        # Per-rank modes based on share_run_id bool
-        elif role == BackendRole.GLOBAL and self.share_run_id:
+        # Per-rank modes based on per_rank_share_run bool
+        elif role == BackendRole.GLOBAL and self.per_rank_share_run:
```
ebsmothers, line 784: A more general question about extensibility: say I want to use my own logger not provided by forge. Do I now have to implement my own custom logic to handle these different cases? (Basically I want to ensure we are not introducing unnecessary friction for users who want to customize beyond what we've directly provided)

src/forge/observability/metrics.py
```diff
@@ -721,22 +799,38 @@ async def _init_shared_local(self, primary_metadata: dict[str, Any]):
...
-        else:
+    async def log_batch(
+        self, metrics: list[Metric], global_step: int, *args, **kwargs
+    ) -> None:
+        if not self.run:
             logger.debug(f"WandbBackend: No run started, skipping log for {self.name}")
+            return
+
+        # Convert metrics to WandB log format
+        log_data = {"step": global_step}
```
ebsmothers: Why did we change global_step -> step here?

src/forge/observability/metrics.py
```diff
@@ -573,21 +640,31 @@ async def init(
...
+        """Log batch of accumulated metrics to backend"""
+        pass
 
-        Args:
-            metrics: list of Metric objects to log.
-            global_step: Step number for x-axis alignment across metrics.
+    def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
+        """Stream single metric to backend immediately.
+
+        NOTE: This method is called synchronously.
```
ebsmothers, line 686: What's the rationale here? In my mind the difference between this and log_batch behavior is a bit unintuitive

src/forge/observability/metrics.py
```diff
@@ -614,24 +693,27 @@ async def log(self, metrics: list[Metric], global_step: int) -> None:
...
     async def finish(self) -> None:
         pass
 
 
 class WandbBackend(LoggerBackend):
     """
-    Weights & Biases logging backend for distributed training.
+    Weights & Biases logging backend.
+
+    For logging mode details, see `forge.observability.metrics.LoggingMode` documentation.
```
ebsmothers, line 742: I think it would also be helpful to provide some images in the documentation demonstrating the resulting figures from each of the different modes.

src/forge/observability/metrics.py
```diff
@@ -418,40 +452,61 @@ async def init_backends(
...
+                raise TypeError(
+                    f"Expected LoggingMode enum for {backend_name}.logging_mode, got {type(mode)}: {mode}."
+                )
 
-            # get metadata from primary backend if any
-            primary_metadata = {}
-            if metadata_per_primary_backend:
-                primary_metadata = metadata_per_primary_backend.get(backend_name, {})
+            # Skip local instantiation. Backend will be instantiated in GlobalLoggingActor.
+            if mode == LoggingMode.GLOBAL_REDUCE:
```
ebsmothers, line 493: Kinda similar to my other comment about supporting multiple logging backends.. I feel like the MetricCollector class is doing a lot (imo too much). I know I wasn't around to give detailed review of the first few PRs, but I do wonder whether all these different logging modes should really all be handled in the same class. Currently the logic is a bit hard for me to understand

apps/grpo/qwen3_1_7b.yaml
felipemello1: this is per backend. You could have scuba logging on streamining mode, console logging global_reduce and wandb logging per rank. If you have a single backend, you define it only once.

src/forge/observability/metrics.py
felipemello1: "It seems to me like we should always use a single run", then you cannot have this:

image

Imagine that multiple run == multiple json files. e.g. rank_0.json, rank_1.json, etc.

---

hopefully the readme can clarify this: https://github.com/meta-pytorch/forge/pull/380/files

src/forge/observability/metrics.py
felipemello1: If you have your own backend, you have to implement a `LoggerBackend`. The logging_mode defines where things get called. 

If "GLOBAL_REDUCE", we call `log_batch(reduced_metrics)` once, from the controller
if "PER_RANK_REDUCE", we call `log_batch(per_rank_reduced_metrics)` once per rank.
if "PER_RANK_NO_REDUCE", we call log_stream(metric) on every rank as soon as the `record_metric` is called.

In the init the user can define how to initialize it, e.g. create a file, initialize a run, etc

```
class LoggerBackend(ABC):
    @abstractmethod
    async def init(...) -> None:
        pass

    @abstractmethod
    async def log_batch(
        self, metrics: list[Metric], global_step: int, *args, **kwargs
    ) -> None:
        pass
	
    @abstractmethod
    def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
        pass
```

src/forge/observability/metrics.py
felipemello1: we do `record_metric(key, value, reduce)` -> `MetricCollector.push`

if "per_rank_no_reduce", we immediately call `MetricCollector.push` -> `backend.log_stream(single_metric)`
else, we just accumulate the metric until flush.

`GlobalLoggingActor.flush` -> `MetricCollector.flush `-> `backend.log_batch(reduced_metrics)`

I am not 100% convinced that they should be different methods, but i also was not comfortable merging them.

src/forge/observability/metrics.py
felipemello1: documentation as in readme, or in this docstring?

src/forge/observability/metrics.py
felipemello1: >  In my mind it's more idiomatic to instantiate different classes for different logger backends

We do it. But the logic for `backend in backends: backend.log(metric)` has to live somewhere. MetricCollector is this wrapper with 4 methods:

```
class MetricCollector:
       def init_backends():
       def push()
       def flush()
       def shutdown()
```
each one of them is sort of just doing `for backend in backends: backend.do_something()`

src/forge/observability/metrics.py
felipemello1: >  feel like the MetricCollector class is doing a lot (imo too much) 

MetricCollector is this:

```
class MetricCollector:
       def init_backends():
       def push()
       def flush()
       def shutdown()
```

Are you saying that the methods are doing too much or we have too many methods?

> Currently the logic is a bit hard for me to understand

Thats fair. Maybe i can try to make it more obvious, or maybe its a documentation issue for a global view, which i address in https://github.com/meta-pytorch/forge/pull/380/files

apps/grpo/qwen3_1_7b.yaml
ebsmothers: Discussed offline. I would like to consider some refactors here, but acknowledge that it's not really in the scope of this PR

src/forge/observability/metric_actors.py
```diff
@@ -208,57 +210,112 @@ def __init__(self):
...
         """
-        self.config = config
+        self.config = {}
 
+        # Validate and normalize each backend config
         if FORGE_DISABLE_METRICS.get_value():
             return
 
         for backend_name, backend_config in config.items():
+            self.config[backend_name] = self._validate_backend_config(
```
ebsmothers, line 296: Small thing, but i don't love `_validate_backend_config` as is. We are doing some sanity checks and potentially updating the config, but this requires understanding logic in one more function. Imo better to just validate and fail (rather than modify) if the arguments are no good, then the config does not get updated in a separate function

src/forge/observability/metrics.py
ebsmothers: Discussed a bit offline, thanks for addressing in #380 as well

PR #364, opened 2025-10-09
Change 32b config

allenwang28:
A reasonable starting point for 32B for perf testing:
- seq_len = 2048
- group_size = 16
- batch_size = 32
- trainer tp=8
- 4 policy replicas

Also updates `local_batch_size` for the MAST configs

### DIFF
--- a/apps/mast/qwen3_14b_mast.yaml
+++ b/apps/mast/qwen3_14b_mast.yaml
@@ -3,7 +3,7 @@
 
 # Global configuration
 group_size: 8
-batch_size: 16
+local_batch_size: 16 # per-device batch size
 max_req_tokens: 512
 max_res_tokens: 512
 model: "Qwen/Qwen3-14B"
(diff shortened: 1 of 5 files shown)


apps/grpo/qwen3_32b.yaml
```diff
@@ -3,18 +3,18 @@
...
+max_res_tokens: 1024
 model: "Qwen/Qwen3-32B"
 off_by_n: 1 # Off by one by default
 
 provisioner:
   launcher: slurm
 
 # Main loop configuration
-rollout_threads: 1   # Recommended to set equal to policy.num_replicas
+rollout_threads: 8
```
casteryh: I feel we could double this. I made rollout_threads to 4 with 1 policy replica. And the waiting for buffer time went from 300s to 60s.

casteryh:
```suggestion
rollout_threads: 32 # make this equal to the batch size seems to work well
```

allenwang28: good call, as another data point I tried 8x # of replicas and it crashed the workload.

Issue #143, opened 2025-09-10
Introduce multi-worker integration tests between RLEngine and PolicyEngine.


pradeepfn:
improve test_policy_update.py to cover multi-gpu/multi-worker tests.

PR #365, opened 2025-10-09
Fix policy update test

JenniferWang:
* Fix the test fixture yaml files
* Re-write the logic to make the round-trip test compatible with multi-nodes. 

Specifically,
* Move validation function to the `policy.py` module because `@endpoint` almost only work with lambda functions. 
* Refactor the test to make setup / tear down clearer

Closes sub-task in  https://github.com/meta-pytorch/forge/issues/411, close https://github.com/meta-pytorch/forge/issues/143

### DIFF
--- a/tests/integration_tests/test_policy_update.py
+++ b/tests/integration_tests/test_policy_update.py
@@ -264,7 +264,7 @@ class TestWeightSync:
             for _, e in errs.items():
                 assert not e, f"Validation failed with exception: {e}"
 
-        await policy.update_weights.fanout(policy_version=v1)
+        await policy.update_weights.fanout(version=v1)
         all_errs = await policy._test_validate_model_params.fanout(
             _test_validate_params_all_zeros
         )
@@ -273,7 +273,7 @@ class TestWeightSync:
                 assert not e, f"Validation failed with exception: {e}"
 
         # Reloading v0, getting back original weights
-        await policy.update_weights.fanout(policy_version=v0)
+        await policy.update_weights.fanout(version=v0)
         all_errs = await policy._test_validate_model_params.fanout(
             _test_validate_params_unchanged
         )
--- a/src/forge/actors/generator.py
+++ b/src/forge/actors/generator.py
@@ -50,14 +50,10 @@ from forge.actors._torchstore_utils import (
 )
 
 from forge.controller import ForgeActor, get_proc_mesh, stop_proc_mesh
-<<<<<<< HEAD:src/forge/actors/generator.py
 from forge.data_models.completion import Completion
 from forge.data_models.prompt import to_prompt
 from forge.env import TORCHSTORE_USE_RDMA
 from forge.interfaces import Policy as GeneratorInterface
-from forge.data.sharding import VLLMSharding
-from forge.data_models.completion import Completion
-from forge.data_models.prompt import to_prompt
 from forge.observability.metrics import record_metric, Reduce
 from forge.observability.perf_tracker import Tracer
 from forge.types import ProcessConfig
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -18,17 +18,6 @@ import torch
 import torch.distributed.checkpoint as dcp
 import torchstore as ts
 
-from forge.actors._torchstore_utils import (
-    DcpHandle,
-    get_dcp_whole_state_dict_key,
-    get_param_key,
-)
-
-from forge.controller import ForgeActor
-from forge.data.utils import batch_to_device
-from forge.observability.metrics import record_metric, Reduce
-from forge.observability.perf_tracker import Tracer
-
 from monarch.actor import current_rank, current_size, endpoint
 from torch import Tensor
 from torch.distributed.checkpoint._nested_dict import flatten_state_dict


Jack-Khuu:
Can we close https://github.com/meta-pytorch/forge/issues/143 once this lands too :)?

tests/integration_tests/test_policy_update.py
```diff
@@ -58,93 +68,106 @@ async def zero_out_model_states(self):
...
 
-        assert isinstance(cfg, DictConfig)
+    await asyncio.gather(
+        policy.shutdown(),
+        ts.shutdown(),
+        RLTrainer.shutdown(rl_trainer),
+    )
+
+    # Cleanup DCP directory
+    path = Path(TEST_DCP_DIR)
```
casteryh, line 221: You can use TemporaryDirectory which handles this automatically

src/forge/actors/policy.py
```diff
@@ -779,3 +768,69 @@ def get_default_sampling_params(vllm_config, overrides=None) -> SamplingParams:
     # We only care about the final output
     params.output_kind = RequestOutputKind.FINAL_ONLY
     return params
+
```
casteryh: Can you explain why this has to be in `policy.py`? Is monarch failing to pickup/serialize the function if it's not in here?
If it has to be here, may be prepend all the functions with `_`.

tests/integration_tests/test_policy_update.py
JenniferWang: I have to change this because NFS does not work for tmp directory

src/forge/actors/policy.py
JenniferWang: Yes, monarch pickled the function with the module path `integration_tests`  but this cannot be resolved in the remote node. Seems that we need to define it in the main modules.

allenwang28:
> While the test is amazing, having it in the policy.py is not really an option IMO.

Oh I completely missed that part. Is it possible to create an inherited Policy that has all of the test functionality and is kept in integration_tests? 

I'm not sure I'm clear what the monarch pickling issue is

joecummings:
@allenwang28 What are our options here? Surely someone else has tried to run a multi node test via Monarch and not have to clutter up an Actor?

Jack-Khuu:
Wild thought (lmk if this errors): For tests specific methods, can we just append them to the class definition in the test? It's the same idea as inheriting, but let's you surgically change the nested PolicyWorker

Within test:
```python
from foo import Bar

def util(self):
  ...

Bar._special_test_util = util

proxy: Bar = Bar()
```

> I'm not sure I'm clear what the monarch pickling issue is

I'd like to see the trace too since at least 2 others have encountered this when using Monarch/Services

JenniferWang:
@joecummings , @allenwang28 , @Jack-Khuu 

The pickle problem is, what I think, caused by how Monarch serializes arguments (in this case, passing a lambda v.s. a function variable to an endpoint on the Policy service. 

```
WARNING  forge.controller.service.replica:replica.py:257 Got failure on replica 0. Error:
A remote actor call has failed.
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 875, in handle
    result = await instrumented()
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 872, in instrumented
    raise e
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 865, in instrumented
  File "/mnt/home/jiyue/forge/src/forge/actors/policy.py", line 451, in _test_validate_model_params
    return await self.policy_worker._test_validate_model_params.call(validate_fn)
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/future.py", line 138, in mark_complete
    func, value = fut.set_result, await coro
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/endpoint.py", line 147, in process
    rank, value = await r._recv()
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 728, in _recv
    return self._process(result)
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 758, in _process
    return rank, super()._process(msg)
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 737, in _process
    raise cast(Exception, payload)
monarch._src.actor.actor_mesh.ActorError: A remote actor call has failed.
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 811, in handle
    args, kwargs = unflatten(message, local_state)
  File "/mnt/data/jiyue/lib/python3.10/site-packages/monarch/_src/actor/pickle.py", line 98, in unflatten
    return up.load()
ModuleNotFoundError: No module named 'tests.integration_tests'
```

allenwang28:
@JenniferWang how do you run this workload?

JenniferWang:
> @JenniferWang how do you run this workload?

```
pytest -s tests/integration_tests/test_policy_update.py::TestWeightSync::test_sanity_check         --config apps/grpo/qwen3_1_7b.yaml
```

JenniferWang:
@allenwang28 , indeed, i think it's because of the way how the script is invoked.

allenwang28:
got it, I wonder if `PYTHONPATH=. pytest ...` works?

JenniferWang:
> got it, I wonder if `PYTHONPATH=. pytest ...` works?

Magic!

casteryh:
> > got it, I wonder if `PYTHONPATH=. pytest ...` works?
> 
> Magic!

Can you add how to run this test to the docstring?

tests/integration_tests/test_policy_update.py
```diff
@@ -35,13 +39,16 @@
 """
 Run tests:
 
-pytest -s tests/integration_tests/test_policy_update.py::TestWeightSync::test_sanity_check \
-    --config tests/integration_tests/artifacts/qwen3_1_7b_tp.yaml --use_dcp=false
+PYTHONPATH=. pytest -s tests/integration_tests/test_policy_update.py::TestWeightSync::test_sanity_check \
```
JenniferWang, line 42: @casteryh  i documented it here

casteryh: 👍

PR #370, opened 2025-10-10
Rename `batch_size` => `local_batch_size`

JenniferWang:
The current naming is confusing. Here's the definition in TorchTitan
https://github.com/pytorch/torchtitan/blob/aa000a3c42e8bb37e51f26eb3e3e024b37ccc479/torchtitan/config/job_config.py#L196
```
    local_batch_size: int = 8
    """Local batch size (i.e., per-device batch size)"""

    global_batch_size: int = -1
    """
    Global batch size (defaults to `training.local_batch_size * data-parallel degree`)
    """
```
Global batch size is the knob for enabling gradient accumulation
https://github.com/pytorch/torchtitan/blob/6a3a9da9564d82a1120c7639ef6236bb4cffa049/README.md?plain=1#L70C72-L70C89

https://github.com/pytorch/torchtitan/blob/6a3a9da9564d82a1120c7639ef6236bb4cffa049/torchtitan/experiments/forge/engine.py#L159

```
        # calculate gradient accumulation steps
        self.gradient_accumulation_steps = global_batch_size // (
            job_config.training.local_batch_size * dp_degree
        )
        assert self.gradient_accumulation_steps > 0
        self.loss_fn = rescale_accumulated_loss(
            self.loss_fn, self.gradient_accumulation_steps
        )
```

### DIFF
--- a/apps/grpo/qwen3_1_7b.yaml
+++ b/apps/grpo/qwen3_1_7b.yaml
@@ -3,7 +3,7 @@
 
 # Global configuration
 group_size: 8
-batch_size: 16
+local_batch_size: 16 # per-device batch size
 max_req_tokens: 512
 max_res_tokens: 512
 model: "Qwen/Qwen3-1.7B"
@@ -56,7 +56,7 @@ trainer:
   lr_scheduler:
     warmup_steps: 1
   training:
-    local_batch_size: ${batch_size}
+    local_batch_size: ${local_batch_size}
     seq_len: 2048
     max_norm: 1.0
     steps: 1000000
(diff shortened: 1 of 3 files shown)


allenwang28:
this is a step in the right direction, but one nit/confusion is that the `local_batch_size` is the *trainer's* local batch size

But `trainer_device_batch_size` sounds verbose and somehow still wrong lol

JenniferWang:
> this is a step in the right direction, but one nit/confusion is that the `local_batch_size` is the _trainer's_ local batch size
> 
> But `trainer_device_batch_size` sounds verbose and somehow still wrong lol

I think for now, it's less confusing to follow the naming from TorchTitan. Let's decide on better naming in the future.

PR #374, opened 2025-10-10
Environment variable related changes

allenwang28:
This PR does a few things:

# Provisioner changes
- This PR originally started to add in `VLLM_HOST_IP`, `world_size` and `rank` as environment variables in `proc_mesh` creation. 
- But then there was a clear need to inherit a few relevant environment variables (like `TORCHSTORE_USE_RDMA`) in the provisioner, and so I added in:

# Environment variable related changes
- Renames env_constants.py to env.py
- Introduces an EnvVar pattern where you can set the name, default value, and description and easily resolve its logical value within code. Reduces boilerplate we had for checking strings etc.
- Applies the changes to the relevant spots in the codebase

### DIFF
--- a/tests/unit_tests/test_env_constants.py
+++ b/tests/unit_tests/test_env_constants.py
@@ -109,12 +109,6 @@ class TestAllEnvVars:
         assert "FORGE_DISABLE_METRICS" in env_var_names
         assert "MONARCH_STDERR_LOG" in env_var_names
 
-    def test_all_env_vars_is_cached(self):
-        """Test that all_env_vars uses caching."""
-        first_call = all_env_vars()
-        second_call = all_env_vars()
-        assert first_call is second_call
-
     def test_all_env_vars_can_iterate_and_get_values(self):
         """Test that all_env_vars can be used to iterate and get values."""
         for env_var in all_env_vars():


src/forge/env.py
```diff
@@ -0,0 +1,117 @@
...
+"""Centralized constants for environment variable names used in the project."""
+
+import functools
+import os
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass
+class EnvVar:
```
joecummings, line 15: Surprising to me that an abstraction like this doesn't exist in the Python world.

src/forge/env.py
```diff
@@ -0,0 +1,117 @@
...
+)
+
+MONARCH_MAX_FRAME_LENGTH = EnvVar(
+    name="HYPERACTOR_CODE_MAX_FRAME_LENGTH",
+    default=1073741824,
+    description="Sets the maximum frame length for Monarch's actor message delivery in bytes.",
+)
+
+
+@functools.cache
```
joecummings: What's the reasoning for caching this?

src/forge/observability/perf_tracker.py
```diff
@@ -132,9 +126,20 @@ def start(self) -> None:
             raise ValueError("Tracer has already been started")
 
         # Start timing (always enabled)
-        time_with_gpu_events = (
-            os.getenv(METRIC_TIMER_USES_GPU, str(self.time_with_gpu)).lower() == "true"
-        ) and torch.cuda.is_available()
+
+        # TODO - follow up on if this env var behavior makes sense.
```
joecummings, line 130: I don't exactly follow this - maybe @felipemello1 can weigh in?

felipemello1: I wanted a way to shutdown the cuda timing, in case we were worried that it was causing OOM or blocking GPU. Currently it lets you make: everything cpu, everything gpu, keep as it is (none)

we could reduce it to 'make everything cpu',  'keep as it is (none)'

src/forge/env.py
felipemello1: what happens if we do:

```
all_env_vars()
Some code change an envvar
all_env_vars()
```

Would we get the first cached or the updated one?

src/forge/observability/perf_tracker.py
allenwang28: yeah I understand why it's implemented this way -- I don't think we need to change it now

src/forge/env.py
allenwang28: hmm, so I assumed that we wouldn't change env vars in the run itself. So to avoid having to create this list every time we create a proc mesh, we cache it

So in your example, we would get the first cached. I think to avoid confusion I'll remove the cache for now...

Issue #360, opened 2025-10-09
Problems with Clean Shutdown and Actor Management in training scripts


DNXie:
### 🐛 Describe the bug

We noticed several issues with shutdown in training scripts:

## 1. No Centralized Shutdown for Actors/Services  (Solved https://github.com/meta-pytorch/forge/pull/357)

- **Problem:** Each actor/service (e.g., DatasetActor, policy, RLTrainer, etc.) must be shut down individually.
```python
await asyncio.gather(
            DatasetActor.shutdown(dataloader),
            policy.shutdown(),
            RLTrainer.shutdown(trainer),
            ReplayBuffer.shutdown(replay_buffer),
            ComputeAdvantages.shutdown(compute_advantages),
            ref_model.shutdown(),
            reward_actor.shutdown(),
        )
        # TODO - add a global shutdown that implicitly shuts down all services
        # and remote allocations
        await shutdown()
```
- **Context:** There is no global shutdown mechanism in the provisioner to handle all services at once. This leads to repetitive and error-prone shutdown code.
- **Suggestion:** Implement a centralized shutdown in the provisioner to manage all actors/services.

---

## 2. Ctrl+C Does Not Exit Cleanly (Gone itself)

- **Problem:** `grpo/main.py` does not exit on a single Ctrl+C as expected; it requires multiple interrupts.
- **Context:** The main async loop is wrapped in a try/except for `KeyboardInterrupt`, but tasks are not cancelled or awaited cleanly, leading to delayed shutdown.

---

## 3. Excessive Monarch Noise on Interrupt/Shutdown (Fixed with https://github.com/meta-pytorch/torchforge/pull/451)

- **Problem:** Interrupting or finishing a run produces noisy Monarch logs with errors about failed stop messages and closed channels.
```bash
[-]E1009 07:20:30.112712 1322189 hyperactor/src/channel/net.rs:796] session unix:@FzPIPHWcrW42EQC1edKBSZcr.7152260238076237936: failed while receiving ack: Connection reset by peer (os error 104)
[...]

PR #376, opened 2025-10-10
Training termination for grpo/main with finite step limit

DNXie:
Previously the GRPO loop ran indefinitely with both training and rollout tasks active.
This PR adds a `cfg.training.steps` limit so training stops after the specified number of steps, then cleanly terminates rollout tasks.

Now logs end with messages like (I tested with `cfg.training.steps=2`):

```
Reached training limit (2 steps). Exiting continuous_training loop.
Shutting down...
WandbBackend global_controller: Finished run
Health loop stopped gracefully.
Health loop stopped gracefully.
Health loop stopped gracefully.
... 
Shutting down provisioner..
```


apps/grpo/main.py
```diff
@@ -433,8 +433,9 @@ async def continuous_rollouts():
     async def continuous_training():
         training_step = 0
         restart_tracer = True  # Flag to control when to restart tracer
+        max_steps = cfg.trainer.training.get("steps", None)
```
joecummings: Can you put this at the top of the main() loop? And also make it required please. The default can still be null / None, but this way it's very visible.

apps/grpo/main.py
```diff
@@ -471,6 +472,10 @@ async def continuous_training():
                 # Flush metrics every training step to WandB
                 await mlogger.flush.call_one(training_step)
 
+        print(
```
felipemello1, line 503: can we get the logger from `src/forge/util/logging.py` instead?

joecummings: We don't really use the logger elsewhere in the main script do we?

apps/grpo/main.py
```diff
@@ -482,14 +487,17 @@ async def continuous_training():
...
     except KeyboardInterrupt:
         print("Training interrupted by user")
+    finally:
+        print("Shutting down...")
+
         for rollout_task in rollout_tasks:
             rollout_task.cancel()
+        # graceful await all tasks, ignore cancellation noise
+        await asyncio.gather(*rollout_tasks, return_exceptions=True)
         training_task.cancel()
```
felipemello1, line 537: can you explain this change a bit? before we would do gather in the try, and they would be impacted by `KeyboardInterrupt`. Now the gather happens after `KeyboardInterrupt`. Would the user possibly have to run 'KeyboardInterrupt' twice?

casteryh: > can you explain this change a bit? before we would do gather in the try, and they would be impacted by `KeyboardInterrupt`. Now the gather happens after `KeyboardInterrupt`. Would the user possibly have to run 'KeyboardInterrupt' twice?

I think after the first KeyboardInterrupt, all the tasks are canceled and now we just gather on the canceled tasks (which should be fast to resolve) for graceful shutdown.

DNXie: The previous `gather(*rollout_tasks, training_task)` call blocked on rollouts indefinitely, even after `training_task` completed (e.g., once `max_steps` was reached).
Since we now have a finite step limit that cleanly terminates `training_task`, we shouldn’t continue waiting on rollout tasks.

With this change, rollout termination is handled explicitly in the `finally` block.
This doesn’t theoretically change how `KeyboardInterrupt` is handled, since an interrupt caught in the `except` would still flow into `finally`.
However, we’ve already seen issues with `KeyboardInterrupt` handling not working properly (see https://github.com/meta-pytorch/forge/issues/360 (2)); I’ll look into that separately in a follow-up PR.

apps/grpo/main.py
DNXie: Done!

apps/grpo/main.py
```diff
@@ -349,6 +349,11 @@ async def main(cfg: DictConfig):
         ),
     )
 
+    if "steps" not in cfg.trainer.training:
+        raise ValueError("`cfg.trainer.training.steps` must be defined (can be null).")
```
felipemello1: i dont think we need it. If `steps` is not in cfg.trainer, and we try to do `max_steps = cfg.trainer.training.steps`, it should error already

src/forge/controller/service/replica.py
```diff
@@ -266,6 +266,11 @@ async def _process_single_request(self, request: ServiceRequest) -> bool:
                 # can be healthy but the request failed.
                 self.mark_failed()
                 success = False
+            except asyncio.InvalidStateError:
+                # Future was already cancelled — safe to ignore during shutdown
+                self.mark_failed()
+                success = False
+                pass
```
DNXie: Added this guard because during shutdown, some client futures are already cancelled while the replica is still trying to set future. Without catching `InvalidStateError`, Monarch logs a “Task exception was never retrieved” trace when calling `set_result` / `set_exception` on an already-done future.

I suppose we didn’t hit this before because our training loops ran indefinitely and were usually stopped manually with Ctrl-C. (Please correct me if I am wrong)

apps/grpo/main.py
DNXie: Yes you are right. It is used in RLTrainer already.  Removed

apps/grpo/main.py
DNXie: The entire main script is using print instead of logger right now.

src/forge/controller/service/replica.py
```diff
@@ -265,6 +266,11 @@ async def _process_single_request(self, request: ServiceRequest) -> bool:
...
```
casteryh: why do you need `pass` here?

apps/grpo/main.py
```diff
@@ -434,7 +436,7 @@ async def continuous_training():
         training_step = 0
         restart_tracer = True  # Flag to control when to restart tracer
 
-        while True:
+        while max_steps is None or training_step < max_steps:
```
casteryh: I think we should make `cfg.training.steps' a required argument. And explicitly setting to -1 means run until interrupted.

DNXie: Made max_steps 
```
max_steps = cfg.trainer.training.steps or -1
```
and the condition 
```
while max_steps < 0 or training_step < max_steps:
```

src/forge/controller/service/replica.py
DNXie: Right.. Removed

apps/grpo/main.py
```diff
@@ -434,7 +437,7 @@ async def continuous_training():
         training_step = 0
         restart_tracer = True  # Flag to control when to restart tracer
 
-        while True:
+        while max_steps < 0 or training_step < max_steps:
```
Jack-Khuu: Nit: `max_steps == -1` is less error prone than allowing any negative

apps/grpo/main.py
```diff
@@ -482,14 +489,18 @@ async def continuous_training():
...
     except KeyboardInterrupt:
         print("Training interrupted by user")
+    finally:
+        print("Shutting down...")
         for rollout_task in rollout_tasks:
             rollout_task.cancel()
+        # graceful await all tasks, ignore cancellation noise
+        await asyncio.gather(*rollout_tasks, return_exceptions=True)
+        # Give replicas time to drain and complete in-flight requests
+        await asyncio.sleep(1)
```
allenwang28: I think we can/should make this part more graceful. Proposal:

Continuous rollouts takes a shutdown event:
```
async def continuous_rollouts(shutdown_evnt: asyncio.Event):
    ...
    while not shutdown_event.is_set(): # no more while True
        ...

    print("Rollout loop got shutdown event, shutting down...")
```

then in our `finally` we can do a 2-phased shutdown:

```
finally:
    print("Shutting down...")
    shutdown_event.set()

    try:
        # give tasks a chance to exit gracefully
        await asyncio.wait_for(
            asyncio.gather(*rollout_tasks, return_exceptions=True),
            timeout=5
        )
    except asyncio.TimeoutError:
        print("Forcing cancellation...")
        for t in rollout_tasks:
            t.cancel()
        await asyncio.gather(*rollout_tasks, return_exceptions=True)
```

apps/grpo/main.py
DNXie: Done.

apps/grpo/main.py
DNXie: Done

PR #377, opened 2025-10-10
[Cleanup] [Policy] Use vLLM ``SamplingParams`` instead of custom class

joecummings:
### Context

We previously had a custom class called ``SamplingConfig`` that would selectively map variables to vLLM's ``SamplingParams``. While this was nice for conversion to/fro dictionary structs, it was an additional redirect for the user to understand what knobs were actually available for sampling.

### Changes

* Deleted ``SamplingConfig``, instead relying directly on ``SamplingParams``.
* Added a debug log of the resolved ``SamplingParams``
* Updated naming in all configs, tests, etc
* Added a link to the vLLM sampling params in every config
* Deleted guided decoding work b/c that parameter was actually deprecated and unused outside of Judges. I think this will be formally figured out in the Judge work.

### Testing

* WandB Link for Qwen runs showing that it still works the same way: https://wandb.ai/jcummings/grpo-training/runs/8jzcn46h?nw=nwuserjcummings
* Passes the updated unit tests

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -18,17 +18,6 @@ import torch
 import torch.distributed.checkpoint as dcp
 import torchstore as ts
 
-from forge.actors._torchstore_utils import (
-    DcpHandle,
-    get_dcp_whole_state_dict_key,
-    get_param_key,
-)
-
-from forge.controller import ForgeActor
-from forge.data.utils import batch_to_device
-from forge.observability.metrics import record_metric, Reduce
-from forge.observability.perf_tracker import Tracer
-
 from monarch.actor import current_rank, current_size, endpoint
 from torch import Tensor
 from torch.distributed.checkpoint._nested_dict import flatten_state_dict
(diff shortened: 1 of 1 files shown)


Jack-Khuu:
Oh this is a fun one, taking a look 

> Deleted guided decoding work b/c that parameter was actually deprecated and unused outside of Judges. I think this will be formally figured out in the Judge work.

The field is being renamed, but the underlying impl is mostly the same (i think they called it out ~2 weeks ago)

tests/sandbox/vllm/main.py
```diff
@@ -37,8 +37,7 @@ async def run(cfg: DictConfig):
     await mlogger.init_backends.call_one(metric_logging_cfg)
 
     if (prompt := cfg.get("prompt")) is None:
-        gd = cfg.policy.get("sampling_config", {}).get("guided_decoding", False)
-        prompt = "What is 3+5?" if gd else "Tell me a joke"
+        prompt = "Tell me a joke"
```
Jack-Khuu, line 40:
```suggestion
prompt := cfg.get("prompt", "Tell me a joke")
```

joecummings:
> This passes the gut test

Maslow's hierarchy of tests goes unit tests, integration tests, gut tests

tests/sandbox/vllm/main.py
joecummings: Ugh can't apply suggestion on deleted lines :/

src/forge/actors/policy.py
```diff
@@ -200,16 +158,18 @@ async def launch(  # pyright: ignore[reportIncompatibleMethodOverride]
             "vllm_worker", PolicyWorker, vllm_config=vllm_config, use_dcp=use_dcp
         )
 
-        if isinstance(sampling_config, Mapping):
-            sampling_config = SamplingConfig(**sampling_config)
+        if isinstance(sampling_params, Mapping):
+            sampling_params = SamplingParams.from_optional(**sampling_params)
```
Jack-Khuu, line 162: Note for whoever in the future needs to use rich fields of vllm/v1 in yaml:

from_optional doesn't play well with rich fields (nesting)

PR #381, opened 2025-10-10
[Cleanup] [Policy] Use ``EngineArgs`` directly instead of custom class

joecummings:
### Context

We previously had a custom class called ``EngineConfig`` that would selectively map variables to vLLM's ``EngineArgs``. While this was nice for conversion to/fro dictionary structs, it was an additional redirect for the user to understand what knobs were actually available for creating the vLLM engine.

### Changes

* Deleted ``EngineConfig``, instead relying directly on ``EngineArgs``.
* Added a debug log of the resolved ``EngineArgs``
* Updated naming in all configs, tests, etc
* Added a link to the vLLM EngineArgs in every config

### Testing

* WandB Link for Qwen runs showing that it still works the same way: https://wandb.ai/jcummings/grpo-training/runs/ebsg2wln
* Passes the updated unit tests

### DIFF
--- a/src/forge/actors/policy.py
+++ b/src/forge/actors/policy.py
@@ -126,7 +126,7 @@ class Policy(PolicyInterface):
             engine_args._is_v1_supported_oracle = lambda *_: True  # Always default on
             logger.debug(f"Resolved engine args: {engine_args}")
 
-        vllm_config = engine_args.create_vllm_config(UsageContext.LLM_CLASS)
+        vllm_config = engine_args.create_engine_config(UsageContext.LLM_CLASS)
         workers = worker_procs.spawn(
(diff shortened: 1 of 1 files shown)


src/forge/actors/policy.py
```diff
@@ -150,10 +121,12 @@ async def launch(  # pyright: ignore[reportIncompatibleMethodOverride]
         policy_proc_config.with_gpus = False
         policy_proc = await get_proc_mesh(process_config=policy_proc_config)
 
-        if isinstance(engine_config, Mapping):
-            engine_config = EngineConfig.from_dict(engine_config)
+        if isinstance(engine_args, Mapping):
+            engine_args = EngineArgs(**engine_args)
```
Jack-Khuu, line 125: Same flag as the other PR, when folks want to use the richer nested fields in EngineArgs we'll need to support it

PR #386, opened 2025-10-13
Fix test_policy_config after refactor to EngineArgs

Jack-Khuu:
Minor fixes after the refactor to EngineArgs in https://github.com/meta-pytorch/forge/pull/381. Specifically, changing the defaults and removing nested instantiation support

---
Before (Other errors show up as the ones listed get fixed)
image

After
image


Jack-Khuu:
Because it was skipped in CI i think in the other PR, along with other tests (we should fix if it isn't resolved by GPU CI)

PR #396, opened 2025-10-13
Metric Logging updates 6/N - Enable SFT metrics / delete old file

felipemello1:
remove old SFT related metric logging system and replaces with new based on monarch

image

notes: 
a) Epoch brief fluctuation can happen because of how we shard the data. Datasets are infinite, and its possible that while 1 shard was exhausted, another shard from epoch 0 was not.
b) need to follow up to add perf metrics to measure training time and max memory


src/forge/data/metric_transform.py
felipemello1, line 1: moved from forge/data/dataset_metrics

daniellepintz (approved):
LGTM!

ebsmothers (approved):

PR #400, opened 2025-10-14
Flatten GRPO main: Group and Episode 

Jack-Khuu:
The updates boil down to 2 changes that don't alter behavior in `grpo/main`:
* `Group` is downgraded from a dataclass to a typedef of`list[Episode]`, since it's never required
* `Episode` now directly holds a `Completion` with redundant attributes in `Episode` being removed
  * _See https://github.com/meta-pytorch/forge/commit/df0e5a95a019d012ac6b9fde89c82631c7320e8e for how the redundant fields are mapped between `Episode` and `ScoredCompletion`._ 

_(There's also various typehint improvements sprinkled in)_

---

Note: This PR does not address or utilize `Episode` from data_models, but convergence is imminent

```
```

Wandb looks roughly the same
Before: torchforge/grpo-training/runs/wca6wke2
After torchforge/grpo-training/runs/ul34xjr9

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -29,7 +29,6 @@ from forge.controller.actor import ForgeActor
 from forge.controller.provisioner import init_provisioner, shutdown
 from forge.data.rewards import MathReward, ThinkingReward
 from forge.data_models.completion import Completion
-from forge.env import MONARCH_HOSTMESH_V1
 from forge.observability.metric_actors import get_or_create_metric_logger
 from forge.observability.metrics import record_metric, Reduce
 from forge.observability.perf_tracker import Tracer
@@ -306,12 +305,6 @@ async def main(cfg: DictConfig):
     mlogger = await get_or_create_metric_logger()
     await mlogger.init_backends.call_one(metric_logging_cfg)
 
-    # In the host mesh v0 case, actors on remote hosts are not able to communicate
-    # with one another. Therefore we use the controller as our storage volume.
-    if not MONARCH_HOSTMESH_V1.get_value():
-        await ts.initialize(strategy=ts.ControllerStorageVolumes())
-        print("Torchstore successfully initialized with controller storage strategy")
-
     # ---- Setup services ---- #
 
     (
@@ -343,21 +336,19 @@ async def main(cfg: DictConfig):
 
     print("All services initialized successfully!")
     shutdown_event = asyncio.Event()
-    # In the HostMesh v1 case, we spawn a torchstore storage volume
-    # per trainer process.
+    # Here we spawn a torchstore storage volume per trainer process.
     # We initialize after service initialization because torchstore currently
     # requires access to the underlying proc meshes in the local rank strategy.
     # We should be able to hide this in the future.
-    if MONARCH_HOSTMESH_V1.get_value():
-        # TODO: support multiple host meshes
-        trainer_num_procs = cfg.actors.trainer["procs"]
-        trainer_host_mesh_name = cfg.actors.trainer["mesh_name"]
-        trainer_hosts = provisioner.get_host_mesh(trainer_host_mesh_name)
-        await ts.initialize(
-            mesh=trainer_hosts.spawn_procs(per_host={"procs": trainer_num_procs}),
-            strategy=ts.LocalRankStrategy(),
-        )
-        print("Torchstore successfully initialized with local rank strategy")
+    # TODO: support multiple host meshes
+    trainer_num_procs = cfg.actors.trainer["procs"]
+    trainer_host_mesh_name = cfg.actors.trainer["mesh_name"]
+    trainer_hosts = provisioner.get_host_mesh(trainer_host_mesh_name)
+    await ts.initialize(
+        mesh=trainer_hosts.spawn_procs(per_host={"procs": trainer_num_procs}),
+        strategy=ts.LocalRankStrategy(),
+    )
+    print("Torchstore successfully initialized with local rank strategy")
 
     # ---- Core RL loops ---- #
     async def continuous_rollouts():
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -309,6 +309,9 @@ export PATH="${CUDA_HOME}/bin:$PATH"
 export CUDA_INCLUDE_DIRS=$CUDA_HOME/include
 export CUDA_CUDART_LIBRARY=$CUDA_HOME/lib64/libcudart.so
 
+# Temporary measure until this environment variable is removed
+export MONARCH_HOST_MESH_V1_REMOVE_ME_BEFORE_RELEASE=1
+
 # Add only CUDA compat libs to LD_LIBRARY_PATH (safe for system tools)
 if [ -n "${LD_LIBRARY_PATH:-}" ]; then
   export LD_LIBRARY_PATH="/usr/local/cuda-${CUDA_VERSION}/compat:${LD_LIBRARY_PATH}"
--- a/src/forge/controller/launcher.py
+++ b/src/forge/controller/launcher.py
@@ -26,8 +26,6 @@ from monarch.tools.commands import info
 from monarch.tools.components import hyperactor
 from monarch.tools.config import Config, Workspace
 
-from forge.env import MONARCH_HOSTMESH_V1
-
 from forge.types import Launcher, LauncherConfig
 
 _MAST_AVAILABLE = False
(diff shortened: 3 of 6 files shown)


apps/grpo/main.py
```diff
@@ -42,73 +43,43 @@
...
+    completion: Completion | None = None
     ref_logprobs: torch.Tensor | None = None
     reward: float | None = None
     advantage: float | None = None
 
     @property
-    def request_tensor(self):
-        tensor = torch.tensor(self.request_tokens, dtype=torch.long)
+    def policy_version(self) -> int | None:
+        return getattr(self.completion, "generator_version", None)
```
JenniferWang: Why not `self.completion.generator_version`?

apps/grpo/main.py
```diff
@@ -42,73 +43,43 @@
...
     pad_id: int
     request_len: int
     response_len: int
     target: Any | None = None
-    # processed data
-    response: str | None = None
-    request_tokens: list[int] | None = None
-    response_tokens: list[int] | None = None
+    # Processed data
+    completion: Completion | None = None
```
JenniferWang, line 51: I find `Completion` being an odd collection of fields -- it is after all a low level contract, so I think `GeneratorResponse` is probably clearer.

apps/grpo/main.py
```diff
@@ -42,73 +43,43 @@
...
-                    request_len=request_len,
-                    response_len=response_len,
-                    target=target,
-                )
-            )
-        return cls(str(group_id), episodes)
-
-
-def collate(batches: list[list[Episode]]):
+def collate(
```
JenniferWang, line 83: It would be really helpful to add docstring to this function (in `ReplayBuffer`)

apps/grpo/main.py
```diff
@@ -387,68 +358,61 @@ async def continuous_rollouts():
...
+            for i, episode in enumerate(episodes):
                 episode.ref_logprobs = ref_logprobs[i]
             del ref_logprobs, input_ids
-            t.step("compute_logprobs")
+            t.step("gc_ref_logprobs")
 
             # Calculate advantages and add to replay buffer
-            advantages = await compute_advantages.compute.call_one(group)
-            for episode, advantage in zip(group.episodes, advantages):
+            advantages = await compute_advantages.compute.call_one(episodes)
```
JenniferWang, line 407: Not related to this diff but now since we're scrutinizing the main flow again, I think making `compute_advantages` its own `Actor` is very weird and probably the opposite to an "optimization"

1. We do not expose capability to specify the hostmesh for a specific actor -- ideally, this should be collocated with the generator replica that produces this batch.
2. ComputeAdvantage only needs the rewards; so very likely the entire episodes are serialized.

I wonder, if for now it should be just inlined in the `sample` call; or allocating a proc on the Policy mesh along side the PolicyWorker to handle the computation but chain the calls and return the result together in `sample`

allenwang28: @JenniferWang these are good points. I want to propose an idea (not for you to implement @Jack-Khuu just brainstorming if this makes sense)

```
with policy.session() as s:
    host: HostMesh =  await s.get_host_mesh() # returns the host mesh associated with this replica
    advantages = host.run_task(compute_advantages) # where compute_advantages is a function
```

apps/grpo/main.py
```diff
@@ -42,73 +43,43 @@
...
+        response_tokens: torch.Tensor = self.completion.token_ids
+        tensor = torch.tensor(response_tokens, dtype=torch.long)
         if tensor.shape[0] < self.response_len:  # right pad
             diff = self.response_len - tensor.shape[0]
             tensor = F.pad(tensor, (0, diff), value=self.pad_id)
         return tensor
 
 
-@dataclass
-class Group:
```
allenwang28, line 78: sorry Jack, I actually like the group 😅  since this is a fundamental unit within GRPO I think we should keep it

apps/grpo/main.py
joecummings: The completion name matches the OpenAI name for things returned from their API server: https://docs.vllm.ai/en/v0.4.1/getting_started/examples/openai_completion_client.html. 

I believe vLLM is also planning on aligning on this kind of format. 

I don't have a strong preference on the debate here, just sharing some extra information.

apps/grpo/main.py
```diff
@@ -387,68 +358,61 @@ async def continuous_rollouts():
         while not shutdown_event.is_set():
             t = Tracer("main_perf/continuous_rollouts")
             t.start()
-            sample = await dataloader.sample.call_one()
+            sample: dict[str, str] = await dataloader.sample.call_one()
```
joecummings: I don't think we need to overdo it on the type hints in our apps. They're not public APIs.

Open to thoughts here though.

apps/grpo/main.py
```diff
@@ -387,68 +358,61 @@ async def continuous_rollouts():
...
                 return
 
             t.step("data_loading")
 
             prompt, target = sample["request"], sample["target"]
-            responses = await policy.generate.route(prompt)
-            # TODO: this shall be part of the responses metadata instead of a separate call
-            version = await policy.get_version.route()
-
+            responses: list[Completion] = await policy.generate.route(prompt)
```
joecummings, line 368: Same comment about type hints here.

apps/grpo/main.py
```diff
@@ -387,68 +358,61 @@ async def continuous_rollouts():
...
             t.step("policy_generation")
 
             assert (
```
joecummings: Can we remove this assertion?

apps/grpo/main.py
```diff
@@ -387,68 +358,61 @@ async def continuous_rollouts():
...
-                request=prompt,
-                policy_version=version,
-                pad_id=pad_id,
-                request_len=max_req_tokens,
-                response_len=max_res_tokens,
-                target=target,
-            )
 
+            # Construct episodes and calculate rewards
+            episodes: list[Episode] = []
```
joecummings: It's implied already that episodes would be a list of episodes, no?

apps/grpo/main.py
```diff
@@ -387,68 +358,61 @@ async def continuous_rollouts():
...
             input_ids = torch.ones(
                 (group_size, max_req_tokens + max_res_tokens),
                 dtype=torch.long,
                 device="cuda",
```
joecummings: This probably doesn't need to be on cuda anymore. The ref model will move it to the appropriate device.

apps/grpo/main.py
```diff
@@ -387,68 +358,61 @@ async def continuous_rollouts():
...
                 input_ids, max_req_tokens, return_logprobs=True
             )
             t.step("reference_model_calculate_logprobs")
 
-            for i, episode in enumerate(group.episodes):
+            for i, episode in enumerate(episodes):
                 episode.ref_logprobs = ref_logprobs[i]
             del ref_logprobs, input_ids
-            t.step("compute_logprobs")
+            t.step("gc_ref_logprobs")
```
joecummings: Let's keep timing minimal. I HIGHLY doubt anyone will really care about how long gc the ref_logprobs and input_ids takes. And if they want that info, they can add it back.

apps/grpo/main.py
```diff
@@ -387,68 +358,61 @@ async def continuous_rollouts():
...
+                episodes.append(episode)
+
+                # Build input_ids for reference logprobs
+                input_ids[i, :max_req_tokens] = episode.request_tensor
+                input_ids[i, max_req_tokens:] = episode.response_tensor
 
             t.step("reward_evaluation")
 
-            ref_logprobs = await ref_model.forward.route(
+            ref_logprobs: torch.Tensor = await ref_model.forward.route(
```
joecummings: type hint comment

apps/grpo/main.py
joecummings: Chained calls would be cool :eyes:

apps/grpo/main.py
joecummings: What if we just called the variable name that holds all these episodes a "group" without a dataclass? Still in line with the GRPO notation, but lighter weight.

apps/grpo/main.py
Jack-Khuu: Completion is a naming convention that the industry has adopted for generation (I don't like it either)

apps/grpo/main.py
Jack-Khuu: `Group = List[Episode]`

Compromise?

apps/grpo/main.py
Jack-Khuu: No preference, just added it cause i was there

apps/grpo/main.py
Jack-Khuu: This one I want, we've been bitten by the changing return types in the past

apps/grpo/main.py
Jack-Khuu: This looks legit; +1 on chained calls

joecummings (approved):

PR #406, opened 2025-10-14
Remove HostMesh v0 APIs

allenwang28:
Tested this locally and on SLURM. The monarch hostmesh piece will be automatically set in the install script for now so users don't have to do this.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -309,9 +309,6 @@ export PATH="${CUDA_HOME}/bin:$PATH"
 export CUDA_INCLUDE_DIRS=$CUDA_HOME/include
 export CUDA_CUDART_LIBRARY=$CUDA_HOME/lib64/libcudart.so
 
-# Temporary measure until this environment variable is removed
-export MONARCH_HOST_MESH_V1_REMOVE_ME_BEFORE_RELEASE=1
-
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -309,6 +309,9 @@ export PATH="${CUDA_HOME}/bin:$PATH"
 export CUDA_INCLUDE_DIRS=$CUDA_HOME/include
 export CUDA_CUDART_LIBRARY=$CUDA_HOME/lib64/libcudart.so
 
+# Temporary measure until this environment variable is removed
+export MONARCH_HOST_MESH_V1_REMOVE_ME_BEFORE_RELEASE=1
+
```
vidhyav: Is it possible to stage this in two PRs?

First, move to v1 as default and then remove all the v0 code as another PR? Can help us easily revert if needed in the future.

src/forge/env.py
```diff
@@ -99,12 +99,6 @@ def get_value(self) -> Any:
     description="Sets the maximum frame length for Monarch's actor message delivery in bytes.",
 )
 
-MONARCH_HOSTMESH_V1 = EnvVar(
-    name="MONARCH_HOST_MESH_V1_REMOVE_ME_BEFORE_RELEASE",
```
vidhyav, line 103: (y)

PR #412, opened 2025-10-14
Add CI badges to README.md

JenniferWang:
image

### DIFF
--- a/.github/workflows/gpu_test.yaml
+++ b/.github/workflows/gpu_test.yaml
@@ -1,4 +1,4 @@
-name: GPU tests
+name: GPU Tests
 
 on:
   schedule:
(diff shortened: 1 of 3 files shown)


joecummings:
Lol we sure about this?

JenniferWang:
> Lol we sure about this?

What's the other common way for people to track the test passing status on Github? I am indifferent so long as there's a way for us to track

joecummings:
> > Lol we sure about this?
> 
> What's the other common way for people to track the test passing status on Github? I am indifferent so long as there's a way for us to track

This is definitely the right way to do things. I'm just concerned that lots of stuff will be broken to start and we'll have a constantly red badge :/

casteryh:
> > > Lol we sure about this?
> > 
> > 
> > What's the other common way for people to track the test passing status on Github? I am indifferent so long as there's a way for us to track
> 
> This is definitely the right way to do things. I'm just concerned that lots of stuff will be broken to start and we'll have a constantly red badge :/

If that's the case, maybe we should have constant red badge then.

Issue #413, opened 2025-10-14
[BUG] RefModel incorrectly reuses trainer checkpoint due to shared default `./checkpoint` folder


DNXie:
### 🐛 Describe the bug

Currently, in grpo/main, when both the `trainer` and reference model (`ref_model`) are configured with checkpoints but without explicitly specifying `folder`, both default to `./checkpoint`.

Example configuration:

```yaml
trainer:
  checkpoint:
    enable: true
    initial_load_path: hf://${model}
    initial_load_in_hf: true
    last_save_in_hf: true
    interval: 500
    async_mode: "disabled"

ref_model:
  checkpoint:
    enable: true
    initial_load_path: hf://${model}
    initial_load_in_hf: true
```

Neither of trainer or ref_model specifies the `folder`. [The default value is `./checkpoint`](https://github.com/pytorch/torchtitan/blob/7c1048075bd7b47edbed6db50986c8e7e7492b0a/torchtitan/config/job_config.py#L449). So the above config is equivalent to:

```yaml
trainer.checkpoint.folder: ./checkpoint
ref_model.checkpoint.folder: ./checkpoint
```

If the trainer has saved checkpoints under `./checkpoint`, Titan’s loading logic will [**ignore `initial_load_path`**](https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/checkpoint.py#L600-L604) and instead automatically [pick up the latest local checkpoint](https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/checkpoint.py#L610) (e.g., `./checkpoint/step-x`, where `x` is largest number it can find ).

As a result, the reference model unintentionally loads the **trainer’s checkpoint weights** instead of its own intended HF initialization.

### **Proposed Fix**

Hardcode the reference model’s checkpoint folder to an empty string (`folder=""`) in [`reference_model.py`](https://github.com/meta-pytorch/forge/blob/9782f57c8919ae93bfdff073c81e2da4b27743bf/src/forge/actors/reference_model.py#L40) to ensure it bypasses Titan’s folder existence check and always loads from `initial_load_path`.

### **Alternative Plan**
Alternatively, enforce a validation rule such that:

* Instead of hardcoding, make the behavior explicit and user-configurable.
```yaml
ref_model:
  checkpoint:
    enable: true
    folder: ""            # Empty folder = force load from initial_load_path
    initial_load_path: hf://${model}
    initial_load_in_hf: true
[...]

PR #414, opened 2025-10-14
hardcode refmodel.checkpoint.folder to empty

DNXie:
Solving https://github.com/meta-pytorch/forge/issues/413

After this, the checkpoint for reference model will be load only from  [`initial_load_path`](https://github.com/meta-pytorch/forge/blob/main/apps/grpo/qwen3_1_7b.yaml#L112). User still have the flexibility to change ref model by modifying `initial_load_path`.

Tested on grpo/main. `ref_model` now will always load from the `initial_load_path` regardless of trainer's checkpoint config.

### DIFF
--- a/src/forge/actors/reference_model.py
+++ b/src/forge/actors/reference_model.py
@@ -87,7 +87,11 @@ class ReferenceModel(ForgeActor):
     @endpoint
     async def setup(self):
         engine_config = {f.name: getattr(self, f.name) for f in fields(self)}
-        self.engine = ForgeEngine(ForgeJobConfig(**engine_config))
+        engine_config = ForgeJobConfig(**engine_config)
+        engine_config.checkpoint.folder = (
+            ""  # hardcode to empty to force load from initial_load_path
+        )
+        self.engine = ForgeEngine(engine_config)
         self.engine.checkpointer.load()
         self.model = self.engine.model_parts[0]  # No pipeline parallelism yet
         self.model.eval()


src/forge/actors/reference_model.py
```diff
@@ -115,7 +87,11 @@ def __post_init__(self):
     @endpoint
     async def setup(self):
         engine_config = {f.name: getattr(self, f.name) for f in fields(self)}
-        self.engine = ForgeEngine(ForgeJobConfig(**engine_config))
+        engine_config = ForgeJobConfig(**engine_config)
+        engine_config.checkpoint.folder = (
+            ""  # hardcode to empty to force load from initial_load_path
```
JenniferWang, line 92: Where does ref model get the initial weights after the change?

DNXie: from [`initial_load_path`](https://github.com/meta-pytorch/forge/blob/main/apps/grpo/qwen3_1_7b.yaml#L112)

JenniferWang: I see. So maybe add a validator on the ref model's config and only allow fields that differs per model. for example, Also put `enable: true` and ` initial_load_in_hf: true` in the override together with the `folder`.

JenniferWang (approved):

src/forge/actors/reference_model.py
pbontrager: Can you add a todo to make this configurable in the future? Sometimes you do want to update your reference model just very infrequently.

pbontrager: @JenniferWang that would have to be on the titan side right? Also, I think this module will be reusable for any "forward pass" model in the near future and not just for reference model. So I don't know if it's worth doing anything more than this now.

allenwang28: can this not just be accomplished by setting this in the config?

```
  checkpoint:
    enable: true
    initial_load_path: hf://${model}
    folder: "" 
```

DNXie: @allenwang28 Yes, it is the same. But I thought it would be a little confusing to the users.
@pbontrager  If user want to change the reference model, they can just update the `initial_load_path` in the config.  Does it resolve some of your concern?

PR #421, opened 2025-10-15
Rename `_test` method in `generator.py`

JenniferWang:
Trying to make the `Generator` API a bit cleaner.

### DIFF
--- a/src/forge/actors/generator.py
+++ b/src/forge/actors/generator.py
@@ -494,13 +494,11 @@ class Generator(ForgeActor):
     @endpoint
     async def save_model_params(self):
         """Used for debugging purpose. Save model parameters before weight update."""
-        logger.info("[Generator] save model parameters for debugging.")
         await self.worker.save_model_params.call()
 
     @endpoint
     async def validate_model_params(self, validate_fn):
         """Used for debugging purpose. Validate saved params using validate_fn."""
-        logger.info("[Generator] start validating saved model parameters.")
         return await self.worker.validate_model_params.call(validate_fn)
 
 
@@ -514,8 +512,6 @@ class GeneratorWorker(ForgeActor):
     """
 
     vllm_config: VllmConfig
-    # Used for debugging purposes only
-    debug_saved_params = {}
 
     @endpoint
     async def setup(self):
@@ -607,18 +603,17 @@ class GeneratorWorker(ForgeActor):
     @endpoint
     async def save_model_params(self):
         """Used for debugging purposes. Save model parameters before weight update."""
-        logger.info("[GeneratorWorker] save model parameters for debugging.")
+        self._debug_saved_params = {}
         for name, param in self.worker.model_runner.model.named_parameters():
-            self.debug_saved_params[name] = param.detach().cpu()
+            self._debug_saved_params[name] = param.detach().cpu()
         logger.info(
             "[GeneratorWorker] finished saving model parameters, len = %d",
-            len(self.debug_saved_params),
+            len(self._debug_saved_params),
         )
 
     @endpoint
     async def validate_model_params(self, validate_fn):
         """Used for debugging purposes. Validate saved params using validate_fn."""
-        logger.info("[GeneratorWorker] start validating saved model parameters.")
         return validate_fn(
-            self.debug_saved_params, self.worker.model_runner.model, logger
+            self._debug_saved_params, self.worker.model_runner.model, logger
         )


src/forge/actors/generator.py
```diff
@@ -508,16 +508,16 @@ async def shutdown(  # pyright: ignore[reportIncompatibleMethodOverride]
         await stop_proc_mesh(actor._generator_proc)
 
     @endpoint
-    async def _test_save_model_params(self):
-        """Save model parameters before weight update, used for tesing purposes only."""
-        logger.info("[Generator] save model parameters for testing.")
-        await self.generator_worker._test_save_model_params.call()
+    async def save_model_params(self):
+        """Save model parameters before weight update, used for debugging purposes."""
```
Jack-Khuu: Nit: Can we bump the for debugging purpose to the front? It's easy to miss at the end of a line

```suggestion
        """(For Debugging/Testing): Save model parameters before weight update."""
```

src/forge/actors/generator.py
```diff
@@ -483,16 +483,16 @@ async def shutdown(  # pyright: ignore[reportIncompatibleMethodOverride]
         await stop_proc_mesh(actor._generator_proc)
 
     @endpoint
-    async def _test_save_model_params(self):
-        """Save model parameters before weight update, used for tesing purposes only."""
-        logger.info("[Generator] save model parameters for testing.")
-        await self.worker._test_save_model_params.call()
+    async def save_model_params(self):
+        """Used for debugging purpose. Save model parameters before weight update."""
+        logger.info("[Generator] save model parameters for debugging.")
```
joecummings: Can we remove the logging?

src/forge/actors/generator.py
```diff
@@ -483,16 +483,16 @@ async def shutdown(  # pyright: ignore[reportIncompatibleMethodOverride]
...
+        await self.worker.save_model_params.call()
 
     @endpoint
-    async def _test_validate_model_params(self, validate_fn):
-        """Validate updated model params using validate_fn."""
-        logger.info("[Generator] start validating model parameters.")
-        return await self.worker._test_validate_model_params.call(validate_fn)
+    async def validate_model_params(self, validate_fn):
+        """Used for debugging purpose. Validate saved params using validate_fn."""
+        logger.info("[Generator] start validating saved model parameters.")
```
joecummings: Remove logging

src/forge/actors/generator.py
```diff
@@ -596,20 +596,20 @@ async def update_weights(self, version: int) -> None:
...
             "[GeneratorWorker] finished saving model parameters, len = %d",
-            len(self._test_prev_params),
+            len(self.debug_saved_params),
         )
 
     @endpoint
-    async def _test_validate_model_params(self, validate_fn):
-        """Validate updated model params using validate_fn."""
-        logger.info("[GeneratorWorker] start validating model parameters.")
+    async def validate_model_params(self, validate_fn):
```
joecummings, line 615: Actually - could we expose an endpoint to get the underlying model? Then these methods could be entirely deleted, no?

Sorry if I'm misunderstanding. But I don't like having a debug param saved on this model :/

JenniferWang: I think it's a fair tradeoff to expose certain method for debugging purpose.
The beauty of exposing a validator function here is it's a very flexible and reliable way to poke at the property on each worker shard and saves the cost of message SerDer. 
Another restriction here is Policy is behind Service interface....

joecummings: > saves the cost of message SerDer.

Do we imagine testing w/  these methods at a substantial scale?

src/forge/actors/generator.py
```diff
@@ -596,20 +596,20 @@ async def update_weights(self, version: int) -> None:
         t.stop()
 
     @endpoint
-    async def _test_save_model_params(self):
-        """Save model parameters before weight update, used for tesing purposes only."""
-        logger.info("[GeneratorWorker] save model parameters for testing.")
+    async def save_model_params(self):
```
joecummings, line 604: Could this method return all the model params? Instead of saving them to the model?

src/forge/actors/generator.py
JenniferWang: > > saves the cost of message SerDer.
> 
> Do we imagine testing w/ these methods at a substantial scale?

My hunch is so long as we're still iterating on weight sync strategy, having some capability to poke into the states across multiple workers under multi-node, multi-replica run might be super valuable.

src/forge/actors/generator.py
casteryh: Sorry I am late to the party. The short answer is no: the workers are remote and it's not feasible to send all the parameters through RPC.

src/forge/actors/generator.py
casteryh: Why remove logging? These methods are only used in tests and won't spam the logs otherwise.

PR #422, opened 2025-10-15
CI badge should point to main branch

casteryh:


.github/workflows/unit_test.yaml
```diff
@@ -2,7 +2,13 @@ name: Unit Tests
 
 on:
   pull_request:
-
+  schedule:
```
joecummings: Why would we run this every hour?

Should only run on PR and push to main

casteryh: done

PR #430, opened 2025-10-15
shared memory multiprocess prefetch for weight update

casteryh:
## What this PR does
- Add a SharedTensor class that's backed by shared memory. It exposes a handle that can be trivially serialized and shared across processes inside a host. This is necessary since pytorch's native tensor backed by shared memory is serialized by value in monarch/cloudpickle.
- Spawn dedicated procs colocated with the workers that fetch weights from torchstore to shared memory which will be forwarded to GeneratorWorker. Note naively parallelize weight fetch by doing asyncio.gather() will cause monarch RDMA operations to fail, having multiple procs circumvents this.
- Let the dedicated procs prefetch while generator is still generating.

## Perf
__TL;DR e2e weight sync time is now ~50s for QWen3 32b; one training step takes <70s__

| | before | after |
|--|-------|------|
| wait for existing generations | 5 - 15s    | 5 - 15s |
| update_weights (total)  | ~80s | ~50s |
| update_weights (since generation completes) | ~80s | 30s - 45s
| fetch_weights | n/a | ~45s |
| shared_memory -> vllm worker | n/a | ~5s |

## Tested with
- unit tests for SharedTensor
- integration test with qwen3_32b [paste](https://www.internalfb.com/phabricator/paste/view/P1994309965)
- dcp code path still works (only tested with 8b because 32b is waaaay too slow) [paste](https://www.internalfb.com/phabricator/paste/view/P1994319261)
- apps/grpo/main
  - [run 1](https://meta.wandb.io/torchforge/grpo-training/runs/asvf3twb) dead after 60 steps, disk quota exceeded when writing wandb logs...
  - [run 2](https://meta.wandb.io/torchforge/grpo-training/runs/d3f63st6) crashed after 40+ steps, illegal CUDA memory access (might be a vLLM bug https://github.com/vllm-project/vllm/issues/8177)
  - [run 3](https://meta.wandb.io/torchforge/grpo-training/runs/ap87qb0j) (installed flashinfer) crashed/link broken after 60+ steps, no obvious reasons from log
  - [control run](https://meta.wandb.io/torchforge/grpo-training/runs/1psxse94)  at 9afb76965224873144c2785275241595b69e4b9f (installed flashinfer) crashed/link broken after 40+steps, no obvious reasons from log

### DIFF
--- a/tests/unit_tests/util/test_shared_tensor.py
+++ b/tests/unit_tests/util/test_shared_tensor.py
@@ -408,9 +408,8 @@ class TestSharedTensorMultiprocess:
         """Test reading shared tensor from another process"""
 
         def reader_process(handle_dict, result_queue):
-            shared = SharedTensor(handle=handle_dict)
-            tensor = shared.tensor
-            result_queue.put(tensor.sum().item())
+            with SharedTensor(handle=handle_dict) as shared:
+                result_queue.put(shared.tensor.sum().item())
 
         # Create shared tensor in main process
         shared = SharedTensor.empty((100, 100), torch.float32)
@@ -435,8 +434,8 @@ class TestSharedTensorMultiprocess:
         """Test writing to shared tensor from another process"""
 
         def writer_process(handle_dict, value):
-            shared = SharedTensor(handle=handle_dict)
-            shared.tensor.fill_(value)
+            with SharedTensor(handle=handle_dict) as shared:
+                shared.tensor.fill_(value)
 
         # Create empty shared tensor
         shared = SharedTensor.empty((50, 50), torch.float32)
@@ -458,11 +457,10 @@ class TestSharedTensorMultiprocess:
         """Test bidirectional communication"""
 
         def worker_process(input_handle, output_handle):
-            input_tensor = SharedTensor(handle=input_handle).tensor
-            output_tensor = SharedTensor(handle=output_handle).tensor
-
-            # Compute: output = input * 2
-            output_tensor.copy_(input_tensor * 2)
+            with SharedTensor(handle=input_handle) as input_shared:
+                with SharedTensor(handle=output_handle) as output_shared:
+                    # Compute: output = input * 2
+                    output_shared.tensor.copy_(input_shared.tensor * 2)
 
         # Create input and output tensors
         input_shared = SharedTensor.empty((100, 100), torch.float32)
@@ -605,8 +603,8 @@ class TestSharedTensorHandleToSharedTensor:
         """Test to_shared_tensor in multiprocess scenario"""
 
         def worker_process(handle, result_queue):
-            shared = handle.to_shared_tensor()
-            result_queue.put(shared.tensor.sum().item())
+            with handle.to_shared_tensor() as shared:
+                result_queue.put(shared.tensor.sum().item())
 
         original = SharedTensor.empty((50, 50), torch.float32)
         original.tensor.fill_(3.0)
@@ -831,10 +829,9 @@ class TestSharedTensorCloseAndCleanup:
         """Test that multiple receivers can close independently"""
 
         def receiver_process(handle, value, result_queue):
-            shared = SharedTensor(handle=handle)
-            result = shared.tensor[0, 0].item() == value
-            shared.close()  # Each receiver closes its own reference
-            result_queue.put(result)
+            with SharedTensor(handle=handle) as shared:
+                result = shared.tensor[0, 0].item() == value
+                result_queue.put(result)
 
         # Creator
         shared = SharedTensor.empty((10, 10), torch.float32)


src/forge/actors/generator.py
```diff
@@ -226,11 +231,53 @@ async def setup(self):
             log_stats=None,
         )
         self._start_processing()
+        fetcher_procs = this_host().spawn_procs(
```
allenwang28, line 242: can we guard this with `if prefetch_weights`? It may also be a bit cleaner to put this init in another function w/ some documentation for what the fetchers here do

src/forge/actors/generator.py
```diff
@@ -226,11 +231,53 @@ async def setup(self):
...
+            per_host={"procs": self.n_fetcher_procs}
```
allenwang28, line 243: not that we need to address now, but I think we need to spawn these fetcher procs across all generator nodes right?

casteryh: yes. but I assume `setup` call is broadcasted and every Generator node will spawn their own `fetcher_procs`?

allenwang28: that is correct, I meant if a generator workers span 2 nodes i.e. DeepSeek

In that case we would probably want to spin up the fetchers on the worker nodes right?

casteryh: Correct

src/forge/actors/generator.py
casteryh: will move to a separate function.

src/forge/actors/generator.py
JenniferWang: I am a bit confused -- shouldn't vLLM worker be scoped to a single node?

JenniferWang:
@casteryh could you split the PR with separate concerns?
1. Using shared memory for weight sync when CPU is involved (TorchStore RDMA CPU-CPU, DCP). This can be turned on by default for now.
2. Prefetching the weights while completing the on-the-fly generation requests. 

Also maybe add comment somewhere saying the following is up-for-discussion
1. TorchStore RDMA GPU-GPU.
2. Multi-node vLLM 

I also don't quite get the "before" and "after" table.

casteryh:
> @casteryh could you split the PR with separate concerns?

I think it makes little sense to split, see below.

> 
> 1. Using shared memory for weight sync when CPU is involved (TorchStore RDMA CPU-CPU, DCP). This can be turned on by default for now.

Currently TorchStore RDMA only works with CPU-CPU.
I can enable this for DCP in a separate PR.

> 2. Prefetching the weights while completing the on-the-fly generation requests.

This actually comes automatically once you have separate processes fetching the weights to shared memory.
I can add a flag that waits until all generation completes and then fetch the weights, but why would someone want to do that in the first place?

> Also maybe add comment somewhere saying the following is up-for-discussion
> 
> 1. TorchStore RDMA GPU-GPU.
> 2. Multi-node vLLM

Will do

> I also don't quite get the "before" and "after" table.

Ah maybe it's confusing because I am trying to do two things at once. Basically the speed up comes from 1. multiprocess shared memory (this saves 30 seconds) & 2. prefetch while completing on-the-fly generation (this saves about 10 seconds on average)

JenniferWang:
> This actually comes automatically once you have separate processes fetching the weights to shared memory.
I can add a flag that waits until all generation completes and then fetch the weights, but why would someone want to do that in the first place?

Ahha, yes, I got it now. Okay, so the boolean guard is not "use prefetch or not" -- is should be "use shared memory or not". I think you should try profiling on proc = 8 / replica for policy.

casteryh:
> Ahha, yes, I got it now. Okay, so the boolean guard is not "use prefetch or not" -- is should be "use shared memory or not". I think you should try profiling on proc = 8 / replica for policy.

@JenniferWang 
for tp=8 on policy
almost the same except loading from shared memory to gpu is now faster (because each worker has a smaller shard)
```
generator_perf/_fetch_weights/total_duration_avg_s: 43.30411313060904
generator_perf/_fetch_weights/total_duration_max_s: 44.24197581084445
generator_perf/waiting_for_fetch_weights/total_duration_avg_s: 28.329474024591036
generator_perf/waiting_for_fetch_weights/total_duration_max_s: 40.779690923169255
generator_worker_perf/update_weights_from_shared_memory/total_duration_avg_s: 3.1711842537115444
generator_worker_perf/update_weights_from_shared_memory/total_duration_max_s: 3.380359285045415
```

casteryh:
@allenwang28  @JenniferWang  ptal
do you want me to enable this by default?

JenniferWang:
> > Ahha, yes, I got it now. Okay, so the boolean guard is not "use prefetch or not" -- is should be "use shared memory or not". I think you should try profiling on proc = 8 / replica for policy.
> 
> @JenniferWang for tp=8 on policy almost the same except loading from shared memory to gpu is now faster (because each worker has a smaller shard)
> 
> ```
> generator_perf/_fetch_weights/total_duration_avg_s: 43.30411313060904
> generator_perf/_fetch_weights/total_duration_max_s: 44.24197581084445
> generator_perf/waiting_for_fetch_weights/total_duration_avg_s: 28.329474024591036
> generator_perf/waiting_for_fetch_weights/total_duration_max_s: 40.779690923169255
> generator_worker_perf/update_weights_from_shared_memory/total_duration_avg_s: 3.1711842537115444
> generator_worker_perf/update_weights_from_shared_memory/total_duration_max_s: 3.380359285045415
> ```

Yes, I was thinking that tp = 8 on policy would be worse without shared memory ?

JenniferWang:
I think we should make using shared memory by default for CPU based weight sync, with a flag to turn it off.

casteryh:
> > > Ahha, yes, I got it now. Okay, so the boolean guard is not "use prefetch or not" -- is should be "use shared memory or not". I think you should try profiling on proc = 8 / replica for policy.
> > 
> > 
> > @JenniferWang for tp=8 on policy almost the same except loading from shared memory to gpu is now faster (because each worker has a smaller shard)
> > ```
> > generator_perf/_fetch_weights/total_duration_avg_s: 43.30411313060904
> > generator_perf/_fetch_weights/total_duration_max_s: 44.24197581084445
> > generator_perf/waiting_for_fetch_weights/total_duration_avg_s: 28.329474024591036
> > generator_perf/waiting_for_fetch_weights/total_duration_max_s: 40.779690923169255
> > generator_worker_perf/update_weights_from_shared_memory/total_duration_avg_s: 3.1711842537115444
> > generator_worker_perf/update_weights_from_shared_memory/total_duration_max_s: 3.380359285045415
> > ```
> 
> Yes, I was thinking that tp = 8 on policy would be worse without shared memory ?

yes I believe it was ~100 seconds without shared memory for tp=8, but I have some problem with my slurm node and can't test now.

casteryh:
> I think we should make using shared memory by default for CPU based weight sync, with a flag to turn it off.

done

casteryh:
fixed a memory leak

src/forge/actors/generator.py
```diff
@@ -432,7 +440,7 @@ async def update_weights(self, version: int) -> None:
             >>> generator.update_weights(version)
         """
         # Prefetch only if we are using RDMA
```
JenniferWang:
```suggestion
        # TODO: enable shared memory prefetch for DCP-based weight sync
```

src/forge/util/_shared_tensor.py
```diff
@@ -32,11 +31,102 @@ def to_shared_tensor(self) -> SharedTensor:
...
+        Note: This only unlinks, it does not close any handles. Processes that have
+        opened this shared memory should call close() on their SharedTensor instances.
+        """
+        try:
+            # Attach to the shared memory just to unlink it
+            shm = shared_memory.SharedMemory(name=self.shm_name)
+            shm.close()
+            shm.unlink()
+        except Exception:
+            pass
```
JenniferWang, line 54: What's the consideration behind swallowing the exceptions in cleaning up the resource?

JenniferWang (approved):

src/forge/util/_shared_tensor.py
casteryh: To make this idempotent and safe to be called from multiple-processes. Open to other ideas.

src/forge/actors/generator.py
```diff
@@ -617,3 +717,26 @@ async def validate_model_params(self, validate_fn):
         return validate_fn(
             self._debug_saved_params, self.worker.model_runner.model, logger
         )
+
+
+class _WeightFetcher(ForgeActor):
```
pbontrager, line 723: I think this could be a method on the generator that gets called from main, so prefetch is controlled and visible from the main loop. I am curious if this has to actually be a separate process since this is an async method and I would think most of the time it's waiting on ts.get.

src/forge/actors/generator.py
```diff
@@ -617,3 +717,26 @@ async def validate_model_params(self, validate_fn):
...
+        version: int,
+        param_names: list[str],
+    ) -> dict[str, SharedTensorHandle]:
+        """Fetch weights from torchstore and load them into shared memory."""
+        sd = {}
+        for name in param_names:
+            param_key = get_param_key(version, name)
+            param = await ts.get(param_key)
+            # Use context manager to ensure cleanup after getting handle
+            with SharedTensor(tensor=param) as shared_tensor:
```
pbontrager, line 740: Is the plan to move this to TS and hide the rdma/shared memory logic from the user?

src/forge/actors/generator.py
pbontrager: I also don't follow why you need more than 1. Is it to allow you to parallelize torchstore requests?

src/forge/actors/generator.py
```diff
@@ -92,6 +95,8 @@ class Generator(ForgeActor):
     engine_args: EngineArgs | Mapping = field(default_factory=EngineArgs)
     sampling_params: SamplingParams | Mapping = field(default_factory=SamplingParams)
     use_dcp_for_weight_sync: bool | None = None
+    prefetch_weights_to_shm: bool = True
```
pbontrager, line 98: In general we should try to avoid changing the "public" api when we expect to quickly change the backend again. After launch we should try to keep this in mind.

pbontrager (commented):
This is awesome! I think we'll want to go over how we're doing prefetch again after some of this is upstreamed to torchstore, but otherwise it looks great. I wonder if this is too risky of a change to make before PTC thought?

src/forge/actors/generator.py
casteryh: This has to be a separate actor because it has to be launched in a separate process

src/forge/actors/generator.py
casteryh: agreed.

casteryh:
> I also don't follow why you need more than 1. Is it to allow you to parallelize torchstore requests?

Yes it's 2x faster than 1 process. I haven't tuned this parameter too much though. @pbontrager

src/forge/actors/generator.py
casteryh: Hopefully yes.

casteryh:
> This is awesome! I think we'll want to go over how we're doing prefetch again after some of this is upstreamed to torchstore, but otherwise it looks great. I wonder if this is too risky of a change to make before PTC thought?

I am testing its stability right now. But fwiw, the current main is not stable / well tested either.

casteryh:
> This is awesome! I think we'll want to go over how we're doing prefetch again after some of this is upstreamed to torchstore, but otherwise it looks great. I wonder if this is too risky of a change to make before PTC thought?

we can also switch the flag to be False by default

casteryh:
landing this as the apparent leak is in torchstore and not caused by this PR.

with fix in torchstore:
image
https://meta.wandb.io/torchforge/grpo-training/runs/huufaif1

PR #436, opened 2025-10-16
Update Tutorial Docs: PolicyActor -> Policy Actor

Jack-Khuu:
2 Minor Tutorial Edits:
- Changed `PolicyActor` to `Policy Actor`: `PolicyActor` isn't a true class; `Actor` is the subclass of `Policy` 
i.e. `Policy(ForgeActor)`
- Updated import statement to alias `Generator` to `Policy`
`from forge.actors.generator import Generator as Policy`
  - Usage of "Policy" as variables/RL concept is intentionally unchanged

### DIFF
--- a/docs/source/tutorial_sources/zero-to-forge/1_RL_and_Forge_Fundamentals.md
+++ b/docs/source/tutorial_sources/zero-to-forge/1_RL_and_Forge_Fundamentals.md
@@ -88,7 +88,7 @@ graph LR
     subgraph Services["TorchForge Services (Real Classes)"]
         direction TB
         S1["DatasetActor"]
-        S2["Policy"]
+        S2["Generator"]
         S3["RewardActor"]
         S4["ReferenceModel"]
         S5["ReplayBuffer"]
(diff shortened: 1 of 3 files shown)


docs/source/tutorial_sources/zero-to-forge/1_RL_and_Forge_Fundamentals.md
```diff
@@ -290,7 +290,7 @@ TorchForge handles behind the scenes:
 ### Independent Scaling
 ```python
 
-from forge.actors.policy import Policy
+from forge.actors.generator import Generator as Policy
```
joecummings, line 293: ? Why not just keep as Generator everywhere?

Jack-Khuu: Because our docs/tutorials are written with Policy as the logical component and I wanted to avoid a mass change.

If we think it's worth it, we can do another pass pushing the abstraction deeper

PR #441, opened 2025-10-16
Add warning for existing build/ directory in env setup

casteryh:
## Summary
- Added a check for existing `build/` directory before installing the forge package in `.meta/mast/env_setup.sh`
- When detected, the script warns the user that the directory may contain artifacts from a previous pip installation that could interfere with the current installation
- User is prompted to continue or cancel the installation
- If they continue, they are warned that things might go wrong and given instructions on how to manually remove the directory

## Test plan
- Run the env_setup.sh script with and without an existing build/ directory to verify the warning and prompt work correctly
- Verify that the user can cancel the installation if they choose
- Verify that the user can continue with the installation if they choose

### DIFF
--- a/.meta/mast/env_setup.sh
+++ b/.meta/mast/env_setup.sh
@@ -205,8 +205,26 @@ else
     log_warn "Skipping monarch torch import hack (monarch may not be installed yet)"
 fi
 
-# Step 4: Install forge package
-log_info "Step 4: Installing forge package..."
+# Step 4: Check for existing build directory and warn user
+log_info "Step 4: Checking for existing build directory..."
+if [ -d "build" ]; then
+    log_warn "Detected existing build/ directory at: $(pwd)/build"
+    log_warn "This directory may contain artifacts from a previous pip installation"
+    log_warn "that could interfere with the current installation."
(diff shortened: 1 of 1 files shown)


.meta/mast/env_setup.sh
```diff
@@ -205,8 +205,26 @@ else
...
 
-# Step 4: Install forge package
-log_info "Step 4: Installing forge package..."
+# Step 4: Check for existing build directory and warn user
+log_info "Step 4: Checking for existing build directory..."
+if [ -d "build" ]; then
+    log_warn "Detected existing build/ directory at: $(pwd)/build"
+    log_warn "This directory may contain artifacts from a previous pip installation"
+    log_warn "that could interfere with the current installation."
+    log_warn "If you encounter issues, manually remove it with: rm -rf build"
```
allenwang28, line 214: what is the risk of always deleting it all? lol

JenniferWang (approved):

.meta/mast/env_setup.sh
casteryh: idk lol

PR #442, opened 2025-10-16
Remove unused CLI setup related files

Jack-Khuu:
Removing old CLI system. Only file remaining from the directory is `config.py` (which is used for parsing throughout the repo) and that will be moved to `util/` in a separate PR


casteryh (approved):

casteryh:
as long as nothing in `apps/` `tests/sandbox` `tests/integration_tests` breaks.

pbontrager (commented):
I think this is the right call. You should also search the docs and readme though as I think the cli commands are still referenced places.

PR #443, opened 2025-10-16
Remove chat environment

Jack-Khuu:
The guillotine falls

As titled. Removing unused environment folder (which contained just chat)


casteryh:
Why do we want to remove this?

allenwang28:
> Why do we want to remove this?

I don't think we're using this at all in our GRPO main

Not that environments are wrong or anything like that, but I'm not sure this invocation is what we really want. I'd rather us re-design this once we have a better idea of what we want to do here

PR #444, opened 2025-10-16
Make checkpoint saving `folder` clear in the config

DNXie:
Made it clear in the config where to specify the checkpoint saving folder. And added some necessary comments to clarify the behavior.

### DIFF
--- a/apps/grpo/qwen3_1_7b.yaml
+++ b/apps/grpo/qwen3_1_7b.yaml
@@ -74,8 +74,9 @@ trainer:
     disable_loss_parallel: true
   checkpoint:
     enable: true
-    initial_load_path: hf://${model}
-    initial_load_in_hf: true
+    folder: ./checkpoint              # The folder to save checkpoints to.
+    initial_load_path: hf://${model}  # The path to load the initial checkpoint from. Ignored if `folder` exists.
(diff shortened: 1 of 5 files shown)


apps/grpo/qwen3_1_7b.yaml
```diff
@@ -74,8 +74,9 @@ trainer:
     disable_loss_parallel: true
   checkpoint:
     enable: true
-    initial_load_path: hf://${model}
-    initial_load_in_hf: true
+    folder: ./checkpoint              # The folder to save checkpoints to.
```
JenniferWang, line 77: I think we control these config fields and we should be opinionated on exposing RL friendly config field names and re-map it to TorchTitan fields internally.

Right now, some TorchTitan config names are really confusing: e.g. the ref model logically should not need checkpointing but still it requires `checkpoint.enable = true`.

ebsmothers: Yeah was discussing this with @joecummings too.. ultimately we probably want to provide some kind of internal config mapping that we execute as a kind of training script post-init before we do any of the actual setup of actors (e.g. we can even bake it into our config.parse decorator).

ebsmothers (approved):

PR #446, opened 2025-10-17
Remove unused interfaces and types

Jack-Khuu:
As titled. Specifically removed:
- `interfaces.py`: Environment, Policy
- `types.py`: ForgeEnvInfo, Action, Trajectory, State

Also `test_replay_buffer` was updated to use a locally defined `TestEpisode` instead of a `Trajectory` since the test was the only caller of `Trajectory` (and was using it more as a proxy for `Episode` due to the definition not being centralized)

---
`python -m pytest tests/unit_tests/test_replay_buffer.py::TestReplayBuffer`


Jack-Khuu:
Failing on deleted env tests: Will be rebased onto and landed after https://github.com/meta-pytorch/forge/pull/443

PR #448, opened 2025-10-17
Docs Content Part 1: Homepage and getting started

AlannaBurke:
I split the docs work into 2 PRs, this is the first. Let's get this merged ASAP.

PR #449, opened 2025-10-17
Docs Content Part 2: Concepts

AlannaBurke:
I split the docs work into 2 PRs, this is the second. 

It builds on #448, so let's merge that first.


svekars:
Can you fix the warnings: https://github.com/meta-pytorch/forge/actions/runs/18580111434/job/52973145035?pr=449#step:11:109 ?

docs/source/architecture.md
```diff
@@ -0,0 +1,256 @@
...
+
+```python
+# Create a policy service with 16 replicas, each using 8 processes
+policy = PolicyActor.options(
+    procs=8,
+    with_gpus=True,
+    num_replicas=16
+).as_service()
+
+# Create a lightweight coding environment service
```
allenwang28: let's remove this for now

docs/source/architecture.md
```diff
@@ -0,0 +1,256 @@
...
+**Load Balancing**
+: In async RL, multiple `continuous_rollouts()` tasks run concurrently. Services automatically distribute these rollouts across available replicas - no manual worker pool management.
+
+**Fault Tolerance**
+: If a replica fails during a rollout, services detect it, mark it unhealthy, and route subsequent requests to healthy replicas. The failed replica gets restarted automatically. Your RL code never sees the failure.
+
+**Ephemeral Infrastructure**
+: Services are created with your job and torn down when finished. Want to try a new reward model? Change your Python code. No standing deployments to maintain, no infrastructure to provision ahead of time.
+
+## TorchStore: Distributed Weight Storage
```
allenwang28, line 125: this all looks good to me, but cc @LucasLLC  in case of any desired big changes

docs/source/architecture.md
```diff
@@ -0,0 +1,256 @@
...
+
+**Asynchronous Updates**
+: Training and inference can read/write weights independently, enabling true async RL
+
+**Flexible Storage**
+: Store tensors co-located with trainers, on their own storage tier, sharded or replicated - change with minimal code modifications
+
+### Why TorchStore Matters
+
+Without TorchStore, weight synchronization becomes the bottleneck in async RL. Traditional approaches either:
```
allenwang28: let's not say "without TorchStore" because there are other possible solutions! 

We can just simply say

"Weight Synchronization is a bottleneck in async RL. Traditional approaches either: ..."

docs/source/architecture.md
```diff
@@ -0,0 +1,256 @@
...
+- Require synchronous GPU-to-GPU transfers (blocking training)
+- Use slow network filesystems (minutes per update)
+- Demand complex manual resharding logic (error-prone, hard to maintain)
+
+TorchStore solves all of these, keeping data distributed across the cluster until requested and moving it efficiently with RDMA.
+
+## Resource Management
```
allenwang28: I am not sure the placement of this section makes sense

It's all accurate, but Forge doesn't really solve these problems itself or reference them at all. I'd prefer to remove it altogether

docs/source/architecture.md
```diff
@@ -0,0 +1,256 @@
...
+**Batch Size Tuning**
+: Balance memory usage and throughput for optimal training speed
+
+**Dynamic Batching**
+: Group requests efficiently for inference (vLLM does this)
+
+**Kernel Fusion**
+: Combine operations to reduce memory bandwidth (torch.compile helps)
+
+## Distributed Training Strategies
```
allenwang28, line 191: this is true, but I would rather not maintain a separate documentation about the sharding strategies since we're just using what torchtitan is using.

Can we instead point to their repo: https://github.com/pytorch/torchtitan? cc @tianyu-l @wwwjn in case there is a pointer you want us to link

docs/source/technology_stack.md
```diff
@@ -0,0 +1,120 @@
+# Technology Stack
```
allenwang28, line 1: I have this slide I'll use at PTC: https://docs.google.com/presentation/d/1owgUNluoyHoPmhg-cKG9di_SgDVTC7wI/edit?slide=id.g38d9e7ea0be_1_421#slide=id.g38d9e7ea0be_1_421

in case we want to show a visualization here too

docs/source/architecture.md
tianyu-l: oh, I don't think we have a better thing to link than the repo.
Maybe you could consider the forge entry as well https://github.com/pytorch/torchtitan/tree/main/torchtitan/experiments/forge

tianyu-l: btw as part of this effort, maybe you would consider writing better docs for the forge folder in torchtitan

docs/source/concepts.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Replay Buffer
+
+Stores episodes for training. Can be implemented with various strategies:
+- **FIFO**: Simple queue for on-policy algorithms
+- **Prioritized**: Importance sampling for off-policy learning
+- **Reservoir**: Uniform sampling from history
+- **Hybrid**: Mix multiple strategies
+
+Integrates with TorchStore for efficient distributed storage.
```
allenwang28: let's remove this line, this is not true right now unfortunately

docs/source/concepts.md
```diff
@@ -0,0 +1,150 @@
...
+
+Write code as if nothing fails. When failures do occur:
+- Monarch fails fast by default (like uncaught exceptions)
+- Add fine-grained fault handling exactly where you need it
+- Services automatically route around failed replicas
+- Failed actors restart automatically
+
+You choose your fault tolerance granularity based on your needs.
+
+## Best Practices
```
allenwang28: Let's remove the best practices section for now, as we haven't consolidated on them ourselves yet

docs/source/concepts.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Debugging
+
+- Start with single-GPU training to isolate issues
+- Enable verbose logging for distributed runs
+- Use profiling tools to identify bottlenecks
+- Validate data pipelines before full training
+- Monitor loss curves and generation quality
+
+## Validation
```
allenwang28: Let's remove the validation section here as well

allenwang28 (commented):
It mostly looks good, @AlannaBurke  pls ping me once the changes are incorporated and this PR is only focused on the concepts.md!

AlannaBurke:
@allenwang28 addressed all your comments!

PR #429, opened 2025-10-15
Revert "Metric Logging updates 4/N - better actor name (#351)"

allenwang28:
Reverting commit 1f45470b0294c50cee539fd110722173c6714872 until issues/tests are resolved

Example error we saw before:
```
thread '<unnamed>' panicked at fbcode/monarch/hyperactor/src/reference.rs:652:14:
world_name() called on direct proc
[-]E1015 15:33:56.344937  2041 fbcode/monarch/hyperactor/src/panic_handler.rs:47] panic at fbcode/monarch/hyperactor/src/reference.rs:657:23, stacktrace:   0: std::panicking::update_hook::<hyperactor::panic_handler::set_panic_hook::{closure#0}>::{closure#0}
             at ./fbcode/monarch/hyperactor/src/panic_handler.rs:31:25
   1: std::panicking::rust_panic_with_hook
             at ./xplat/rust/toolchain/sysroot/1.90.0/library/alloc/src/boxed.rs:1985:9
   2: std::panicking::begin_panic_handler::{closure#0}
             at ./xplat/rust/toolchain/sysroot/1.90.0/library/std/src/panicking.rs:706:13
   3: std::sys::backtrace::__rust_end_short_backtrace::<std::panicking::begin_panic_handler::{closure#0}, !>
             at ./xplat/rust/toolchain/sysroot/1.90.0/library/std/src/sys/backtrace.rs:174:18
   4: __rustc::rust_begin_unwind
             at ./xplat/rust/toolchain/sysroot/1.90.0/library/std/src/panicking.rs:697:5
   5: core::panicking::panic_fmt
             at ./xplat/rust/toolchain/sysroot/1.90.0/library/core/src/panicking.rs:75:14
   6: core::option::expect_failed
             at ./xplat/rust/toolchain/sysroot/1.90.0/library/core/src/panicking.rs:268:5
   7: <hyperactor::reference::ActorId>::rank
             at ./xplat/rust/toolchain/sysroot/1.90.0/library/core/src/option.rs:964:21
   8: <monarch_hyperactor::proc::PyActorId>::__pymethod_get_rank__
             at ./fbcode/monarch/monarch_hyperactor/src/proc.rs:320:20
   9: <pyo3::pyclass::create_type_object::GetSetDefType>::create_py_get_set_def::getter
             at ./third-party/rust/vendor/pyo3-0.24.2/src/pyclass/create_type_object.rs:653:50
  10: _PyObject_GenericGetAttrWithDict
[...]

PR #351, opened 2025-10-08
Metric Logging updates 4/N - better actor name

felipemello1:
When logging per rank, we need good naming to make it easier to debug. Now provisioner.py::get_proc_mesh provides a mesh_name (https://github.com/meta-pytorch/forge/pull/378), so i am just using it. 

User can also pass a process name as input. Thats how we get the `Controller` name.
```
mlogger = await get_or_create_metric_logger(process_name="Controller")
```

The process_name then goes: `LocalFetcherActor` -> `MetricCollector` -> `backend.init` --> `wandb.init(name)`

image

PR #451, opened 2025-10-17
fix - Metric logging work with new monarch API

felipemello1:
We landed https://github.com/meta-pytorch/forge/pull/351, which adds better rank names to logging (later reverted in https://github.com/meta-pytorch/forge/pull/429). This adds 351 back, but with fixes for monarch.

Goal of the original PR:
image

Monarch changed APIs, breaking the code in two ways:
1) 
**Situation**: To get the `actor_name` and `ProcMesh` uid, i was using `monarch.actor.context()`. 
**Problem**: Now actor_name returns `{actor_name}_{ProcMeshuid}` (or `client` if outside a PrcoMesh), and calling context.world_name errors if called outside of a `ProcMesh` context.
**Fix**: Never call `.world_name` and get the uid from actor_name, if any. 

2) 
**Situation**: To allow `GlobalLoggingActor` to access the `LocalFetcherActor` on each `ProcMesh`, we would register the (key,value) as `{ProcMesh: LocalFetcherActor}`
**Problem**: The ProcMesh is not hashable anymore.
**Solution**: Instead of using ProcMesh as a key, we create a 'str' UUID for the proc.

---

Extra: updated docstrings to make clear whats going on

### DIFF
--- a/tests/unit_tests/observability/test_utils.py
+++ b/tests/unit_tests/observability/test_utils.py
@@ -6,7 +6,6 @@
 
 """Tests for observability utility functions."""
 
-import pytest
 from forge.controller.actor import ForgeActor
 
 from forge.observability.utils import get_proc_name_with_rank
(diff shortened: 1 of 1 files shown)


felipemello1:
can ignore the unit test failing. CI is using old monarch version. Error is harmless too. Tests pass locally.

tests/sandbox/toy_rl/toy_metrics/main.py
```diff
@@ -95,12 +95,16 @@ async def main():
...
 
     # Spawn services first (triggers registrations via provisioner hook)
-    trainer = await TrainActor.options(**service_config).as_service()
-    generator = await GeneratorActor.options(**service_config).as_service()
+    trainer = await TrainActor.options(
+        **service_config, mesh_name="TrainActor"
+    ).as_service()
+    generator = await GeneratorActor.options(
+        **service_config, mesh_name="GeneratorActor"
+    ).as_service()
```
DNXie, line 107: What's the purpose of `mesh_name` here? Is it for the better display on wandb? Why is it not applied to all main files?

felipemello1: Allen added it last week https://github.com/meta-pytorch/TorchForge/blob/8cef7e7da68747eae0b2dc0e0ea5308262b0c144/apps/grpo/qwen3_1_7b.yaml#L131

src/forge/observability/metric_actors.py
```diff
@@ -102,24 +115,32 @@ async def get_or_create_metric_logger(
     # Setup local_fetcher_actor if needed (unless disabled by environment flag)
     if not proc_has_local_fetcher and not FORGE_DISABLE_METRICS.get_value():
         local_fetcher_actor = proc.spawn(
-            "local_fetcher_actor", LocalFetcherActor, global_logger
+            "local_fetcher_actor", LocalFetcherActor, global_logger, process_name
         )
-        await global_logger.register_fetcher.call_one(local_fetcher_actor, proc)
+        # Generate a unique ID to map procmesh to fetcher
+        proc._uid = str(uuid.uuid4())
```
DNXie, line 121: Thanks for the fix! LGTM

felipemello1: do you mind approving it?

DNXie: I am a little concerning about the broken CI. Wouldn't it cause all the subsequent commits to break as well?

felipemello1: I dont think that errors are related to this PR. But let me confirm by opening a dummy PR

src/forge/controller/provisioner.py
```diff
@@ -333,14 +334,14 @@ async def stop_proc_mesh(self, proc_mesh: ProcMesh):
             )
             return
         async with self._lock:
-            # Deregister local logger from global logger
-            if hasattr(proc_mesh, "_local_fetcher"):
+            # Deregister LocalFetcherActor from GlobalLoggingActor
+            if hasattr(proc_mesh, "_local_fetcher") and hasattr(proc_mesh, "_uid"):
```
DNXie, line 338: Is it possible for a proc_mesh that has `_local_fetcher` but not `_uid`?

felipemello1: no, they should always have both. I guess i was having extra safe here. Is it confusing?

DNXie: I understand you write it like this to be safe. But I just worry it may hide some potential errors. How about raise an error if it has `_local_fetcher ` but not `_uid`?

DNXie (approved):

PR #452, opened 2025-10-17
clean up torchtune

DNXie:


src/forge/cli/config.py
```diff
@@ -56,22 +56,22 @@ def _merge_yaml_and_cli_args(yaml_args: Namespace, cli_args: list[str]) -> DictC
...
     Example:
         >>> config.yaml:
         >>>     a: 1
         >>>     b:
-        >>>       _component_: torchtune.models.my_model
+        >>>       _component_: my_module.models.my_model
         >>>       c: 3
 
-        >>> tune full_finetune --config config.yaml b=torchtune.models.other_model b.c=4
+        >>> python main.py --config config.yaml b=my_module.models.other_model b.c=4
```
Jack-Khuu, line 70: Does this actually work or is it just an example?

DNXie: It is really working 😅

Jack-Khuu:
Might wanna wait for a bit haha, looks like we have another rename inbound

PR #461, opened 2025-10-17
Enable RDMA by default

allenwang28:
- Sets the default RDMA_ENABLED to 1 (which torchstore does already)
- Switches use_dcp to use TorchStore's rdma_available

Tested at: https://wandb.ai/cabernet-team/grpo-training/runs/y24k1l2l


casteryh (approved):
Image

casteryh:
<del> lgtm, can you run it can see if it's actually using rdma (e.g. by looking at if dcp checkpoint exists and weight update time). </del>

The run you shared 404'd for me but I trust you lol

PR #464, opened 2025-10-20
Remove selective_log_softmax and unify with compute_logprobs

joecummings:
This PR removes the `selective_log_softmax` function and consolidates all log probability computation through the existing `compute_logprobs` function.

Changes:
- Removed `selective_log_softmax` from src/forge/util/ops.py
- Added `align` parameter to `compute_logprobs` to handle both usage patterns:
  - align=False: for pre-aligned logits (when model was called with input_ids)
  - align=True: for extracting subset from full-sequence logits
- Updated all references to use `compute_logprobs` with appropriate align flag
- Added comprehensive docstring explaining both usage patterns with examples
- Removed test_selective_log_softmax.py (functionality now covered by compute_logprobs)

The consolidation improves code maintainability by having a single function for log probability computation, while the align parameter handles the two common usage patterns in RL training.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -65,7 +65,7 @@ jobs:
 
           # Verify dependencies can be imported before building docs
           echo "Verifying dependencies..."
-          python -c "import forge; print('✓ forge imported successfully')"
+          python -c "import forge; print('✓ torchforge imported successfully')"
           python -c "import monarch; print('✓ monarch imported successfully')"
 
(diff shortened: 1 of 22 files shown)


JenniferWang:
Could you add unit test for `compute_logprobs` instead?

Jack-Khuu:
> Could you add unit test for compute_logprobs instead?

https://github.com/meta-pytorch/torchforge/blob/main/tests/unit_tests/util/test_compute_logprobs.py

PR #466, opened 2025-10-20
Fix for sandboxed Python coder unit test

allenwang28:
Seemed that mocking stopped working after some Monarch updates, so this moves the sandboxed coder into its own class that can be mocked. Should fix the GPU unit tests


allenwang28:
seems like this fixes the coder unit test but the remaining error is something around observability

PR #474, opened 2025-10-21
Consolidate on install.sh, move to stable packages

ebsmothers:
Tested via:

```
conda create -n forge python=3.10 -y && conda activate forge
./scripts/install.sh
pytest tests/unit_tests
...
======== 342 passed, 3 skipped, 1 warning in 174.97s (0:02:54) ==========
```

Also working on green CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ You can also find our notebook tutorials (coming soon)
 
 ### Basic
 
-torchforge requires the PyTorch 2.9.0 with [Monarch](https://github.com/meta-pytorch/monarch), [vLLM](https://github.com/vllm-project/vllm), and [torchtitan](https://github.com/pytorch/torchtitan). (Note that the basic install script
+torchforge requires PyTorch 2.9.0 with [Monarch](https://github.com/meta-pytorch/monarch), [vLLM](https://github.com/vllm-project/vllm), and [torchtitan](https://github.com/pytorch/torchtitan). (Note that the basic install script
 uses [DNF](https://docs.fedoraproject.org/en-US/quick-docs/dnf/), but could be easily extended to other Linux OS.)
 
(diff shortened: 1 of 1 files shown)


assets/versions.sh
```diff
@@ -8,7 +8,7 @@
 # This file contains all pinned versions and commits for dependencies
 
 # PyTorch version
-PYTORCH_VERSION="2.9.0.dev20250905"
+PYTORCH_VERSION="2.9.0"
 
 # vLLM branch
 VLLM_BRANCH="v0.10.0"
```
joecummings: Can we change these to point to monarch stable, torchtitan stable and torchstore stable

scripts/install.sh
```diff
@@ -95,128 +95,6 @@ detect_os_family() {
     esac
 }
 
-# Install required system packages
-install_system_packages() {
```
joecummings, line 99: We don't need this anymore?

assets/versions.sh
ebsmothers: Yeah right now I have them hardcoded. But good point. Honestly I need to see where else these versions are used.. there's a world where we just delete this file entirely to reduce indirection

scripts/install.sh
ebsmothers: Idk maybe we do, just trying it out

ebsmothers: brought it back just for safety

.github/workflows/unit_test.yaml
joecummings, line 1: These won't work b/c u need a GPU runner

ebsmothers: wait why? they were working before, no?

allenwang28 (approved):

PR #480, opened 2025-10-21
prepare for monarch behavior change

casteryh:
Preparing for monarch's behavior change when we get the new version.
Context: D85172238


casteryh:
> Which new version?

This is not even landed yet but see diff attached. We will eventually get this new behavior.

PR #482, opened 2025-10-21
Metric Logging - Enable kwargs so user can add extra args

felipemello1:
Before we would pass a whole dictionary as input to a logger backend and pop a selected few args

```
class LoggerBackend(ABC):
	def __init__(self, logger_backend_config: dict[str, Any]) -> None:
		self.project = logger_backend_config["project"]
```

This means that if users want to pass extra args, e.g. run name or tags, they couldnt.

Now we use `**kwargs` and pass it directly to the wandb run:

```
class LoggerBackend(ABC):
	def __init__(self, *, logging_mode, per_rank_share_run, **kwargs):
           self.backend_kwargs = kwargs

	...

	async def _init_per_rank(self, run_name: str):
        self.run = wandb.init(name=run_name, **self.backend_kwargs)
```

Tested it with multiple args. We can even pass a custom name. e.g.

logging_mode: global_reduce
image

logging_mode: per_rank_reduce
image

### DIFF
--- a/tests/unit_tests/observability/test_metrics.py
+++ b/tests/unit_tests/observability/test_metrics.py
@@ -85,12 +85,12 @@ class TestMetricCreation:
     async def test_backend_role_usage(self):
         """Test that BackendRole constants are actually used instead of string literals."""
         # Test ConsoleBackend
-        console_backend = ConsoleBackend({})
+        console_backend = ConsoleBackend(logging_mode=LoggingMode.GLOBAL_REDUCE)
         await console_backend.init(role=BackendRole.LOCAL)
 
         # Test WandbBackend role validation without WandB initialization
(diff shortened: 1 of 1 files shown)


wukaixingxp:
LTGM, tested this branch with my yaml:
```
metric_logging:
  wandb:
    entity: "torchforge"
    project: "kaiwu-coding-grpo"
    logging_mode: global_reduce
  console:
    logging_mode: global_reduce
```
and saw 
` 
wandb: 🚀 View run royal-feather-3 at: https://meta.wandb.io/torchforge/kaiwu-coding-grpo/runs/8xpli7ro
wandb: Find logs at: wandb/run-20251021_153007-8xpli7ro/logs`

PR #483, opened 2025-10-21
Remove useless comments

joecummings:


apps/grpo/main.py
```diff
@@ -406,13 +396,11 @@ async def continuous_rollouts():
                 episode.ref_logprobs = ref_logprobs[i]
             del ref_logprobs, input_ids
 
-            # Calculate advantages and add to replay buffer
```
felipemello1, line 409: :O i kind like this. It works as a visual way to segment code blocks

PR #486, opened 2025-10-21
Add Sample-level Logging API

DNXie:
This PR introduces structured **sample-level logging** to complement existing scalar metrics, allowing users to inspect concrete prompt–response–reward examples during RL training.

More discussion: https://github.com/meta-pytorch/forge/issues/301

The current implementation logs 2 samples: one top (highest reward) and one bottom (lowest reward) for each step. Supporting customized sampling strategy is out of scope of this PR. For now, this can be achieved by changing  the filter for `SampleAccumulator`. For example:
```python
class SampleAccumulator(MetricAccumulator):
    """Accumulator for sample-level metrics (e.g., prompt/response/reward dicts).

    Optionally uses a sample filter to decide what to keep at append/flush time.
    """

    def __init__(
        self, reduction: Reduce, filter: TopBottomKFilter | None = TopBottomKFilter(top=2, bottom=2)  # <--- here
    ):
```
---

#### **Summary of Changes**

* Added **`SampleAccumulator`** to support logging structured dict samples (e.g., per-episode data) via  `record_episode_sample` API with `Reduce.SAMPLE`.
* Implemented a simple **`TopBottomKFilter`** for selecting top/bottom samples based on reward (heap-based, O(log k) per append).
* Extended both backends:
  * **ConsoleBackend**: pretty-prints structured samples as JSON.
  * **WandbBackend**: logs in `wandb.Table` for live sample inspection.

  
---

#### **Logged Fields**

Each logged sample includes:

```python
{
        "episode_id": episode.episode_id,
        "policy_version": episode.policy_version,
        "prompt": episode.request,
        "response": episode.response,
        "target": str(episode.target),
        **(
            episode.reward_breakdown or {}
        ),  # per-fn breakdown including the average reward
        "advantage": episode.advantage,
        "request_len": episode.request_len,
        "response_len": episode.response_len,
        "pad_id": episode.pad_id,
    }
```

#### Tests:
- Unit tests
- Verified with `python -m apps.grpo.main  --config apps/grpo/qwen3_1_7b.yaml` with modes:
     - `global_reduce` 
     - `per_rank_reduce`: The table is logged in the run `Controller_xxx`
     - `per_rank_no_reduce` The table is logged in the run `Controller_xxx` and with higher latency.
  
---

#### Backend: wandb
image

#### Backend: Console:
```bash
==========  SAMPLE LOGS STEP 1 ==========
[rollout/sample] (2 samples)
{
  "episode_id": "ea2f341b-bf7a-4647-9913-c5f2c8bb525f",
  "policy_version": 0,
  "prompt": "<|im_start|>system\n\n            Put all your scratchpad work between <think> and </think> tags.\n            Your final answer should be between <answer> and </answer> tags otherwise it will not be scored.\n            <|im_end|>\n<|im_start|>user\nJames buys $3000 worth of stuff from Amazon.  He has to return a TV that cost $700 and a bike that cost $500.  He also sells another bike that cost 20% more than the bike he returned for 80% of what he bought it for.  He then buys a toaster for $100.  How much is he out of pocket for everything?<|im_end|>\n<|im_start|>assistant\n",
  "response": "<think>\nOkay, let me try to figure out how much James is out of pocket for everything. Hmm, let's start by breaking down the problem step by step.\n\nFirst, James buys $3000 worth of stuff from Amazon. So, his initial total cost is $3000. But then he has to return a TV that cost $700 and a bike that cost $500. Wait, so when he returns those items, he's getting them back, right? So maybe that means he's losing the value of those items? But I need to check if that's how it's calculated. \n\nBut the question is about how much he's out of pocket. So, when he returns the TV and bike, he's probably losing the amount he paid for them. So, the TV was $700, bike was $500. So, total money he gave back is 700 + 500 = $1200. But he had $3000 initially. So, if he returns $1200, he's out of pocket for those items? Wait, but maybe he has to pay for them back? Or does returning them mean he's just not keeping them? \n\nWait, maybe I need to think in terms of net cash flow. Let me think. When he buys stuff, he spends money. When he returns stuff, he gives money back, so that's like a deduction from his total expenditure. \n\nSo, initial purchase: $3000. Then he returns $700 (TV) and $500 (bike). So, total returned is $1200. Therefore, his net expenditure so far is $3000 - $1200 = $1800. But then he sells another bike that cost 20% more than the bike he returned. Wait, the bike he returned was $500, so the bike he sold is 20% more than that. Let me calculate that. 20% of 500 is 100, so the bike he sold cost 500 + 100 = $600. But he sells it for 80% of what he bought it for. So, he bought it for $600, so 80% of that is 0.8 * 600 = $480. So",
  "target": "2020",
  "MathReward": 0.0,
  "ThinkingReward": 0.2,
  "reward": 0.1,
  "advantage": -7.449986878782511e-05,
  "ref_logprobs": -1.359375,
  "request_len": 512,
  "response_len": 512,
  "pad_id": 151643
}
{
  "episode_id": "b0a1d37d-c3a6-403a-8e9f-8ae510b6702a",
  "policy_version": 0,
  "prompt": "<|im_start|>system\n\n            Put all your scratchpad work between <think> and </think> tags.\n            Your final answer should be between <answer> and </answer> tags otherwise it will not be scored.\n            <|im_end|>\n<|im_start|>user\nJohn uses 5 liters of fuel per km to travel. How many liters of fuel should John plan to use if he plans to travel on two trips of 30 km and 20 km?<|im_end|>\n<|im_start|>assistant\n",
  "response": "<think>\nOkay, let's see. John is traveling twice, first 30 km and then 20 km. He uses 5 liters per km. So I need to calculate the total fuel needed for both trips.\n\nFirst, I should figure out how much fuel he uses for each trip separately and then add them together. For the first trip, 30 km multiplied by 5 liters per km. Let me do that calculation: 30 times 5 is 150 liters. Then for the second trip, 20 km times 5 liters per km. 20 times 5 is 100 liters. Adding those two amounts together, 150 plus 100 equals 250 liters. So John should plan to use 250 liters of fuel in total. I think that's it. Let me double-check. 30 km trip uses 150 liters, 20 km uses 100. Total is indeed 250. Yep, that seems right.\n</think>\n\nJohn uses 5 liters of fuel per km. For the first trip of 30 km, he uses $30 \\times 5 = 150$ liters. For the second trip of 20 km, he uses $20 \\times 5 = 100$ liters. Total fuel required is $150 + 100 = 250$ liters.\n\n<answer>250</answer>",
  "target": "250",
  "MathReward": 1.0,
  "ThinkingReward": 1.0,
  "reward": 1.0,
  "advantage": 0.5398026704788208,
  "ref_logprobs": -4.90625,
  "request_len": 512,
  "response_len": 512,
  "pad_id": 151643
}
==============================================
```

#### **Notes**

- There tend to be a delay on W&B table logging. Typically you'd have to wait for a few more (sometimes over 10) steps for the tables to be visible on wandb.

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -26,7 +26,7 @@ from forge.actors.replay_buffer import ReplayBuffer
 from forge.actors.trainer import TitanTrainer
 from forge.controller.actor import ForgeActor
 from forge.controller.provisioner import init_provisioner, shutdown
-from forge.data.rewards import MathReward, ThinkingReward
+from forge.data.rewards import LanguageReward, MathReward, ThinkingReward
 from forge.data_models.completion import Completion
 from forge.observability.metric_actors import get_or_create_metric_logger
 from forge.observability.metrics import record_metric, Reduce
@@ -158,7 +158,7 @@ def simple_grpo_loss(
     ref_logprobs: torch.Tensor,
     advantages: torch.Tensor,
     padding_mask: torch.Tensor,
-    beta: float = 0.1,
+    beta: float = 1e-6,
 ) -> torch.Tensor:
     logprobs: torch.Tensor = compute_logprobs(logits, response)
     kl = torch.exp(ref_logprobs - logprobs) - (ref_logprobs - logprobs) - 1
@@ -270,10 +270,15 @@ class DatasetActor(ForgeActor):
         self._epoch = 0
 
         def gsm8k_transform(sample):
-            system_prompt = """
-            Put all your scratchpad work between <think> and </think> tags.
-            Your final answer should be between <answer> and </answer> tags otherwise it will not be scored.
-            """
+            system_prompt = """You are a helpful AI assistant that solves math problems.
+
+Please show your reasoning inside <思考></思考> tags, then provide your final numerical answer inside <answer></answer> tags.
+
+Example:
+Question: What is 12 + 5?
+<思考>12と5を足します。12 + 5 = 17です。</思考>
+<answer>17</answer>
+"""
             request: str = sample["question"]
             as_chat = [
                 {"role": "system", "content": system_prompt},
@@ -392,7 +397,17 @@ async def main(cfg: DictConfig):
         ComputeAdvantages.options(**cfg.actors.compute_advantages).as_actor(),
         ReferenceModel.options(**cfg.services.ref_model).as_service(**cfg.ref_model),
         RewardActor.options(**cfg.services.reward_actor).as_service(
-            reward_functions=[MathReward(), ThinkingReward()]
+            reward_functions=[
+                MathReward(),
+                ThinkingReward(tag="思考"),  # Use Japanese tag
+                LanguageReward(
+                    target_language="ja",
+                    tag="思考",
+                    match_reward=2.0,
+                    debug=True,
+                    debug_sample_rate=0.1,
+                ),  # Japanese language reward with debug
+            ]
         ),
     )
 
--- a/apps/grpo/qwen3_1_7b.yaml
+++ b/apps/grpo/qwen3_1_7b.yaml
@@ -5,7 +5,7 @@
 group_size: 8
 local_batch_size: 16 # per-device batch size
 max_req_tokens: 1024
-max_res_tokens: 1024
+max_res_tokens: 2048
 model: "Qwen/Qwen3-1.7B"
 off_by_n: 1 # Off by one by default
 
--- a/apps/grpo/qwen3_8b.yaml
+++ b/apps/grpo/qwen3_8b.yaml
@@ -2,10 +2,10 @@
 # >>> python -m apps.grpo.main --config apps/grpo/qwen3_8b.yaml
 
 # Global configuration
-group_size: 8
-local_batch_size: 12 # per-device batch size
+group_size: 16
+local_batch_size: 4 # per-device batch size
 max_req_tokens: 1024
-max_res_tokens: 1024
+max_res_tokens: 2048
 model: "Qwen/Qwen3-8B"
 off_by_n: 1 # Off by one by default
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dev = [
     "anyio",
     "pytest-asyncio",
     "multiprocess",
+    "langid",
 ]
 docs = [
     "sphinx==7.2.6",
--- a/src/forge/data/rewards.py
+++ b/src/forge/data/rewards.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import random
 import re
 
 
@@ -57,15 +58,28 @@ class MathReward:
 
 
 class ThinkingReward:
-    """Reward class for evaluating use of <think> tags in reasoning."""
+    """Reward class for evaluating use of thinking tags in reasoning.
 
-    def __init__(self, partial_reward: float = 0.2, full_reward: float = 1.0):
+    Args:
+        partial_reward: Reward for partial tag usage (incomplete/malformed)
+        full_reward: Reward for well-formed thinking blocks with content
+        tag: Tag name to use (default "think", can use "思考" for Japanese, etc.)
+    """
+
+    def __init__(
+        self, partial_reward: float = 0.2, full_reward: float = 1.0, tag: str = "think"
+    ):
         self.partial_reward = partial_reward
         self.full_reward = full_reward
+        self.tag = tag
+        # Build regex patterns for the specified tag
         self._THINK_BLOCK_RE = re.compile(
-            r"<\s*think\s*>(.*?)<\s*/\s*think\s*>", re.IGNORECASE | re.DOTALL
+            rf"<\s*{re.escape(tag)}\s*>(.*?)<\s*/\s*{re.escape(tag)}\s*>",
+            re.IGNORECASE | re.DOTALL,
         )
-        self._THINK_TAG_ATTEMPT_RE = re.compile(r"<\s*/?\s*think\s*>", re.IGNORECASE)
+        self._THINK_TAG_ATTEMPT_RE = re.compile(
+            rf"<\s*/?\s*{re.escape(tag)}\s*>", re.IGNORECASE
+        )
 
     def __call__(self, prompt: str, response: str, target: str | None = None) -> float:
         """Compute thinking reward."""
@@ -80,3 +94,128 @@ class ThinkingReward:
         elif has_attempt:
             return self.partial_reward
         return 0.0
+
+
+class LanguageReward:
+    """Reward class for evaluating the language used in responses.
+
+    This reward uses langid to detect the language and rewards responses that use
+    the target language. The detection strategy depends on the format:
+    - If exactly one thinking block: detect language of the block content
+    - Otherwise (no blocks or multiple blocks): detect language of whole response
+
+    Note: Format enforcement (single vs multiple blocks) is handled by ThinkingReward.
+    This reward focuses purely on language detection.
+
+    Args:
+        target_language: ISO 639-1 language code (e.g., 'en', 'ja', 'zh', 'es')
+        match_reward: Reward when detected language matches target (default: 1.0)
+        no_match_reward: Reward when language doesn't match (default: 0.0)
+        tag: Tag name to use (default "思考" for multilingual, can use "think", etc.)
+        debug: If True, print debug samples showing model outputs and detected language
+        debug_sample_rate: Fraction of calls to debug (e.g., 0.1 = 10% of calls)
+
+    Note: Requires langid to be installed. Install with: pip install langid
+    """
+
+    def __init__(
+        self,
+        target_language: str = "ja",
+        match_reward: float = 1.0,
+        no_match_reward: float = 0.0,
+        tag: str = "思考",
+        debug: bool = False,
+        debug_sample_rate: float = 0.1,
+    ):
+        self.target_language = target_language
+        self.match_reward = match_reward
+        self.no_match_reward = no_match_reward
+        self.tag = tag
+        self.debug = debug
+        self.debug_sample_rate = debug_sample_rate
+        self._debug_counter = 0
+        # Build regex pattern for the specified tag
+        self._THINK_BLOCK_RE = re.compile(
+            rf"<\s*{re.escape(tag)}\s*>(.*?)<\s*/\s*{re.escape(tag)}\s*>", re.DOTALL
+        )
+
+        # Lazy import langid with helpful error message
+        try:
+            import langid
+
+            self._langid = langid
+        except ImportError:
+            raise ImportError(
+                "langid is required for LanguageReward but is not installed. "
+                "Please install it with: pip install langid"
+            ) from None
+
+    def __call__(self, prompt: str, response: str, target: str | None = None) -> float:
+        """Compute language reward based on detected language.
+
+        Detection strategy:
+        - If exactly one thinking block: detect language of block content
+        - Otherwise: detect language of whole response
+
+        Args:
+            prompt: The input prompt (unused but kept for signature consistency)
+            response: The model response
+            target: Optional target string (unused but kept for signature consistency)
+
+        Returns:
+            match_reward if detected language matches target, no_match_reward otherwise
+        """
+
+        # TODO: refactor pending https://github.com/meta-pytorch/torchforge/issues/187
+        should_debug = self.debug and (random.random() < self.debug_sample_rate)
+
+        if not response:
+            if should_debug:
+                print(
+                    f"\n[LanguageReward] Empty response | Reward: {self.no_match_reward}"
+                )
+            return self.no_match_reward
+
+        # Extract all thinking blocks
+        matches = self._THINK_BLOCK_RE.findall(response)
+
+        # Determine what text to analyze
+        if len(matches) == 1:
+            # Single block: detect language of block content only
+            text_to_analyze = matches[0].strip()
+            detection_mode = "single block"
+        else:
+            # No blocks or multiple blocks: detect language of whole response
+            text_to_analyze = response.strip()
+            detection_mode = f"{len(matches)} blocks, using whole response"
+
+        # Remove extra whitespace
+        text_to_analyze = re.sub(r"\s+", " ", text_to_analyze).strip()
+
+        if not text_to_analyze:
+            if should_debug:
+                print(f"\n[LanguageReward] Empty text | Reward: {self.no_match_reward}")
+            return self.no_match_reward
+
+        # Detect language using langid
+        detected_lang, confidence = self._langid.classify(text_to_analyze)
+
+        # Check if language matches target
+        reward = (
+            self.match_reward
+            if detected_lang == self.target_language
+            else self.no_match_reward
+        )
+
+        if should_debug:
+            sample = text_to_analyze[:1000].replace("\n", " ")
+            match_symbol = "✓" if detected_lang == self.target_language else "✗"
+            print(
+                f"\n[LanguageReward] Detection mode: {detection_mode}"
+                f"\n  Target: {self.target_language} | Detected: {detected_lang} | "
+                f"Confidence: {confidence:.2f}"
+                f"\n  Sample: {sample}..."
+                f"\n  → Reward: {reward} {match_symbol}"
+            )
+
+        return reward
--- a/tests/unit_tests/rl/test_language_reward.py
+++ b/tests/unit_tests/rl/test_language_reward.py
@@ -1,0 +1,266 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import sys
+import unittest
+from unittest.mock import patch
+
+
+class TestLanguageReward(unittest.TestCase):
+    def setUp(self):
+        """Set up test fixtures before each test method."""
+        # Import after patching to avoid ImportError
+        from forge.data.rewards import LanguageReward
+
+        self.LanguageReward = LanguageReward
+        self.reward_en = LanguageReward(target_language="en")
+        self.reward_ja = LanguageReward(target_language="ja")
+
+    def test_init_default_values(self):
+        """Test LanguageReward initialization with default values."""
+        reward = self.LanguageReward()
+        self.assertEqual(reward.target_language, "ja")
+        self.assertEqual(reward.match_reward, 1.0)
+        self.assertEqual(reward.no_match_reward, 0.0)
+
+    def test_init_custom_values(self):
+        """Test LanguageReward initialization with custom values."""
+        reward = self.LanguageReward(
+            target_language="ja",
+            match_reward=0.9,
+            no_match_reward=0.1,
+        )
+        self.assertEqual(reward.target_language, "ja")
+        self.assertEqual(reward.match_reward, 0.9)
+        self.assertEqual(reward.no_match_reward, 0.1)
+
+    def test_init_missing_langid(self):
+        """Test LanguageReward initialization without langid installed."""
+        # Remove langid from modules if it exists
+        langid_module = sys.modules.get("langid")
+        if "langid" in sys.modules:
+            del sys.modules["langid"]
+
+        with patch.dict("sys.modules", {"langid": None}):
+            with self.assertRaises(ImportError) as context:
+                # Re-import to trigger the ImportError
+                import importlib
+
+                import forge.data.rewards
+
+                importlib.reload(forge.data.rewards)
+                forge.data.rewards.LanguageReward()
+
+            self.assertIn("langid is required", str(context.exception))
+            self.assertIn("pip install langid", str(context.exception))
+
+        # Restore langid module if it existed
+        if langid_module is not None:
+            sys.modules["langid"] = langid_module
+
+    def test_regex_pattern(self):
+        """Test that regex pattern is compiled correctly."""
+        reward = self.LanguageReward()
+        self.assertIsNotNone(reward._THINK_BLOCK_RE)
+
+    def test_call_with_english_thinking(self):
+        """Test __call__ with English text in thinking blocks."""
+        response = "<思考>This is English reasoning about math problems.</思考>"
+        result = self.reward_en("prompt", response)
+        self.assertEqual(result, 1.0)
+
+    def test_call_with_japanese_thinking(self):
+        """Test __call__ with Japanese text in thinking blocks."""
+        response = "<思考>これは日本語で考えています。数学の問題を解きます。</思考>"
+        result = self.reward_ja("prompt", response)
+        self.assertEqual(result, 1.0)
+
+        # English reward should give no_match_reward for Japanese text
+        result = self.reward_en("prompt", response)
+        self.assertEqual(result, 0.0)
+
+    def test_call_with_chinese_thinking(self):
+        """Test __call__ with Chinese text in thinking blocks."""
+        response = "<思考>这是中文思考。我们需要解决这个数学问题。</思考>"
+        reward_zh = self.LanguageReward(target_language="zh")
+        result = reward_zh("prompt", response)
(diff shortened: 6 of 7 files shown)


apps/grpo/main.py
```diff
@@ -173,24 +185,29 @@ async def evaluate_response(self, prompt: str, response: str, target: str) -> fl
...
+            # count fn calls
             record_metric(
                 f"reward/evaluate_response/count_{reward_fn_name}_calls",
                 1,
                 Reduce.SUM,
             )
 
         avg_reward = total_rewards / len(self.reward_functions)
-        return avg_reward
+        reward_breakdown["reward"] = avg_reward
```
felipemello1: I dont like too much having to deal with "reward" being a specific key. E.g., one of the reward function names could be "reward".

What if we just do:
```
return reward_breakdown, reward_mean
```

it looks safer to me

apps/grpo/main.py
```diff
@@ -396,11 +417,14 @@ async def continuous_rollouts():
                 episode.ref_logprobs = ref_logprobs[i]
             del ref_logprobs, input_ids
 
+            # Calculate advantages and add to replay buffer
             advantages = await compute_advantages.compute.call_one(episodes)
             for episode, advantage in zip(episodes, advantages):
                 episode.advantage = advantage
                 await replay_buffer.add.call_one(episode)
+                record_episode_sample("rollout/sample", episode)
```
felipemello1: i usually try to do the pattern {class_name}/{function_name}/{metric_name}

Maybe we could do "main_samples/continuous_rollouts/reward_table"?

src/forge/observability/metrics.py
```diff
@@ -135,12 +140,32 @@ def reduce_metrics_states(states: list[dict[str, dict[str, Any]]]) -> list[Metri
...
+        >>> metrics = reduce_metrics_states(states)
+        >>> for m in metrics:
+        ...     print(m)
+        Metric(key='loss', value=2.0, reduction=Reduce.MEAN)
+        Metric(
+            key='reward/sample',
+            value=[{'episode_id': 1, 'reward': 0.5},
+                   {'episode_id': 2, 'reward': 1.0}],
+            reduction=Reduce.SAMPLE,
+        )
```
felipemello1: i think we can keep the minimal example that was here before, and in Reduce.SAMPLE put more details if we have to

src/forge/observability/metrics.py
```diff
@@ -182,6 +207,79 @@ def reduce_metrics_states(states: list[dict[str, dict[str, Any]]]) -> list[Metri
...
+        "target": str(episode.target),
+        **(
+            episode.reward_breakdown or {}
+        ),  # per-fn breakdown including the average reward
+        "advantage": episode.advantage,
+        "request_len": episode.request_len,
+        "response_len": episode.response_len,
+        "pad_id": episode.pad_id,
+    }
+    record_metric(table_name, sample, Reduce.SAMPLE)
```
felipemello1: since the underlying API  of `record_metric` remains the same, IMO `record_episode_sample` adds cognitive load.

All this function is doing is converting episode to a dictionary, right? Then lets go to the episode dataclass and add a method 'to_dict".

Then the user can do:

`record_metric(key, sample.to_dict(), Reduce.SAMPLE)`, and keep the same pattern they are used to. That way, we teach them that this actually works for **ANY** dict, not only this sample dict.

src/forge/observability/metrics.py
```diff
@@ -182,6 +207,79 @@ def reduce_metrics_states(states: list[dict[str, dict[str, Any]]]) -> list[Metri
...
+
+
+#################
+# SampleFilters #
+#################
+
+
+class TopBottomKFilter:
```
felipemello1: This is a good abstraction, but If we go this route, then we probably need an ABC class for filter too. We also have nice parameters here like top_k, bottom_k, key, but dont give users any way to access them or to change the filter.

Exposing these options would require touching multiple functions and adding more to the config, which we decided not to do.

So it makes me think that its too early for an abstraction that we dont have practical use for. I would prefer to have this logic embedded in SampleAccumulator.

Here is what i would like to propose:
1. Rename SampleAccumulator -> TopBottomSampleAccumulator
2. Hardcode it to look for the key "filter_score" --> this way, users can use this key for whatever is more convenient to them, e.g. length, count of words, any specific reward, etc.
3. "filter_score" key should probably be a str that we import, since multiple functions are using it. Something like:

`from forge.observability.metrics import FILTER_SCORE_KEY`

src/forge/observability/metrics.py
```diff
@@ -896,13 +1047,58 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
...
+            # Add rows (fill missing columns with None)
+            for s in table_rows:
+                values = [s.get(c) for c in table.columns]
+                table.add_data(*values)
+
+            # Log the same table object (INCREMENTAL update)
+            self.run.log({f"{table_name}_table": table})
+            logger.info(
+                f"WandbBackend: Appended {len(table_rows)} rows to incremental table '{table_name}' at step {step}"
+            )
```
felipemello1:
```suggestion
```

src/forge/observability/metrics.py
```diff
@@ -896,13 +1047,58 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
...
+            else:
+                table = self._tables[table_name]
+
+            # Add rows (fill missing columns with None)
+            for s in table_rows:
+                values = [s.get(c) for c in table.columns]
+                table.add_data(*values)
+
+            # Log the same table object (INCREMENTAL update)
+            self.run.log({f"{table_name}_table": table})
```
felipemello1:
```suggestion
            self.run.log({table_name: table})
```

src/forge/observability/metrics.py
```diff
@@ -896,13 +1047,58 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
...
         if self.run:
+            # Convert each incremental table to immutable before finishing
+            for table_name, incr_table in self._tables.items():
+                final_table = wandb.Table(
+                    columns=incr_table.columns,
+                    data=incr_table.data,
+                    log_mode="IMMUTABLE",
+                )
+                self.run.log({table_name: final_table})
+                logger.info(f"WandbBackend: Finalized table {table_name}")
```
felipemello1:
```suggestion
                logger.debug(f"WandbBackend: Finalized table {table_name}")
```

src/forge/observability/metrics.py
```diff
@@ -896,13 +1047,58 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
...
     def get_metadata_for_secondary_ranks(self) -> dict[str, Any]:
         if self.run and self.per_rank_share_run:
             return {"shared_run_id": self.run.id}
         return {}
 
     async def finish(self) -> None:
+        import wandb
+
         if self.run:
+            # Convert each incremental table to immutable before finishing
```
felipemello1: why do we need to do this? Is it part of their best practices? I would imagine that if we are finishing the run then we dont need to log a "final_table".

could you please research a bit and see if we can delete this code?

src/forge/observability/metrics.py
```diff
@@ -896,13 +1047,58 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
...
+        for table_name, table_rows in samples.items():
+            if not table_rows:
+                continue
+
+            # If table doesn't exist yet, create it in INCREMENTAL mode
+            if table_name not in self._tables:
+                columns = list(table_rows[0].keys())
+                table = wandb.Table(columns=columns, log_mode="INCREMENTAL")
+                self._tables[table_name] = table
+                logger.info(
```
felipemello1:
```suggestion
                logger.debug(
```

src/forge/observability/metrics.py
```diff
@@ -896,13 +1047,58 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
...
+        if not self.run:
+            return
+
+        for table_name, table_rows in samples.items():
+            if not table_rows:
+                continue
+
+            # If table doesn't exist yet, create it in INCREMENTAL mode
+            if table_name not in self._tables:
+                columns = list(table_rows[0].keys())
```
felipemello1: its risky to use only the first row. Lets use a set and iterate over all rows?

src/forge/observability/metrics.py
```diff
@@ -750,105 +899,102 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
     async def finish(self) -> None:
         pass
 
+    async def log_samples(self, samples: Dict[str, List[dict]], step: int) -> None:
+        """Pretty-print sample-level logs to console."""
+        import json
```
felipemello1: we can import json at the top

src/forge/observability/metrics.py
```diff
@@ -750,105 +899,102 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
     async def finish(self) -> None:
         pass
 
+    async def log_samples(self, samples: Dict[str, List[dict]], step: int) -> None:
```
felipemello1: i dont like the API here. All the other logs communicate `Metric` or `list[Metric]`. I think we should preserve it.

felipemello1: i am debating with myself if we should have `log_batch_samples` and `log_stream_samples`.

If we keep only 1, then: 
- its a bit weird that we have `log_stream` and `log_batch`, but this doesnt apply to samples. 
- We also need to do asyncio.create_task.
- in the stream mode, its a bit too verbose with all the "=====" printing.

If we keep 2, then its a new method for every backend.

Any thoughts?

src/forge/observability/metrics.py
```diff
@@ -559,7 +705,13 @@ def push(self, metric: Metric) -> None:
 
         # For PER_RANK_NO_REDUCE backends: stream without reduce
         for backend in self.per_rank_no_reduce_backends:
-            backend.log_stream(metric=metric, global_step=self.global_step)
+
+            if metric.reduction == Reduce.SAMPLE:
+                # Wrap singleton Metric into expected {key: [list_of_dicts]} format
+                sample = {metric.key: [metric.value]}
```
felipemello1: i dont think we need that. We can pass metric directly and let the logger handle it. The idea is that we always pass the information as Metric, so we know what to expect

felipemello1 (changes requested):
First pass. Thanks for the PR! Multiple people are eager to see it landing.

I left some comments. I think we are close to landing, but we need to make some changes. Thank you for moving forward with this!

src/forge/observability/metrics.py
DNXie: I added this because this is the recommended practice from wandb. See [doc](https://docs.wandb.ai/models/tables/log_tables#incremental-mode).

src/forge/observability/metrics.py
DNXie: I am not sure about this one. I don't think integrate this sampling logic into the `SampleAccumulator` is a good idea. As we've already discussed in https://github.com/meta-pytorch/torchforge/issues/301,
only supporting this one filtering function might already be too restrictive even at the experimental stage. In practice we often want more than just the absolute best and worst samples. I think in the future, we would want to expose the API to the user and let them implement customized filtering functions. If we can agree on this, then separating them should be the right choice.

src/forge/observability/metrics.py
DNXie: I agree. But we actually didn't include all the entries from Episode. Specifically, we excluded `ref_logprobs`, `pad_id`,  and `completion`. I can do a `to_sample_dict` to include only the entries we are interested in.

src/forge/observability/metrics.py
DNXie: We cannot remove the `_table` postfix. wandb would have a problem with this and not showing/recognizing the table

src/forge/observability/metrics.py
felipemello1: how about we just do `to_dict(exclude:listr[str])` or just `sample_dict = to_dict(); del sample_dict["pad_id", "completion", "ref_logprobs"]`?

src/forge/observability/metrics.py
felipemello1: oh wow, thats surprising that "_table" in the name is a hard requirements if you are already giving a Table object. If you are sure about it, then please add a comment explaining why table is necessary.

src/forge/observability/metrics.py
felipemello1: I will be annoying on this one :(

Lets make a deal: If/when users complain about the filtering and there is a need to expose the API, we come back to this design. It should be easy to revert it back. Until then, I think they can get pretty far by customizing the filter_score logic on the recipe. Thanks for pushing back and discussing these options with me!

src/forge/observability/metrics.py
felipemello1: Can you please reference the link as a comment?

apps/grpo/main.py
DNXie: I will name it `main_samples/continuous_rollouts/sample_table`

src/forge/observability/metrics.py
DNXie: Since now we are callling 
```python
record_metric(
                "main/continuous_rollouts/count_rollout_iterations", 1, Reduce.SUM
            )
```
We don't need a "_table" here

But I added a check here just in case
```python
# table_name has to end with _table to be recognized by wandb
            if not table_name.endswith("_table"):
                table_name += "_table"
            self.run.log({f"{table_name}": table})
```

apps/grpo/main.py
DNXie: Done

src/forge/observability/metrics.py
DNXie: Reverted back

src/forge/observability/metrics.py
DNXie: Done.

src/forge/observability/metrics.py
DNXie: Done

src/forge/observability/metrics.py
DNXie: Done

src/forge/observability/metrics.py
DNXie: removed

src/forge/observability/metrics.py
DNXie: > in the stream mode, its a bit too verbose with all the "=====" printing.

What do you expect to see in stream mode?

src/forge/observability/metrics.py
DNXie: Done

felipemello1:
thanks @DNXie! i will review it tomorrow. Do you mind solving the conflicts too when you have a chance?

src/forge/observability/metrics.py
DNXie: I would have to add the sample example back. I simplified it though. As I was testing the code and debugging, I found it so hard to understand the behavior and debug without this example here.

apps/grpo/main.py
```diff
@@ -73,6 +76,32 @@ def response_tensor(self) -> torch.Tensor:
...
+            "advantage": self.advantage,
+            "request_len": self.request_len,
+            "response_len": self.response_len,
+            "pad_id": self.pad_id,
+            "ref_logprobs": self.ref_logprobs,
+            "completion": self.completion,
+        }
+
+        if self.reward_breakdown is not None and "reward_breakdown" not in exclude:
+            result.update(self.reward_breakdown)
```
felipemello1, line 96: this is flattening the rewards, but it is not obvious that this is what we are doing. I wonder if (a) we should flatten. But assuming we should, lets add a comment here.

Do you see any strong argument for not leaving it as a dictionary? not sure if this is poorly displayed in wandb

apps/grpo/main.py
```diff
@@ -47,10 +47,13 @@ class Episode:
     request_len: int
     response_len: int
     target: Any | None = None
+    request: str | None = None
+    response: str | None = None
```
felipemello1, line 50: do we need those? I believe this is redundant. They are already in `Completion`. It would be redundant.

https://github.com/DNXie/forge/blob/main/src/forge/data_models/completion.py#L19C5-L23C14

apps/grpo/main.py
```diff
@@ -73,6 +76,32 @@ def response_tensor(self) -> torch.Tensor:
             tensor = F.pad(tensor, (0, diff), value=self.pad_id)
         return tensor
 
+    def to_dict(self, exclude: list[str] | None = None) -> dict[str, Any]:
```
felipemello1, line 78: I am thinking that we could add some of the fields from Completion, i.e. prompt, text,  stop_reason, generator_version, metadata. Wdyt?

src/forge/observability/metrics.py
```diff
@@ -182,6 +201,17 @@ def reduce_metrics_states(states: list[dict[str, dict[str, Any]]]) -> list[Metri
     return reduced_metrics
 
 
+def record_episode_sample(table_name: str, episode):
```
felipemello1: i think we can delete this and in the code just do:

```
sample = episode.to_dict(exclude=["ref_logprobs", "completion"])
record_metric(table_name, sample, Reduce.SAMPLE)
```

I dont think we gain much with this abstraction

src/forge/observability/metrics.py
```diff
@@ -68,6 +72,7 @@ class Reduce(Enum):
     MAX = "max"
     MIN = "min"
     STD = "std"
+    SAMPLE = "sample"
```
felipemello1, line 75: I am bothered by the term "Sample". In reality, we can log anything that is a dictionary / table. It doesnt need to be a sample, right?

It is like having a `Reduce.TIME`, when we are just doing .mean, for example.

I think we could rename it to `TABLE`, and keep it aligned with wandb API, which is log.table.

i.e., rename `sample` -> `table` everywhere

felipemello1: Before you go with the change, do you see a better alternative? Should it be `sample` -> `dict`?

src/forge/observability/metrics.py
```diff
@@ -392,6 +422,79 @@ def reset(self) -> None:
...
+class SampleAccumulator(MetricAccumulator):
+    """Accumulator for sample-level metrics with top-k and bottom-k filtering.
+
+    Keeps the top-k and bottom-k samples by a given key (e.g., reward).
+    Useful for logging only the best and worst samples from a batch.
+    """
+
+    def __init__(
+        self, reduction: Reduce, top_k: int = 1, bottom_k: int = 1, key: str = "reward"
+    ):
```
felipemello1, line 434: i think that the key name should be "score". If someone wants to rank based on anything that is not reward (e.g. length), they can just do:

```mydict["score"] = len(mydict["response"])```

File: apps/grpo/main.py (hunk 15 above)
Comment (DNXie) [reply apps/grpo/main.py:96]: The `record_episode_sample` function needs this to be flattened. I can flatten this alternatively in `record_episode_sample`, wdyt?

Comment (felipemello1) [reply apps/grpo/main.py:96]: i think we should delete record_episode_sample :X 

Why does it need it to be flat?

File: src/forge/observability/metrics.py [hunk 21]
```diff
@@ -750,6 +867,16 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
...
 
+    async def log_samples(self, samples: List[Metric], step: int) -> None:
+        """Pretty-print sample-level logs to console."""
+
+        logger.info(f"==========  SAMPLE LOGS STEP {step} ==========")
+        for sample in samples:
+            table_name, table_rows = sample.key, sample.value
+            logger.info(f"[{table_name}] ({len(table_rows)} samples)")
+            logger.info(json.dumps(table_rows, indent=2, ensure_ascii=False))
+        logger.info("==============================================\n")
```
felipemello1:
```suggestion
        for sample in samples:
            table_name, table_rows = sample.key, sample.value
            logger.info(f"[{table_name}] ({len(table_rows)} samples)")
            logger.info(json.dumps(table_rows, indent=2, ensure_ascii=False))
```

src/forge/observability/metrics.py
```diff
@@ -896,13 +1024,71 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
...
+            return
+
+        for sample in samples:
+            table_name, table_rows = sample.key, sample.value
+            if not table_rows:
+                continue
+
+            # Convert to list if single sample. This happens when logging stream
+            if isinstance(table_rows, dict):
+                table_rows = [table_rows]
```
felipemello1: instead, when creating the metric, can we make sure that its always a list, so we dont need this check in multiple places?

src/forge/observability/metrics.py
```diff
@@ -896,13 +1024,71 @@ def log_stream(self, metric: Metric, global_step: int, *args, **kwargs) -> None:
...
+                columns = sorted(columns)  # Sort for consistent column ordering
+                table = wandb.Table(columns=columns, log_mode="INCREMENTAL")
+                self._tables[table_name] = table
+                logger.debug(
+                    f"WandbBackend: Created new incremental table: {table_name} with columns: {columns}"
+                )
+            else:
+                table = self._tables[table_name]
+
+            # Add rows (fill missing columns with None)
```
felipemello1, line 1061: what happens if the `table_rows` has a new column not present in `table.columns`? it looks like we would drop this value silently. Can we add a warning? I think we can do that with sets. I havent used python sets api in a long time, but i think something like:

```
input_extra_columns = table_rows.diff(set(table.columns))
if input_extra_columns>
     logger.info("input table {table_name} has not expected column {input_extra_columns}. Extra columns will be ignored.")
```

apps/grpo/main.py
DNXie:
```python
def record_episode_sample(table_name: str, episode):
    """
    Record a structured sample-level log for a single episode.
    Args:
        table_name (str): logging prefix (e.g. "rollout/sample").
        episode (Episode): episode object with filled attributes.
    """
    sample = episode.to_dict(exclude=["ref_logprobs", "completion"])
    record_metric(table_name, sample, Reduce.SAMPLE)

```
Because `record_episode_sample` calls `record_metric` where `sample` needs to be a flattened dict. 
I don't think deleting the API is a good idea as it has some ad-hoc process of the dict, which we don't want to expose to the users.

DNXie: I could exclude it but just so you know that I excluded `ref_logprobs ` is that it requires further processing to be logged. Otherwise it would lead to an empty table.

src/forge/observability/metrics.py
DNXie: I think table sounds good

src/forge/observability/metrics.py
DNXie: Normally it is always a list, this particular case just happen when logging stream. So I don't see how I can avoid these two lines of code. We handle it either here or before calling this function in the streaming mode

src/forge/observability/metrics.py
DNXie: Fine. but now with `sample['score']=sample['reward']`

src/forge/observability/metrics.py
DNXie: But do we also change `SampleAccumulator` to `TableAccumulator`? I don't think it is a good idea, because the logic of the accumulator is to handle samples not tables specifically.

apps/grpo/main.py
DNXie: I think it is more straightforward to keep record of them now because [completion.prompt](https://github.com/DNXie/forge/blob/main/src/forge/data_models/completion.py#L20) is of type `Prompt` instead of str, which is a sequence of conversations. If we don't keep record of `request:str` here, which is the last message, we have to deal with this `Prompt` type in `to_dict` which is really not a good way to make it general.

apps/grpo/main.py
DNXie: prompt and text are already covered. I think we either flatten the entire completion or do not introduce much overwhelming information here. In my opinion, only prompt and response are the most important thing. I think we could just keep it this way. After all we are just providing an interface and a use case of the API. The users can decide what information they need. Let's add them when we see needs

felipemello1 (approved):
LG! Thanks for addressing all of my comments and getting this to the finish line!

PR #507, opened 2025-10-28
Fix learning rate logging when using LRSchedulersContainer

DNXie:
Now the lr starts from 1e-05 (specified in the config) instead of 0.001

Log: 
```
rl_trainer/learning_rate: 1e-05
```

image

### DIFF
--- a/src/forge/actors/trainer.py
+++ b/src/forge/actors/trainer.py
@@ -166,11 +166,7 @@ class RLTrainer(ForgeActor):
 
         t.step("forward_backward")
 
-        current_lr = (
-            self.engine.lr_schedulers.get_last_lr()[0]
-            if hasattr(self.engine.lr_schedulers, "get_last_lr")
-            else 0.001
-        )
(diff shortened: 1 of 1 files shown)


DNXie:
cc @wukaixingxp

wukaixingxp:
Thanks for the quick fix, just curious what should we do if `self.engine.lr_schedulers.schedulers` has more than one lr_scheduler in the list.. any reason we make the `self.engine,lr_schedulers.schedulers` a list?

DNXie:
@casteryh @wukaixingxp 
I am not 100% sure why there would be multiple lr schedulers either.. 
But according to the [docstring](https://github.com/pytorch/torchtitan/blob/89c631cdcefd05885af511513507099148f2bd1d/torchtitan/components/lr_scheduler.py#L40) here, all the lr schedulers are the same. And in this [example](https://github.com/pytorch/torchtitan/blob/89c631cdcefd05885af511513507099148f2bd1d/torchtitan/train.py#L497), it also only accesses the first lr_scheduler.

PR #509, opened 2025-10-29
[sft] load chat_template.jinja if available

jfncs:
Add support for loading `chat_template.jinja` in HuggingFaceModelTokenizer.

If the file is missing, fall back to using `chat_template` in tokenizer_config.json.

Ref: https://github.com/huggingface/transformers/pull/33957

### DIFF
--- a/src/forge/data/tokenizer.py
+++ b/src/forge/data/tokenizer.py
@@ -42,7 +42,6 @@ class HuggingFaceBaseTokenizer(BaseTokenizer):
         *,
         tokenizer_config_json_path: str | None = None,
         generation_config_path: str | None = None,
-        chat_template_path: str | None = None,
     ):
         self.tokenizer = Tokenizer.from_file(tokenizer_json_path)
         if not (tokenizer_config_json_path or generation_config_path):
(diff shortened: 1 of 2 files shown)


felipemello1:
hey @daniellepintz , mind taking a look since its related to a previous PR you wrote?

src/forge/data/tokenizer.py
```diff
@@ -51,6 +52,10 @@ def __init__(
         if tokenizer_config_json_path:
             with open(tokenizer_config_json_path, "rb") as f:
                 self.config = json.load(f)
+            if chat_template_path:
```
daniellepintz: instead of overwriting this, why not check in `HuggingFaceModelTokenizer` if the path exists and if it does then create the `self.template` directly from the path? 
then no need to pass chat_template_path to `HuggingFaceBaseTokenizer`

daniellepintz:
Thanks for the PR @jaysonfrancis, mind adding a bit of description to summary and also a test plan?

src/forge/data/tokenizer.py
jfncs: thanks @daniellepintz, that makes sense!

PR #511, opened 2025-10-30
[wip] Add DeepseekV3 SFT config

daniellepintz:
Test Plan:
```
python -m apps.sft.main --config apps/sft/deepseek_v3.yaml
```

### DIFF
--- a/apps/sft/deepseek_v3.yaml
+++ b/apps/sft/deepseek_v3.yaml
@@ -1,0 +1,70 @@
+# >>> python -m apps.sft.main --config apps/sft/deepseek_v3.yaml
+
+
+# TODO: required by torchtitan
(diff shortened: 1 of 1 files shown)


apps/sft/deepseek_v3.yaml
```diff
@@ -0,0 +1,70 @@
...
+  seq_len: 2048
+  max_norm: 1.0
+  steps: 1000
+  compile: false
+  dataset: "c4"
+
+parallelism:
+  data_parallel_replicate_degree: 1
+  data_parallel_shard_degree: -1
+  tensor_parallel_degree: 1
```
ebsmothers, line 39: Based on a conversation with @tianyu-l this should work for 32 nodes
```suggestion
  data_parallel_shard_degree: -1
  tensor_parallel_degree: 8
```

It also may be worth looking at some of the runs [here](https://verl.readthedocs.io/en/latest/perf/dpsk.html#deepseek-671b) for reference.

Not sure of the latest status of ETP in titan but it'd be worth confirming that as well.

PR #512, opened 2025-10-30
[RFC] - Config is code

felipemello1:
This PR attempts to showcase the different styles we could use for our config system. A push here is if we can leave the .yaml and use .py instead, under the reasoning that "config is code". 

Python can enable a) better type safety, b) import directly instead of using strings, and c) more flexibility for the user to define custom code within the config.

On the other hand, yaml can be easier to read and hydra provides some nice freebies.

Must have:
- Lazy instantiation: Currently we don't have any way to instantiate a function through configs, e.g. 'loss: my_loss_fn'.
- Handle complex patterns, e.g. nesting

Nice to have:
- Composition: Our code is naturally fragmented (generator, rewards, etc). Touching one config without touching the others is a relevant feature.
- Ability to code in the config, e.g. if statements, sum values, check conditions, etc.
- CLI overrides

Proposed options:
1. OmegaConfg / Hydra
2. Fiddle (cfg option by google)
3. Factory + dataclasses
4. Plain python with partials
5. Plain python with dictionary
TODO: toml

I would like to hear what others think about the options, if we should stick with yaml and fully explore hydra/omegaconf or change to .py system.

### DIFF
--- a/brainstorming/configs/config_dataclasses.py
+++ b/brainstorming/configs/config_dataclasses.py
@@ -1,0 +1,131 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Dataclass config with inner Config classes.
+
+Pros:
+- Type safety for instantiated configs
+
+Cons:
+- Requires modifying target classes (not feasible for external libraries - Need to use wrapper)
+- Boilerplate (every class needs Config + __init__)
+
+eg:
+
+```
+class TokenizerWithConfig:
+    @dataclass
+    class Config:
+        path: str
+
+        def build(self) -> "TokenizerWithConfig":
+            return TokenizerWithConfig(self)
+
+    def __init__(self, config: Config):
+        self.config = config
+        self.path = config.path
+```
+"""
+
+from dataclasses import dataclass
+
+import torch
+
+from mock_with_config import (
+    ComponentConfig,
+    LlamaModelWithConfig,
+    MultiHeadAttentionWithConfig,
+    TokenizerWithConfig,
+)
+
+
+@dataclass
+class DataArgs:
+    """Plain dataclass for non-instantiated config block (PATTERN 4)."""
+
+    batch_size: int = 4
+    shuffle: bool = True
+
+
+def llama3_2_1b_full():
+    output_dir = "/tmp/torchtune/llama3_2_1B/full"
+
+    return {
+        "output_dir": output_dir,
+        # PATTERN 1: Simple Component Instantiation
+        "tokenizer": TokenizerWithConfig.Config(
+            path="/tmp/Llama-3.2-1B-Instruct/original/tokenizer.model",
+        ),
+        # PATTERN 2: Component with Nested Instantiation
+        "model": LlamaModelWithConfig.Config(
+            attn_config=MultiHeadAttentionWithConfig.Config(
+                num_heads=32,
+            )
+        ),
+        # PATTERN 3: Component Needing Runtime Args (Partial)
+        "optimizer": ComponentConfig(
+            component_cls=torch.optim.AdamW,
+            kwargs={"lr": 2e-5},
+        ),
+        # PATTERN 4: Non-Instantiated Config Block (Plain Data)
+        "data_args": DataArgs(
+            batch_size=4,
+            shuffle=True,
+        ),
+        # PATTERN 5: Plain Top-Level Hyperparameters
+        "epochs": 1,
+        "gradient_accumulation_steps": 8,
+    }
+
+
+if __name__ == "__main__":
+    # =========================================================================
+    # Scenario 1: Basic Instantiation
+    # =========================================================================
+    cfg = llama3_2_1b_full()
+
+    # PATTERN 1: Simple Component Instantiation
+    tokenizer = cfg["tokenizer"].build()
+
+    # PATTERN 2: Component with Nested Instantiation
+    model = cfg["model"].build()
+
+    # PATTERN 3: Component Needing Runtime Args (Partial)
+    optimizer = cfg["optimizer"].build(model.parameters())
+
+    # =========================================================================
+    # Scenario 2: Override Config Values
+    # =========================================================================
+    cfg2 = llama3_2_1b_full()
+
+    # PATTERN 1: Simple Component Instantiation
+    cfg2["tokenizer"].path = "/new/path"
+
+    # PATTERN 2: Component with Nested Instantiation
+    cfg2["model"].attn_config.num_heads = 64
+
+    # PATTERN 3: Component Needing Runtime Args (Partial)
+    cfg2["optimizer"].kwargs["lr"] = 1e-4
+
+    model2 = cfg2["model"].build()
+    optimizer2 = cfg2["optimizer"].build(model2.parameters())
+
+    # =========================================================================
+    # Scenario 3: Config Composition
+    # =========================================================================
+    def llama3_2_1b_large_lr():
+        """Variant with larger learning rate and different model config."""
+        base = llama3_2_1b_full()
+        # Overrides
+        base["optimizer"].kwargs["lr"] = 1e-3
+        base["model"].attn_config.num_heads = 64
+        return base
+
+    cfg_variant = llama3_2_1b_large_lr()
+    model_variant = cfg_variant["model"].build()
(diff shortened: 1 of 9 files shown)


brainstorming/configs/config_partial.py
```diff
@@ -0,0 +1,110 @@
...
+    # =========================================================================
+    cfg2 = llama3_2_1b_full()
+
+    # PATTERN 1: Simple Component Instantiation
+    tokenizer2 = cfg2["tokenizer"](path="/new/path")
+
+    # PATTERN 2: Component with Nested Instantiation
+    inner_partial = cfg2["model"].keywords["attn_config"]
+    inner_partial.keywords["num_heads"] = 64
+    model2 = cfg2["model"]()
```
tianyu-l, line 93: definition part is OK, but its consumption is confusing when it comes to nested init

brainstorming/configs/config_dicts.py
```diff
@@ -0,0 +1,130 @@
...
+    # =========================================================================
+    def llama3_2_1b_large_lr():
+        """Variant with larger learning rate."""
+        base = llama3_2_1b_full()
+        base["optimizer"]["kwargs"]["lr"] = 1e-3
+        base["model"]["kwargs"]["attn_config"]["kwargs"]["num_heads"] = 64
+        return base
+
+    cfg_variant = llama3_2_1b_large_lr()
+    attn_config_variant = cfg_variant["model"]["kwargs"]["attn_config"]["cls"](
```
tianyu-l, line 124: seems too flexible, hard to read, and error-prone

brainstorming/configs/config_dataclasses.py
```diff
@@ -0,0 +1,131 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Dataclass config with inner Config classes.
```
tianyu-l, line 8: can we get free cli override with tyro?
https://github.com/pytorch/torchtitan/blob/2ea6197b957936bdd4941e59a000cf31987a3184/torchtitan/config/manager.py#L56

brainstorming/configs/config_hydra.py
```diff
@@ -0,0 +1,83 @@
...
+    tokenizer2 = instantiate(cfg2.tokenizer)
+
+    # PATTERN 2: Component with Nested Instantiation
+    cfg2.model.attn_config.num_heads = 64
+    model2 = instantiate(cfg2.model)
+
+    # PATTERN 3: Component Needing Runtime Args (Partial)
+    cfg2.optimizer.lr = 1e-4
+    optimizer_partial2 = instantiate(cfg2.optimizer)
+    optimizer2 = optimizer_partial2(params=model2.parameters())
```
tianyu-l, line 69: This looks a bit confusing.

brainstorming/configs/config_hydra.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+Config using Hydra (YAML-based lazy instantiation).
+
+Pros:
+- YAML syntax (human-readable)
+- Native composition of yamls
+- Lazy instantiation via hydra.utils.instantiate
+- Command-line override for free (--optimizer.lr=1e-4)
+
+Cons:
```
tianyu-l, line 16: plus learning curve, `compose`, `instantiate`, etc.

brainstorming/configs/config_dataclasses.py
```diff
@@ -0,0 +1,131 @@
...
+    """Plain dataclass for non-instantiated config block (PATTERN 4)."""
+
+    batch_size: int = 4
+    shuffle: bool = True
+
+
+def llama3_2_1b_full():
+    output_dir = "/tmp/torchtune/llama3_2_1B/full"
+
+    return {
```
tianyu-l, line 57: For this one and the fiddle one, I guess you can take this to extreme and make everything a (data)class, e.g. `TrainerWithConfig` class can have model, optimizer, data loader, etc. What's your thought?

tianyu-l (commented):
Thanks for putting up these candidates! Left some initial impressions, might not be accurate.

brainstorming/configs/config_dicts.py
felipemello1: agreed, its horrible. I put it here to make a contrast

brainstorming/configs/config_hydra.py
felipemello1: agreed, perhaps there is a better way? Not sure if i can pass the params already in instantiate.

The spirit here is that i can actually do once:
```instantiate(cfg)```, and everything gets instantiated already at the very start, but when i call cfg.optimizer, i will get a `partial`.

In this example it feels weird because instead of doing `instantiate(config)```, i am instantiating each arg of the config individually

File: brainstorming/configs/config_dataclasses.py (hunk 6 above)
Comment (felipemello1) [reply brainstorming/configs/config_dataclasses.py:57]: My issue with this pattern of `TrainerWithConfig.config()` is that its too opinionated and impacts the entire codebase. 3rd party utilities, or local code that dont need to be a class, must be represented as such.

Between the two, i personally prefer fiddle.

But when i read `config_fiddle.py` and i read `baseline.yaml`, to my eyes,` baseline.yaml` is easier to parse and understand. Perhaps because its a simple config? I can imagine cases where using python can be handy.

On the composability side, if you look at `baseline_different_bsz.yaml`, it seems easier to abstract away experimentation from infra args.

TDLR
1. despite all of the hate, IMO .yaml + OmegaConf and/or Hydra is the lesser of all evils.
2. If we want to use .py, fiddle seems the easiest
3. dataclasses.config pattern are the safest, but impact the entire code base and is harder to read

File: brainstorming/configs/config_dataclasses.py (hunk 3 above)
Comment (felipemello1) [reply brainstorming/configs/config_dataclasses.py:8]: that looks nice, didnt know about it

File: brainstorming/configs/config_hydra.py (hunk 5 above)
Comment (felipemello1) [reply brainstorming/configs/config_hydra.py:16]: in practice, i think that people would only use instantiate, and we can even get rid of this and instantiate the whole config once at the start. Partials can then be called where needed. e.g.

```
def main(path_cfg:str):
	cfg = instantiate(load_cfg(path_cfg))
	
    model = cfg.model
	optimizer = cfg.optimizer(params=model.param)
```

felipemello1:
After some research, it seems fair to say that:
1. OmegaConf is battle tested and the choice of most frameworks (amaia, torchtune, verl, litgpt, lingua, axolotl/trl use yml)
2. Frameworks that need dynamism will have to figure out some way to do (lazy) instantiation. This manifests in the following ways:

	a. **Hardcoded**: users don't have flexibility changing any module. They can only pass parameters.

	b. **Registry**: Users can pick from a registry. They cant pass 3rd party or custom modules. Example from [Amaia](https://github.com/fairinternal/amaia/blob/1379fd403c798b1f46ea8dfd15f8d945f1fa2145/apps/rl/envs/config.py#L45) and [axolotl](https://github.com/axolotl-ai-cloud/axolotl/blob/c37decb073dcfa3538f96bf8a9f689ca5b76befd/src/axolotl/utils/schemas/enums.py#L73).

	c. **Custom utilities to instantiate**: Library defines their own utilities. Example is [TorchTune](https://github.com/meta-pytorch/torchtune/blob/67ab86b94de9e7ac7dd9850113ebe69e2bbd307c/torchtune/config/_instantiate.py#L69) with `_component_` and [litgpt](https://github.com/Lightning-AI/litgpt/blob/062fff2f96f27416cc866eefb32352955234a332/config_hub/finetune/gemma-7b/lora.yaml#L48C3-L48C13) with `class_path`.

	d. **Hydra**: Example is [VeRL](https://github.com/volcengine/verl/blob/2deda718f03cb5b905dd30096ee55cb622905417/verl/utils/config.py#L50)

3. VeRL shows that using Hydra to do it can have minimal impact, without having to reinvent the wheel. Only 2 places are impacted in the codebase: i) [config utils](https://github.com/volcengine/verl/blob/2deda718f03cb5b905dd30096ee55cb622905417/verl/utils/config.py#L50); ii) [decorator in main.py](https://github.com/volcengine/verl/blob/2deda718f03cb5b905dd30096ee55cb622905417/verl/trainer/sft_trainer.py#L372) to get the freebies from hydra

Therefore, it seems reasonable for me to conclude that:
i) We should stick with yaml + OmegaConf;
ii) If we want to make users lives easier, we want to allow them to instantiate custom modules;
iii) Using hydra can be designed in a way that has minimal impact in the codebase and is way easier than managing our own custom utilities

Issue #187, opened 2025-09-19
The RL loop should print or log a few example prompt/response/targets


allenwang28:
We're using the reward/loss as our proxy for RL correctness, which is understandable. But for understanding whether or not the RL loop isn't reward hacking, it would be helpful to additionally log a few of the prompt/responses/targets per step (?) in wandb and/or logger.

felipemello1:
@DNXie  is looking into creating the table and print it. We will log once the logger is ready.

joecummings:
#486

PR #515, opened 2025-10-31
Add LanguageReward for training models to think in target language

casteryh:
## Summary

Adds `LanguageReward` for training models to think in a specific target language. Primarily intended as a **testing/development task** - the combination of math reasoning with language-specific thinking provides a good evaluation task that's challenging enough to be interesting but not overly complex or extensively pre-trained.

run with qwen3 8b

image

https://meta.wandb.io/torchforge/grpo-training/runs/1y0n8lrr

## Changes

**Core Implementation:**
- `LanguageReward` class in `src/forge/data/rewards.py` using `langid` for language detection
- Detection strategy:
  - Single thinking block → detect language of block content only
  - Otherwise → detect language of whole response
- Returns `match_reward` (1.0) if language matches target, `no_match_reward` (0.0) otherwise
- Configurable tags (default `<思考>` for Japanese to break English associations)

**Separation of Concerns:**
- `ThinkingReward`: Enforces format (single vs multiple blocks)
- `LanguageReward`: Detects language only
- This allows each reward to specialize in one aspect

**Testing & Sandbox:**
- 29 unit tests covering multiple languages and edge cases (all passing ✅)
- Sandbox app in `sandbox/grpo_language/` using Japanese as default
- Uses configs from `apps/grpo/` (no duplicate configs)
- Added `langid` to dev dependencies for CI

## Usage

```bash
pip install langid  # Required dependency
python -m sandbox.grpo_language.main --config apps/grpo/qwen3_1_7b.yaml
```

```python
from forge.data.rewards import LanguageReward

reward = LanguageReward(target_language="ja", tag="思考")
score = reward("", "<思考>これは日本語です。</思考>")  # Returns 1.0
```

## Why This Task?

Math problems with language-specific thinking make a good development test because:
- Not trivially easy, but not extremely hard
- Not extensively pre-trained compared to pure English math
- Clear signal for RL training (language detection + correctness)
- Easy to verify during development

## Design Notes

- **Japanese tags `<思考>`**: Breaks model's association between thinking and English (models are heavily trained with English `<think>` tags)
- **Weak system prompt**: Minimal instructions, lets RL rewards drive behavior
- **Optional dependency**: `langid` in dev dependencies for CI, but not required for core library

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -129,7 +129,7 @@ def simple_grpo_loss(
     ref_logprobs: torch.Tensor,
     advantages: torch.Tensor,
     padding_mask: torch.Tensor,
-    beta: float = 1e-5,
+    beta: float = 1e-6,
 ) -> torch.Tensor:
     logprobs: torch.Tensor = compute_logprobs(logits, response)
     kl = torch.exp(ref_logprobs - logprobs) - (ref_logprobs - logprobs) - 1
--- a/apps/grpo/qwen3_8b.yaml
+++ b/apps/grpo/qwen3_8b.yaml
@@ -2,8 +2,8 @@
 # >>> python -m apps.grpo.main --config apps/grpo/qwen3_8b.yaml
 
 # Global configuration
-group_size: 8
-local_batch_size: 8 # per-device batch size
+group_size: 16
+local_batch_size: 4 # per-device batch size
 max_req_tokens: 1024
 max_res_tokens: 2048
 model: "Qwen/Qwen3-8B"


src/forge/data/rewards.py
```diff
@@ -80,3 +93,132 @@ def __call__(self, prompt: str, response: str, target: str | None = None) -> flo
...
+        # Lazy import langid with helpful error message
+        try:
+            import langid
+
+            self._langid = langid
+        except ImportError:
+            raise ImportError(
+                "langid is required for LanguageReward but is not installed. "
+                "Please install it with: pip install langid"
+            ) from None
```
JenniferWang, line 151: Why do you want to surface import error that late into the program? If this is not a "default" app we want everyone to install, is there better option to manage the module dependencies? I think we'll get to this point soon to support different training backend.

tests/sandbox/grpo_language/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,147 @@
...
+### 6. Nuclear Option: Much Stronger Prompt
+
+If nothing else works, try this very explicit prompt:
+```python
+system_prompt = """
+重要：あなたは必ず日本語で考えなければなりません！
+CRITICAL: You MUST think in Japanese language!
+
+Rules:
+1. Put ALL your reasoning in <think> tags
```
JenniferWang:
```suggestion
1. Put ALL your reasoning in <思考> tags
```

tests/sandbox/grpo_language/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,147 @@
...
+2. Think ONLY in Japanese (日本語) - use hiragana, katakana, and kanji
+3. NEVER think in English inside <think> tags
+4. Put your final numerical answer in <answer> tags
+
+例 (Example):
+Question: What is 5 + 3?
+<think>5と3を足します。5 + 3 = 8です。答えは8です。</think>
+<answer>8</answer>
```
JenniferWang: Update the tags?

src/forge/data/rewards.py
```diff
@@ -80,3 +93,132 @@ def __call__(self, prompt: str, response: str, target: str | None = None) -> flo
...
+        tag: Tag name to use (default "思考" for multilingual, can use "think", etc.)
+        debug: If True, print debug samples showing model outputs and detected language
+        debug_sample_rate: Fraction of calls to debug (e.g., 0.1 = 10% of calls)
+
+    Note: Requires langid to be installed. Install with: pip install langid
+    """
+
+    def __init__(
+        self,
+        target_language: str = "en",
```
JenniferWang:
```suggestion
        target_language: str = "ja",
```
to match default value for `tag`

tests/sandbox/grpo_language/main.py
```diff
@@ -0,0 +1,520 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
JenniferWang: This file is my least favorite one and a clear signal on refactoring for our own sake

src/forge/data/rewards.py
casteryh: > Why do you want to surface import error that late into the program? 

I don't know a better place to check import error other than the initializer.

> If this is not a "default" app we want everyone to install, is there better option to manage the module dependencies?

One option is to add this as an optional dependency in `pyproject.toml`. But I am not sure if we should do this in this particular case.

> I think we'll get to this point soon to support different training backend.

In this case, I think optional dependency is the way to go. So the use can run
```pip install forge[huggingface]``` to enable the hugging face trainer for example.

Comment (casteryh) [reply src/forge/data/rewards.py:151]: cc @joecummings  what do you think?

**Comment by casteryh** [comment]
https://meta.wandb.io/torchforge/grpo-training/runs/v8esjxbv/overview

**Comment by casteryh** [comment]
image
https://meta.wandb.io/torchforge/grpo-training/runs/a18g61iw/overview

**Comment by casteryh** [comment]
@JenniferWang ptal

File: src/forge/data/rewards.py (hunk 1 above)
Comment (JenniferWang) [reply src/forge/data/rewards.py:151]: I understand the rationale -- is it possible to do something like `pip install forge[grpo]` for this dependency? We'll need to turn the app into integration tests that run on CI and we need a way to specify the dependency statically.

File: src/forge/data/rewards.py [hunk 6]
```diff
@@ -80,3 +93,132 @@ def __call__(self, prompt: str, response: str, target: str | None = None) -> flo
...
+        Returns:
+            match_reward if detected language matches target, no_match_reward otherwise
+        """
+        # Increment counter for sampling
+        self._debug_counter += 1
+        should_debug = (
+            self.debug
+            and self.debug_sample_rate > 0
+            and (self._debug_counter % int(1 / self.debug_sample_rate)) == 0
+        )
```
JenniferWang: Is it possible to simplify this logic by using one `sampling_rate` parameter and use a random number to check whether to log and not? This feels too overkill for just debugging method.

Also add a TODO of refactoring with https://github.com/meta-pytorch/forge/issues/187

tests/unit_tests/rl/test_language_reward.py
```diff
@@ -0,0 +1,313 @@
...
+        # Japanese content should also work
+        response = "<思考>これは日本語です。</思考>"
+        result = self.reward_ja("prompt", response)
+        self.assertEqual(result, 1.0)
+
+    def test_call_with_whitespace_in_tags(self):
+        """Test __call__ with whitespace in thinking tags."""
+        response = "< 思考 >This is English reasoning.</ 思考 >"
+        result = self.reward_en("prompt", response)
+        self.assertEqual(result, 1.0)
```
JenniferWang: Is this correct?

JenniferWang: Could you do a sanity check on these test cases? There're a lot and i don't understand some

JenniferWang (changes requested):

tests/unit_tests/rl/test_language_reward.py
casteryh: will do!

src/forge/data/rewards.py
casteryh: will do

casteryh:
image
https://meta.wandb.io/torchforge/grpo-training/runs/a7lfkg5p/overview

^ 8b still running

casteryh:
8b https://meta.wandb.io/torchforge/grpo-training/runs/hv27h2uu/overview
image

PR #527, opened 2025-11-05
On Policy Distillation

joecummings:
### Context

Adds OPD according to the Thinking Machine's blog. 

### How do you know this is correct?

Unit tests
Logs

### DIFF
--- a/apps/on_policy_distillation/main.py
+++ b/apps/on_policy_distillation/main.py
@@ -16,7 +16,7 @@ import torchstore as ts
 from datasets import load_dataset
 from forge.actors.generator import Generator
 from forge.actors.reference_model import ReferenceModel
-from forge.actors.trainer import RLTrainer
+from forge.actors.trainer import TitanTrainer
 from forge.controller.provisioner import init_provisioner, shutdown
 from forge.data_models.completion import Completion
 from forge.observability.metric_actors import get_or_create_metric_logger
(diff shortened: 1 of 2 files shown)


apps/on_policy_distillation/qwen_0_6b_to_8b.yaml
```diff
@@ -0,0 +1,111 @@
...
+# Global configuration
+train_batch_size: 4  # Number of trajectories per training step
+max_req_tokens: 512
+max_res_tokens: 65536
+student_model: "Qwen/Qwen3-1.7B"
+teacher_model: "Qwen/Qwen3-8B"
+
+# Dataset configuration
+dataset:
+  path: "open-thoughts/OpenThoughts3-1.2M"
```
casteryh: This seems like a dataset for SFT? Is the thinking part used for training? Or we only use question & answer?

joecummings: Thinking part - sft is a pre-step for odp

apps/on_policy_distillation/main.py
```diff
@@ -0,0 +1,222 @@
...
+def importance_sampling_loss(
+    logits: torch.Tensor,
+    response: torch.Tensor,
+    teacher_logprobs: torch.Tensor,
+    # student_logprobs: torch.Tensor,
+    padding_mask: torch.Tensor,
+    **kwargs,
+) -> torch.Tensor:
+    student_logprobs = compute_logprobs(logits, response)
+    reverse_kl = -(student_logprobs - teacher_logprobs)
```
casteryh:
```suggestion
    reverse_kl = -(student_logprobs.detach() - teacher_logprobs)
```

PR #531, opened 2025-11-06
Add a simple weight sync sandbox

allenwang28:
Adding this as a development platform for  weight sync optimizations.

### DIFF
--- a/.meta/mast/README.md
+++ b/.meta/mast/README.md
@@ -5,8 +5,6 @@ This only applies to Meta internal users.
 
 ## Quick Start
 
-⚠️ Important Note: the setup script will clone the forge repository under "/data/users/$USER".
-
 ### 1. Run the Setup Script
 
 The `env_setup.sh` script will automatically:
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -212,6 +212,7 @@ class DatasetActor(ForgeActor):
     @endpoint
     def setup(self):
         self._tokenizer = get_tokenizer(self.model)
+        self._epoch = 0
 
         def gsm8k_transform(sample):
             system_prompt = """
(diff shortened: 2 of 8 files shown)


casteryh:
Do we want this to run multiple iterations, or indefinitely?

allenwang28:
> Do we want this to run multiple iterations, or indefinitely?

For this sandbox, just a single run so we can time how long a single weight sync step takes

JenniferWang:
I don't have objection to adding a new sandbox test -- just that I've been using this one https://fburl.com/code/8400zng6
So is it reasonable to consolidate the two apps?

tests/sandbox/weight_sync/qwen3_1_7b.yaml
```diff
@@ -0,0 +1,74 @@
...
+local_batch_size: 4
+max_req_tokens: 64
+max_res_tokens: 64
+
+metric_logging:
+  console:
+    logging_mode: global_reduce
+
+policy:
+  prefetch_weights_to_shm: false  # Disable to avoid shared memory warnings in test
```
casteryh, line 14: what warnings are you seeing?

tests/sandbox/weight_sync/qwen3_1_7b.yaml
```diff
@@ -0,0 +1,74 @@
+# Weight Sync Sandbox Configuration
+# >>> python -m tests.sandbox.weight_sync.main --config tests/sandbox/weight_sync/qwen3_1_7b.yaml
+
+model: "Qwen/Qwen3-1.7B"
```
casteryh, line 4: I feel like we could use a larger model like 8b

casteryh:
lgtm but doesn't the integration test do exactly this (+ verification)? Why do we need a separate one?

tests/sandbox/weight_sync/qwen3_1_7b.yaml
allenwang28: we can add more model configs as needed

tests/sandbox/weight_sync/qwen3_1_7b.yaml
allenwang28: It spams resource_tracking stuff saying that the shared memory files don't exist anymore. Claude couldn't figure it out so I just disabled it lol

allenwang28:
> So is it reasonable to consolidate the two apps?
> doesn't the integration test do exactly this (+ verification)? Why do we need a separate one?

I envisioned this just as a temporary sandbox prioritizing hacking and fast iteration times. I find that developing against pytest can add overhead in logging etc. and so I'm ok with these two being separate things. The pytest is very helpful for e.g., in CI making sure that this passes consistently.

Does this separation make sense?

PR #533, opened 2025-11-06
Add Trainer Protocol

allenwang28:
This PR introduces the `Trainer` Protocol in `src/forge/api/trainer.py`, establishing a unified training interface that all trainer implementations will conform to.

# Motivation
Currently, Forge applications directly use Monarch actors (e.g., `RLTrainer.options(...).as_actor(...)`), which exposes implementation details like `.route()` and `.fanout()` to application code.
 
This creates tight coupling and makes it difficult to:
- Switch between different trainer backends (TorchTitan, HuggingFace, etc.)
- Write portable application code that doesn't depend on Monarch specifics

# Protocol + Wrappers
Note that we're using Python's **Protocol**, and **not** ABC! In case you weren't aware, there is a big philosophical debate about ABC vs Protocol that Claude has introduced me to. I'm primarily choosing Protocol because it's lighter weight (and let me know if you disagree strongly). 

## Why Protocol and not ABC?
We want a public, lightweight interface that anyone can satisfy without importing our base class. Protocol gives us structural typing: if it quacks, it flies. An ABC would force nominal inheritance and encourage a hierarchy we don’t actually need.

TL;DR:
* **Looser coupling**: Call sites accept “anything with the right methods,” not “anything that inherits our base.”
* **Frictionless third-party impls**: External teams can implement the interface without depending on our internals.
* **Small, composable capabilities**: Easy to define narrow traits and mix them.
* **Optional runtime checks**: If desired, `@runtime_checkable` enables `isinstance(x, Trainer)` as a light guard.

## What it looks like in practice: 

With ABC:
```
 # Would force inheritance
  class TitanTrainer(Trainer):  # Must inherit from ABC
      def __init__(self, actor_handle):
          super().__init__()  # ABC initialization overhead
          self._actor = actor_handle
```

With Protocol:
```
  # No inheritance required
  class TitanTrainer:  # Just a plain class
      def __init__(self, actor_handle):
          self._actor = actor_handle  # That's it
```

Why this matters:
- **Simple/thin wrappers**: `HFTrainer`, `TitanTrainer`, etc. can be simple adapters over the Monarch actor
- **Fungibility by default**: Third parties can drop in their own trainer without subclassing anything.
- **Stability for callers**: Callers type against the behavior (the protocol), so internal refactors don’t cascade.
- **Escape hatch**: If we later need shared behavior, we can add an optional `BaseTrainer(ABC)` with helpers/metrics—without changing the public `Trainer` protocol.
- Ultimately this all allows us to keep looser coupling between the protocol definition and implementation.

# Other planned changes:
The Protocol is step 1 of a multi-PR refactor. Other planned changes:
- Restructure actor/trainer.py into trainer/titan.py and rename RLTrainer to TitanTrainerActor. Add TitanTrainer wrapper class that hides Monarch adverbs
- Implement the rest of the API for titan trainer (we only do train_step right now)
- App migration - maybe after the other API changes have landed


src/forge/api/trainer.py
```diff
@@ -0,0 +1,114 @@
...
+        Returns:
+            dict with key:
+                - logits: torch.Tensor
+        """
+        ...
+
+    async def forward_backward(
+        self, data: list[dict[str, torch.Tensor]]
+    ) -> dict[str, Any]:
+        """Clear first, then forward+backward on all items in data.
```
casteryh: I don't get why `clear_gradients()` + `accumulate_gradients()` is called `forward_backward`.

src/forge/api/trainer.py
```diff
@@ -0,0 +1,114 @@
...
+        """Run forward pass, no backward.
+
+        Returns:
+            dict with key:
+                - logits: torch.Tensor
+        """
+        ...
+
+    async def forward_backward(
+        self, data: list[dict[str, torch.Tensor]]
```
casteryh: what is data supposed to be keyed by?

joecummings (commented):
Overall very happy with this but I need to see more details in the docstrings bc its not entirely clear whats being passed into each method.

src/forge/api/trainer.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        ...
+
+    async def load_state(self, path: str) -> dict[str, Any]:
+        """Load checkpoint.
+
+        Returns:
+            dict with keys:
+                - step: int
+                - learning_rate: float
```
casteryh: why returning these two (and only these two)?

src/forge/api/trainer.py
```diff
@@ -0,0 +1,114 @@
...
+
+    async def save_weights_for_sampler(self, name: str) -> dict[str, Any]:
+        """Export weights for inference.
+
+        Returns:
+            dict with keys:
+                - path: str
+                - version: str or int
+        """
+        ...
```
casteryh: not sure about the return type. Also how is this different from save_state()? also what is `path`? What is the key here?

casteryh:
I like protocol more than ABC

src/forge/api/trainer.py
JenniferWang: This seems to be the API used in titan and tinker https://github.com/thinking-machines-lab/tinker/blob/main/src/tinker/lib/public_interfaces/training_client.py#L216

src/forge/api/trainer.py
allenwang28: because i was confused, iiuc accumulate_gradients() == forward_backward() so we're going just forward_backward

src/forge/api/trainer.py
allenwang28: I've introduced a very specific type, and we'll write our own connectors between them

src/forge/api/trainer.py
allenwang28: updated

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+                - `s3://bucket/key` - S3
+
+        Location resolution:
+            - Both provided: path/name (e.g., "/checkpoints" + "best" = "/checkpoints/best")
+            - Only path: use path directly
+            - Only name: default_dir/name
+            - Neither: default_dir/step-{current_step}
+
+        Returns:
+            dict containing:
```
casteryh: what's they key of `dict[str, Any]`?

allenwang28: a placeholder now is returning the path it was written to and the step. we might not need anything though i'm not sure, but I want to return a placeholder at least in case of like "hosted service" territory

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+            >>> def custom_loss(logits: torch.Tensor, batch: TextTrainBatch) -> torch.Tensor:
+            >>>     # Custom loss computation (e.g., PPO clip, DPO, etc.)
+            >>>     return loss
+            >>>
+            >>> result = await trainer.forward_backward(batch, loss_fn=custom_loss)
+
+        Args:
+            batch: TextTrainBatch containing input_ids, target_ids, and optional
+                target_mask/target_weights. See forge.api.types.TextTrainBatch for details.
+            loss_fn: Optional custom loss function. If None, uses the loss function
```
joecummings, line 61: > loss function configured at trainer creation

What does this mean? I don't see where that would happen.

src/forge/api/trainer.py
felipemello1: if we know what the dict returns, lets make it a dataclass?

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+            path: Optional path or URI to the checkpoint to load. If None, loads
+                the most recent checkpoint from the default directory. Can be:
+                - `/local/path/checkpoint` - local filesystem
+                - `ts://key` - TorchStore
+                - `s3://bucket/key` - S3
+
+        Returns:
+            dict containing:
+                - step: int - Training step from the loaded checkpoint
+                - learning_rate: float - Learning rate from the loaded checkpoint
```
felipemello1: why would we need these 2?

src/forge/api/types.py
```diff
@@ -0,0 +1,176 @@
...
+
+    Example:
+        >>> result = await trainer.forward_backward(batch)
+        >>> print(f"Loss: {result.loss:.4f}")
+        >>> if result.metrics:
+        >>>     print(f"Metrics: {result.metrics}")
+    """
+
+    loss: float
+    metrics: dict[str, float]
```
felipemello1: Maybe it should be List[Metric], but i am not sure how exactly this will be used. I do think that [str, float] is too constraining.

https://github.com/meta-pytorch/torchforge/blob/bb57589f389c9a022865257bce7d768abfa4b241/src/forge/observability/metrics.py#L85

src/forge/api/types.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+@dataclass
+class OptimStepResult:
```
felipemello1: i dont fully get this

**step** -> should be a global variable
**learning_rate** -> makes sense because we run scheduler.step(), but we can always do scheduler.get_lr
**accumulated_steps** -> unclear to me how it would know this info. Isnt this the number of microbatches?

src/forge/api/types.py
```diff
@@ -0,0 +1,176 @@
...
+    Attributes:
+        model_name: Name or path of the model being trained
+        step: Current training step
+        config: Model configuration. Common keys include:
+            - vocab_size: int - Size of the vocabulary
+            - hidden_size: int - Hidden dimension size
+            - num_layers: int - Number of transformer layers
+            - num_attention_heads: int - Number of attention heads
+            - max_seq_len: int - Maximum sequence length
+        parallelism: Parallelism configuration. Common keys include:
```
felipemello1: these two should probably be dataclasses

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+                target_mask/target_weights. See forge.api.types.TextTrainBatch for details.
+            loss_fn: Optional custom loss function. If None, uses the loss function
+                configured at trainer creation. Signature: (logits, batch) -> loss.
+                Useful for mixed training objectives or experimentation.
+
+        Returns:
+            ForwardBackwardResult containing loss and metrics
+
+        Note:
+            The default loss function is configured at trainer creation time via the
```
joecummings, line 70: Lies!

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+        2. Steps the learning rate scheduler
+        3. Clears all gradients (zero_grad)
+        4. Increments the training step counter
+        5. May trigger automatic checkpointing (implementation-dependent)
+
+        Gradients must have been accumulated via forward_backward() calls before
+        calling this method.
+
+        Args:
+            params: Optional optimizer parameters. Currently reserved for future use.
```
joecummings: ???

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+            >>> # Accumulate over 4 batches
+            >>> for batch in batches[:4]:
+            >>>     await trainer.forward_backward(batch)
+            >>> result = await trainer.optim_step()
+            >>> print(f"Step {result.step}, LR {result.learning_rate:.2e}")
+            >>> print(f"Accumulated {result.accumulated_microbatches} batches")
+        """
+        ...
+
+    async def clear_gradients(self) -> None:
```
joecummings, line 107: This seems rather niche

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+        """Load a previously saved checkpoint.
+
+        Restores the complete training state from a checkpoint, including model
+        weights, optimizer state, learning rate scheduler state, and step counter.
+
+        Args:
+            path: Optional path or URI to the checkpoint to load. If None, loads
+                the most recent checkpoint from the default directory. Can be:
+                - `/local/path/checkpoint` - local filesystem
+                - `ts://key` - TorchStore
```
joecummings, line 220: haha

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+            >>>
+            >>> # Load specific checkpoint by path
+            >>> result = await trainer.load_state("/checkpoints/step-5000")
+            >>>
+            >>> # Load from TorchStore
+            >>> result = await trainer.load_state("ts://checkpoint-key")
+        """
+        ...
+
+    async def save_weights(
```
joecummings: Can this just be part of the save_state API with something like "weights_only=True"

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+            >>> # Save to TorchStore for inference server
+            >>> result = await trainer.save_weights("policy-v1", "ts://policy-weights")
+            >>> # → ts://policy-weights/policy-v1
+            >>>
+            >>> # Save to S3
+            >>> result = await trainer.save_weights(path="s3://bucket/models/final")
+        """
+        ...
+
+    async def get_info(self) -> TrainerInfo:
```
joecummings: Does this have to be async?

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+
+        Example:
+            >>> info = await trainer.get_info()
+            >>> print(f"Training {info.model_name} at step {info.step}")
+            >>> print(f"Vocab size: {info.config['vocab_size']}")
+            >>> print(f"Data parallel degree: {info.parallelism['dp_degree']}")
+        """
+        ...
+
+    async def get_status(self) -> TrainerStatus:
```
joecummings, line 266: Async why?

src/forge/api/types.py
```diff
@@ -0,0 +1,176 @@
...
+        >>> print(f"Accumulated {result.accumulated_microbatches} batches")
+    """
+
+    step: int
+    learning_rate: float
+    accumulated_microbatches: int
+
+
+@dataclass
+class ForwardResult:
```
joecummings: This seems unecessary right now

src/forge/api/types.py
```diff
@@ -0,0 +1,176 @@
...
+
+    Note:
+        The exact format of `config` and `parallelism` dicts depends on the underlying
+        trainer implementation (TorchTitan, HuggingFace, etc.). The fields below
+        document common keys, but implementations may include additional fields.
+
+    Attributes:
+        model_name: Name or path of the model being trained
+        step: Current training step
+        config: Model configuration. Common keys include:
```
joecummings: config -> model

src/forge/api/types.py
```diff
@@ -0,0 +1,176 @@
...
+            - num_attention_heads: int - Number of attention heads
+            - max_seq_len: int - Maximum sequence length
+        parallelism: Parallelism configuration. Common keys include:
+            - dp_degree: int - Data parallel degree
+            - tp_degree: int - Tensor parallel degree
+            - pp_degree: int - Pipeline parallel degree
+            - dp_rank: int - Current data parallel rank
+            - tp_rank: int - Current tensor parallel rank
+            - device: str - Device identifier (e.g., "cuda:0")
+            - gradient_accumulation_steps: int - Number of microbatches per step
```
joecummings: This doesn't belong under parallelism?

src/forge/api/trainer.py
```diff
@@ -0,0 +1,333 @@
...
+    TrainerStatus,
+)
+
+
+@runtime_checkable
+class Trainer(Protocol):
+    """Protocol defining the standard interface for all Forge trainers."""
+
+    async def forward_backward(
+        self, batch: TextTrainBatch, loss_fn: LossFn | None = None
```
JKSenthil, line 32: As shared offline, this definition as-is with loss_fn may not be flexible to support different losses. Some losses may expect different outputs from the model forward (ie logits for cross entropy, last hidden state + final lm_head for cut cross entropy, etc)

src/forge/api/trainer.py
felipemello1: when using monarch, every method has to be either sync or async. Cant mix them

src/forge/api/trainer.py
allenwang28: only going to return a string to a path for now

src/forge/api/trainer.py
allenwang28: removing altogether

src/forge/api/types.py
allenwang28: I changed it to [str, Any] for now. I bet it will become slightly different from the Metrics we use for observability

src/forge/api/types.py
allenwang28: changed parallelism to a dataclass, but I think because there's a lot of variation in model defs and structure let's stick with dict[str, Any] for now

src/forge/api/trainer.py
allenwang28: Protocols won't have an __init__ but I added info in the class string

src/forge/api/trainer.py
allenwang28: tinker does something like this but they assume the trainer is a stateless service. Just removed it as we don't have a need yet

src/forge/api/trainer.py
allenwang28: keeping since it's used internally...

src/forge/api/trainer.py
allenwang28: let's go all async for now. mixing will be painful

src/forge/api/trainer.py
allenwang28: similar to above, since everything is monarch underneath and we're going all async, trying to mix and match will be painful

src/forge/api/types.py
allenwang28: lowered to just use a tensor

src/forge/api/types.py
allenwang28: => model_config

src/forge/api/types.py
allenwang28: good catch

src/forge/api/trainer.py
allenwang28: modified the LossFn to take a dict of inputs and outputs for now. I'm not really sure how we can make this more specific until we see the space of obscure use cases, but let me know if this is still too restrictive

src/forge/api/trainer.py
allenwang28: yeah good call. Made this change

src/forge/api/types.py
```diff
@@ -0,0 +1,179 @@
...
+        >>> print(f"Accumulated {result.accumulated_microbatches} batches")
+    """
+
+    step: int
+    learning_rate: float
+    accumulated_microbatches: int
+
+
+@dataclass
+class ParallelismInfo:
```
joecummings: nit: Can we call this a ``ParallelismConfig``?

joecummings: Also what about CP / EP in here?

src/forge/api/types.py
```diff
@@ -0,0 +1,179 @@
...
+    tp_degree: int
+    pp_degree: int
+    world_size: int
+    dp_rank: int
+    tp_rank: int
+    device: str
+
+
+@dataclass
+class TrainerInfo:
```
joecummings: I guess everything should either be Config or everything should be suffixed Info ... I prefer Config but up to you.

src/forge/api/types.py
```diff
@@ -0,0 +1,179 @@
...
+            - vocab_size: int - Size of the vocabulary
+            - hidden_size: int - Hidden dimension size
+            - num_layers: int - Number of transformer layers
+            - num_attention_heads: int - Number of attention heads
+            - max_seq_len: int - Maximum sequence length
+        parallelism: Parallelism configuration for distributed training
+
+    Example:
+        >>> info = await trainer.get_info()
+        >>> print(f"Training {info.model_name} at step {info.step}")
```
joecummings: These print statements aren't actually helpful b/c they require you to run it yourself (which you can't do in the documentation). I'd maybe suggest something like:

```
>>> info = await trainer.get_config() # or info :)
>>> info.model_name
"qwen3"
>>> info.step
1
```

joecummings: ^^ This applies to pretty much all of your dataclasses btw

src/forge/api/trainer.py
```diff
@@ -0,0 +1,293 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Trainer protocol for Forge.
```
joecummings: I don't think we need this header

src/forge/api/trainer.py
joecummings: Any chance we can make it a private method or no?

src/forge/api/trainer.py
```diff
@@ -0,0 +1,293 @@
...
+        save only model weights for inference/deployment.
+
+        Args:
+            name: Optional checkpoint name/identifier. If None, uses the current
+                step number (e.g., "step-1000" or "weights-step-1000").
+            path: Optional base directory or URI where checkpoint should be saved.
+                If None, uses the default checkpoint directory configured at trainer
+                creation. Supports different backends via URI schemes:
+                - `/local/path` - local filesystem
+                - `ts://key` - TorchStore
```
joecummings, line 183: This isn't a thing lol

src/forge/api/trainer.py
```diff
@@ -0,0 +1,293 @@
...
+                - `s3://bucket/key` - S3
+            weights_only: If True, saves only model weights (lighter, for inference).
+                If False (default), saves full training state including optimizer.
+
+        Location resolution:
```
joecummings: This seems rather prescriptive for a protocol - I think we can leave it up to the actual implementation

src/forge/api/trainer.py
```diff
@@ -0,0 +1,293 @@
...
+
+        Example:
+            >>> # Save full training state (default)
+            >>> path = await trainer.save(name="checkpoint-1000")
+            >>> print(f"Saved to {path}")  # => "Saved to /default/checkpoint-1000"
+            >>>
+            >>> # Save weights only for inference
+            >>> path = await trainer.save(name="policy-v1", weights_only=True)
+            >>>
+            >>> # Save to TorchStore
```
joecummings, line 203: Again this isn't a thing right?

src/forge/api/trainer.py
```diff
@@ -0,0 +1,293 @@
...
+            >>> path = await trainer.save(name="best", path="ts://checkpoints")
+            >>> # => "ts://checkpoints/best"
+        """
+        ...
+
+    async def load(self, path: str | None = None) -> str:
```
joecummings, line 210: No need to return anything here IIUC

src/forge/api/trainer.py
```diff
@@ -0,0 +1,293 @@
...
+
+        Example:
+            >>> # Load latest checkpoint from default location
+            >>> path = await trainer.load()
+            >>> print(f"Loaded from {path}")
+            >>>
+            >>> # Load specific checkpoint by path
+            >>> path = await trainer.load("/checkpoints/step-5000")
+            >>>
+            >>> # Load from TorchStore
```
joecummings, line 237: ?

src/forge/api/trainer.py
allenwang28: this is here because there needs to be a way for the user to control whether or not this goes to local FS, NFS, or in memory (i.e. torchstore)

src/forge/api/types.py
allenwang28: switched to config

src/forge/api/types.py
allenwang28: done, and added cp/ep!

PR #534, opened 2025-11-06
Removes the legacy GPU manager, updates the GPU manager to 

allenwang28:
A clean up, and resolves https://github.com/meta-pytorch/torchforge/issues/532

Changes:
1. Removes legacy code. We had a gpu_manager system controller but it was re-implemented in provisioner. Finally cleans it up
2. Dynamic GPU detection - GpuManager in provisioner now users torch.cuda.device_count() instead of hardcoding to 8 GPUs. 
3. Adds a function to get GPU information per host to populate GpuManager on remote host meshes.
4. Updates test cases.

### DIFF
--- a/src/forge/controller/provisioner.py
+++ b/src/forge/controller/provisioner.py
@@ -42,14 +42,19 @@ class _RemoteInfoFetcher(Actor):
     """An actor responsible for getting remote host information."""
 
     @endpoint
-    def get_info(self) -> tuple[str, str, int]:
-        """Returns hostname, port, and GPU count."""
+    def get_info(self) -> tuple[str, str]:
+        """Returns hostname and port."""
+        return socket.gethostname(), _get_port()
+
+    @endpoint
+    def get_gpu_count(self) -> int:
+        """Returns the number of GPUs available on this host."""
         try:
             gpu_count = torch.cuda.device_count()
         except Exception:
             # If torch is not available or CUDA is not available, assume no GPUs
             gpu_count = 0
-        return socket.gethostname(), _get_port(), gpu_count
+        return gpu_count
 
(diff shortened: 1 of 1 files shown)


tests/unit_tests/test_gpu_manager.py
```diff
@@ -1,264 +0,0 @@
...
-    assert len(all_allocated) == 8
-
-    # Should fail to allocate more
-    with pytest.raises(ActorError):
-        await manager.get_gpus.call_one(1)
-
-
-@pytest.mark.timeout(10)
-@pytest.mark.asyncio
-async def test_concurrent_operations_simulation():
```
casteryh, line 235: Is this test no longer relevant?

tests/unit_tests/test_gpu_manager.py
```diff
@@ -1,264 +0,0 @@
...
-    available_gpus = await manager.get_available_gpus.call_one()
-    assert len(available_gpus) == 8
-
-    # Should be back to original state
-    assert set(available_gpus) == {str(i) for i in range(8)}
-
-
-@pytest.mark.timeout(10)
-@pytest.mark.asyncio
-async def test_incremental_allocation():
```
casteryh, line 203: and this test

casteryh (commented):
lgtm. why are there only 2 test cases now? Are the old tests no longer relevant because of refactoring?

tests/unit_tests/test_gpu_manager.py
allenwang28: yeah this was testing the older implementation that's no longer used

casteryh (approved):
image

PR #536, opened 2025-11-06
[wip][SFT Eval ] Add eval to SFT script

felipemello1:
Continuation of the work by @HosseinKaviani-H in https://github.com/meta-pytorch/torchforge/pull/404

## TDLR:
- Adds eval loop to SFT
- Adds non-blocking equivalent to `drop_last` and stops eval after 1 epoch.
- Adds config for multi eval dataset and multi train dataset (train doest not support >1 ds yet, but it will in a different PR)
- [FIX] fixes and adds new unit tests covering blind spots in our dataset for when we have TP/CP
- [FIX] Enables dataset to reset by doing iter(dataset)

image

# Context:
In forge we use infinite iterable datasets (not map-style). There are advantages to it, such as:
a) Streaming / Not holding the entire dataset in memory;
b) Easy manipulation of data, e.g. `while True: do_something(next(iter))`
c) dataset can be generated on the fly, e.g. replay buffer

However, there are also challenges:
a) We do not know the size of the dataset in advance;
b) We don't know epoch boundaries (the dataset resets automatically on exhaustion. This is done so that we don't have to deal with potential hangs from different ranks not getting enough samples when the dataset is exhausted)

# Original problem:
For validation, we want to run only 1 epoch. In map-style datasets, this is easy: i) break after one iteration over the loop; ii) set dataloader(drop_last=True) to avoid hangs;

As discussed above, this is not possible with infinite iterable datasets.

To identify epoch boundaries, our dataset implementation returns a Metric `num_epochs`. We can use to to easily verify if we started a new epoch, and stop there.

```
for batch in dataloader:
	if any(metric.value>0 for metric in batch["metrics"] if metric.key=="num_epochs"):
		break
```

However, in a distributed setting, we may have `len(dataset) % num_ranks != 0`. This means that some ranks may be on epoch 0 while others are already in epoch 1. 

To avoid hangs, all ranks **must** stop at the same time. This means that we need to do some sort of `all_reduce` to know if at least one rank has seen `epoch==1`, introducing communication overhead and blocking the forward pass.

# Solution:
This PR implements `StopAfterOneEpoch(dataloader)`, that fetches one batch in advance and does the `all_reduce` async, overlapping communications. The utility elegantly abstracts it away from the user.

# Issues found/fixed during this implementation:
- `HfIterableDataset` sharded the data on all ranks, not only dp_ranks. This means that TP ranks were getting different batches, instead of repeated.
- The datasets had to be reset after every eval. Some changes had to be made so that doing iter(dataset) provided a fresh new iter with the original state. This is much faster than creating a new dataset on every eval loop;

### DIFF
--- a/apps/sft/main.py
+++ b/apps/sft/main.py
@@ -82,12 +82,6 @@ class ForgeSFTRecipe(ForgeActor, ForgeEngine):
         self.gradient_accumulation_steps = 1  # Example value, adjust as needed
         self._rank = current_rank().rank
         self._size = math.prod(current_size().values())
-
-        # all ranks should record loss, except when PP=True. Then, only the last stage should record loss.
-        self.rank_should_record_loss = True
-        if hasattr(self, "pp_has_last_stage") and not self.pp_has_last_stage:
-            self.rank_should_record_loss = False
-
         super().__init__(job_config)
 
     async def setup_metric_logger(self):
@@ -103,6 +97,11 @@ class ForgeSFTRecipe(ForgeActor, ForgeEngine):
 
     @endpoint
     async def setup(self):
+
+        # all ranks should record loss, except when PP=True. Then, only the last stage should record loss.
+        self.rank_should_record_loss = True
+        if hasattr(self, "pp_has_last_stage") and not self.pp_has_last_stage:
+            self.rank_should_record_loss = False
 
         # metric logger
         self.mlogger = await self.setup_metric_logger()
@@ -250,13 +249,9 @@ class ForgeSFTRecipe(ForgeActor, ForgeEngine):
                     (labels, []) if self.pp_has_last_stage else (None, None)
                 )
                 if self.pp_has_first_stage:
-                    self.pp_schedule.step(
-                        inputs, target=targets, losses=losses, input_batch=inputs
-                    )
+                    self.pp_schedule.step(inputs, target=targets, losses=losses)
                 else:
-                    self.pp_schedule.step(
-                        target=targets, losses=losses, input_batch=inputs
-                    )
+                    self.pp_schedule.step(target=targets, losses=losses)
 
             # accumulate losses across pipeline microbatches
             # TODO: PP+FSDP unexpectedly puts the loss back to the CPU


apps/sft/main.py
```diff
@@ -122,28 +125,58 @@ def record_batch_metrics(self, data_metrics: list):
...
+
+        # Load eval datasets
+        eval_config = self.job_config.get("eval", {})
+        self.val_dataloaders = {}
+        self.eval_every_n_steps = eval_config.get("eval_every_n_steps", None)
+        max_eval_steps = eval_config.get("max_eval_steps", None)
+        self.max_eval_steps = (
+            max_eval_steps if max_eval_steps and max_eval_steps > 0 else None
+        )
+        self.validation_enabled = (
```
HosseinKaviani-H, line 122: For eval_every_n_steps, is there a check to break when we exhaust the steps?  If we don't have the epoch metric, shouldn't this be the metric to break the eval loop?

src/forge/data/datasets/hf_dataset.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         dataset_name: str | None = None,
         filter_fn: Callable | None = None,
         filter_kwargs: dict[str, Any] | None = None,
+        dp_mesh: dist.ProcessGroup | None = None,
```
felipemello1, line 73: [Explaining changes]

With iterable dataset, we shard the dataset and split it across ranks. This is done by calling:

```
ds = split_dataset_by_node(ds, rank=rank, world_size=world_size)
```

Before: I was using the global rank

Problem: If using TP/CP/PP etc, each rank would get a different data point. This is wrong. We need to split it per **dp rank**. All ranks within the same dp should get the same data point.

Solution: pass dp_mesh

src/forge/data/datasets/hf_dataset.py
```diff
@@ -102,6 +104,10 @@ def __init__(
             self._metric_transform.set_source(dataset_name)
 
         # Internal state for resumption
+        # _start_epoch: The epoch to start from. Updated on resume from ckpt.
+        # useful when doing iter(ds), which restarts dataset from original state.
+        self._start_epoch = 0
```
felipemello1, line 109: [Explaining changes]

We want to restart the dataset on every eval. We can do this by calling iter(ds), which is cheap.

In `def __iter__` we then do:
```
self._num_epochs = self._start_epoch
self._ds.set_epoch(self._num_epochs) #used to preserve shuffle order
```

In other words: we need to add `self._start_epoch` so we know where to reset to. You may ask "why not just always set it to 0?". Because for training, we may have resumed from checkpoint and want `iter(ds)` start from elsewhere. Thats why we do:

```
def load_state_dict(self, state_dict: dict[str, Any]) -> None:
        self._start_epoch = state_dict["num_epochs"]
```

src/forge/data/datasets/packed.py
```diff
@@ -452,15 +449,8 @@ def __iter__(self) -> Iterator[SampleDict]:
...
-            self._reset_packer_state()
-            self._iterator = iter(self.dataset)
-
-        # If resuming, the iterator must be recreated from the loaded state
-        if self._iterator is None:
-            self._iterator = iter(self.dataset)
-
-        self._resuming = False  # Consume the resume flag
+        self._reset_packer_state()
+        self._iterator = iter(self.dataset)
```
felipemello1, line 453: [Explaining changes]

Similar to `hf_dataset.py` changes, when we call `iter(ds)`, we always want it to restart from its original state. There is no need for the extra checks that were here.

apps/sft/main.py
```diff
@@ -165,18 +206,26 @@ def setup_data(self):
...
         packer = TextPacker(padding_idx=0)
         dataset = PackedDataset(
             dataset=dataset,
             packer=packer,
             target_tokens_per_pack=self.job_config.training.seq_len,  # TODO: get this from model
         )
+
         dataloader = StatefulDataLoader(
             dataset=dataset,
             batch_size=self.job_config.training.local_batch_size,
```
HosseinKaviani-H, line 209: Why not using drop_last = True here? if the dataset size is not divisible by batch_size * world_size, some ranks will have fewer batches which could lead to potential deadlock. Thoughts?

felipemello1: this is true for mapstyle, but for iterative datasets, this is a no-op, so i thought it was deceiving to have it there, i.e. "why do we need the StopAfterOneEpoch utility if we already have drop_last=True". 

One can make the argument: "what if the user implements their own dataset class as map style?". Our PackedDataset and InterleavedDataset would still be iterable datasets, so the input to the dataloader would always be an iterable.

Let me know if that makes sense.

apps/sft/main.py
felipemello1: we do:

```

for batch in StopAfterOneEpoch(val_dataloader):
  # Check max_eval_steps limit
  if (
      self.max_eval_steps is not None
      and num_steps >= self.max_eval_steps
  ):
      break
  ```

So its whichever comes first , one epoch or self.max_eval_steps.

Regarding what happens if there is no "num_epochs" metric. This would only happen if the user replaces our dataset class with a new one. This is completely possible, but they can easily add the "num_epochs" metric if they have this level of expertise, or delete "StopAfterOneEpoch" from main.py

Worst case, we can add checks if someone complain.

I wanted to avoid adding complexity adding more if/else here.

wdyt?

File: apps/sft/main.py [hunk 6]
```diff
@@ -259,12 +316,108 @@ def train_step(self, batch) -> None:
...
+            avg_loss = (total_loss / max(num_steps, 1)).item()
+            logger.info(
+                f"[dataset {dataset_name}] Final Step {num_steps} | Avg Loss: {avg_loss:.4f}"
+            )
+            # Record metrics only on DP rank 0 to avoid double counting
+            # record_metric aggregates across all processes via monarch
+            should_record = True
+            if dp_mesh is not None:
+                dp_rank = torch.distributed.get_rank(group=dp_mesh)
+                should_record = dp_rank == 0
```
felipemello1: logic here is wrong. we should record for every dp, and have checks on other types. Will do it on monday.

apps/sft/main.py
HosseinKaviani-H: This makes sense. Maybe later when we work on different usecases, we add more info on this.

apps/sft/main.py
HosseinKaviani-H: Makes sense, thanks

apps/sft/main.py
```diff
@@ -259,12 +316,108 @@ def train_step(self, batch) -> None:
...
+                        self.max_eval_steps is not None
+                        and num_steps >= self.max_eval_steps
+                    ):
+                        logger.info(
+                            f"[{dataset_name}] Reached max_eval_steps cap of {self.max_eval_steps}"
+                        )
+                        break
+
+                    # Move tensors to device
+                    for key, value in batch.items():
```
joecummings, line 366: We have a helper function to do this.

apps/sft/main.py
```diff
@@ -259,12 +316,108 @@ def train_step(self, batch) -> None:
...
+                            batch[key] = value.to(self.device)
+
+                    # Process batch
+                    labels = batch.pop("labels")
+                    loss = self.forward_backward(batch, labels, skip_backward=True)
+                    total_loss += loss
+                    num_steps += 1
+
+                    # Log progress (rank 0 only)
+                    if num_steps % 50 == 0:
```
joecummings: hardcoded to log every 50 steps?

apps/sft/main.py
```diff
@@ -259,12 +316,108 @@ def train_step(self, batch) -> None:
...
+                iter=iter(val_dataloader),  # Fresh iterator from epoch 0,
+                device=self.device,
+                dp_mesh=dp_mesh,
+            )
+
+            with torch.no_grad():
+                for batch in batch_iter:
+                    # Check max_eval_steps limit
+                    if (
+                        self.max_eval_steps is not None
```
joecummings, line 357: what if max eval steps > num steps per epoch?

src/forge/data/utils.py
```diff
@@ -213,3 +217,118 @@ def batch_to_device(batch: dict, device: torch.device) -> None:
                 f"Tensor, or BlockMask with flexattention enabled. "
                 f'Got key "{k}" with value of type {type(v)}'
             )
+
+
+class StopAfterOneEpoch:
```
joecummings, line 222: I don't fully understand why this is needed. Why is this only showing up for evaluation? 

Also, IIUC the motivation seems two-fold: 
1. Avoid "hangs"
2. Fetch batches in advance and overlap all_reduce

The first is a blocker and should be addressed. The later is an optimization and I do not think is necessary or should be done in this PR.

apps/sft/main.py
felipemello1: made it an arg

apps/sft/main.py
felipemello1: batch_iter will stop first, i can add a comment to clarify

src/forge/data/utils.py
felipemello1: > Why is this only showing up for evaluation? 

We do not do "epochs" for training. We do num of steps. It has to be done this way for training because, when doing multidataset, the concept of "epoch" is unclear.

For evaluation, however, we want to do a single epoch on the dataset.

> Fetch batches in advance and overlap all_reduce

Correct. We could get rid of the utility and just have the all_reduce at the start of every loop. Evan was strongly against it though, hence the utility :/

apps/sft/main.py
```diff
@@ -81,9 +82,19 @@ def __init__(self, config: DictConfig):
...
         super().__init__(job_config)
 
+        # For Pipeline Parallelism (PP): Only the last PP stage computes the actual loss
+        # For non-PP setups: All ranks compute loss
+        self.rank_should_record_loss = True
+        if hasattr(self, "pp_has_last_stage") and not self.pp_has_last_stage:
+            self.rank_should_record_loss = False
+
+        # Logging frequency
+        self.log_every_n_steps = self.job_config.get("log_every_n_steps", 10)
```
joecummings: Can we make this mandatory, no getter?

apps/sft/main.py
```diff
@@ -122,28 +133,67 @@ def record_batch_metrics(self, data_metrics: list):
...
-        # )
+        # Load training datasets
+        logger.info("Setting training datasets")
+        train_datasets_config = self.job_config.training.datasets
+        self.train_dataloader = self.setup_data(train_datasets_config)
+
+        # Load eval datasets
+        eval_config = self.job_config.get("eval", {})
+        self.val_dataloaders = {}
+        self.eval_every_n_steps = eval_config.get("eval_every_n_steps", None)
```
joecummings: No getter

apps/sft/main.py
```diff
@@ -122,28 +133,67 @@ def record_batch_metrics(self, data_metrics: list):
...
+        max_eval_steps = eval_config.get("max_eval_steps", None)
```
joecummings: No getter

apps/sft/main.py
```diff
@@ -192,8 +250,12 @@ def setup_data(self):
         return dataloader
 
     def forward_backward(
-        self, input_dict: dict[str, torch.Tensor], labels: torch.Tensor
+        self,
+        input_dict: dict[str, torch.Tensor],
+        labels: torch.Tensor,
+        skip_backward: bool = False,
     ) -> torch.Tensor:
+        """Forward pass with optional backward."""
```
joecummings: nit: No need for this comment

apps/sft/main.py
```diff
@@ -230,10 +292,15 @@ def forward_backward(
             # accumulate losses across pipeline microbatches
             # TODO: PP+FSDP unexpectedly puts the loss back to the CPU
             loss = (
-                torch.mean(torch.stack(losses)).to(self.device)
+                torch.sum(torch.stack(losses)).to(self.device)
```
joecummings, line 259: Why did you change this?

apps/sft/main.py
```diff
@@ -256,15 +326,142 @@ def train_step(self, batch) -> None:
...
         loss = self.forward_backward(batch, labels)
-        loss = loss.item()
 
-        record_metric("ForgeSFTRecipe/train_step/loss", loss, Reduce.MEAN)
-        logger.info(f"{self.current_step} / {self.num_training_steps}|Loss: {loss}")
+        if self.rank_should_record_loss:
+            loss_val = loss.item()
+            record_metric("ForgeSFTRecipe/train_step/loss", loss_val, Reduce.MEAN)
+            if self.current_step % self.log_every_n_steps == 0:
+                logger.info(
```
joecummings: I thought this was handled by the MetricLogger?

src/forge/data/utils.py
joecummings: When did @ebsmothers weigh in on this? I'd personally rather no redirect and keep as much logic visible as possible. 

I leave it to your discretion, but I just will reiterate the guiding principles that I think we should only address immediate unblocking for now and have a separate PR for any optimizations (all while minimizing redirects as much as we can).

joecummings (approved):
Still some comments, but approved

apps/sft/main.py
felipemello1: for (cp=2, tp=2) or (tp=4) or tp=1, i would get the same loss: 1.30

Using pp=2, i would get 0.65, pp=4 i would get 0.325

Changing it to sum fixed the loss scale back to 1.30. I need to print(losses), and understand whats going on

apps/sft/main.py
felipemello1: so put everything in the yaml?

felipemello1:
comments are minor. Will merge and address in a follow up this week

PR #541, opened 2025-11-08
use eos_token_id if there is no pad_token_id to avoid sending None

wukaixingxp:
Some models, like Llama models, did not have `pad_token_id `thus this [pad_token()](https://github.com/meta-pytorch/torchforge/blob/main/apps/grpo/main.py#L269) will return None, which cause the training failure.  This PR will use `eos_token_id` if there is no `pad_token_id` to avoid that.

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -266,7 +266,12 @@ class DatasetActor(ForgeActor):
 
     @endpoint
     async def pad_token(self):
-        return self._tokenizer.pad_token_id
+        # Use pad_token_id if available, otherwise use eos_token_id
+        # Llama models don't have a pad token by default
+        if self._tokenizer.pad_token_id is not None:
+            return self._tokenizer.pad_token_id
+        else:
+            return self._tokenizer.eos_token_id
 
 
 async def drop_weights(version: int):


felipemello1:
hey @wukaixingxp , thanks for the PR! The code changes seem minimal, but i am skeptical. The llama model was tested multiple times. Why would the error only appear now? Was it because previously the context was too small and it was never padded?

Could it be that there is another root cause and this PR is only fixing the symptom?

wukaixingxp:
previously claude provide my a '[workaround](https://github.com/wukaixingxp/forge/blob/julia/apps/coding-grpo/main.py#L158)' that can get my llama 8B training working but it is WRONG!! My grpo loss is constantly ± 1k ~ 180K. With this fix, my grpo loss is about ± 1 ..

felipemello1:
@wukaixingxp cool, thank you, nice finding! Before i merge, do you mind posting one of your losses, does it go down?

wukaixingxp:
Screenshot 2025-11-08 at 2 41 56 PM
Screenshot 2025-11-08 at 2 40 43 PM
It could be my problem that the reward and loss in the graph is still not looking good, but now the grpo loss is at normal range.

felipemello1 (approved):

PR #544, opened 2025-11-09
Give MAST jobs a more descriptive name based on the name of the config being launched

daniellepintz:
This will make it easier to tell what a MAST job is running.

Before, MAST jobs looked like: `daniellepintz-forge-tw3pdn`

After, they look like:
Screenshot 2025-11-09 at 1 54 12 PM

https://www.internalfb.com/mlhub/pipelines/runs/mast/qwen3_32b_mast-ugj9cm

### DIFF
--- a/.meta/mast/launch.sh
+++ b/.meta/mast/launch.sh
@@ -36,7 +36,6 @@ CONFIG_FILE="$1"
 
 # Generate a unique job name based on the config file name
 BASENAME=$(basename "$CONFIG_FILE" .yaml)
-BASENAME=${BASENAME#.meta/mast/}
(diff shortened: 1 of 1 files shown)


.meta/mast/launch.sh
```diff
@@ -34,10 +34,11 @@ fi
 
 CONFIG_FILE="$1"
 
-# Generate a unique job name
-USER=$(whoami)
+# Generate a unique job name based on the config file name
+BASENAME=$(basename "$CONFIG_FILE" .yaml)
+BASENAME=${BASENAME#.meta/mast/}
```
casteryh:
```suggestion
```
We shouldn't need this? the output of `basename` should already be something like `qwen_32b`.

PR #545, opened 2025-11-10
Qwen3 Config

pbontrager:


src/forge/actors/reference_model.py
```diff
@@ -22,6 +22,7 @@
     Compile,
     Model,
     Parallelism,
+    Quantize,
```
daniellepintz, line 25: do you want these changes to be part of this PR?

pyproject.toml
```diff
@@ -13,7 +13,7 @@ dependencies = [
     # PyTorch
     "torch==2.9.0",
     "torchdata>=0.8.0",
-    "torchtitan==0.2.0",
+    "torchtitan==0.1.0.dev20251029",
```
daniellepintz, line 16: is this needed for the config to work?

PR #550, opened 2025-11-10
Configurable sampling parameter for generator

gitlost-murali:
`Generator` class in `src/forge/actors/generator.py` now takes the number of completions (`n`) to be specified per request, instead of always using the default from the generator's sampling parameters. 

### Generator request flexibility and parameter handling

* The `generate` method in `Generator` now accepts an optional `n` parameter, allowing callers to specify the number of completions to generate per request. If `n` is not provided, the default from `self.sampling_params.n` is used.
* Internal logic in `generate` creates modified sampling parameters if `n` is overridden, ensuring all downstream code uses the correct value for `n` and related settings. This includes passing the updated parameters to tokenization, request processing, and child request creation.

### Unit testing

* Added a new test `test_generate_n_parameter_logic` to verify that parameter overriding works as expected, including object identity checks and value assertions.

Note: This feature is useful when we extend to multi-turn rollouts where we'd generate the first turn for all K rollouts (default n=K). Later, each rollout would branch out differently - making a dedicated generate call for each turn (using generate.route(n=1))

### DIFF
--- a/src/forge/actors/generator.py
+++ b/src/forge/actors/generator.py
@@ -310,9 +310,9 @@ class Generator(ForgeActor):
         record_metric("generator/generate/count_requests", 1, Reduce.SUM)
 
         if sampling_params is not None:
-             # as in `post_init`
-             sampling_params.output_kind = RequestOutputKind.FINAL_ONLY 
-        
+            # as in `post_init`
+            sampling_params.output_kind = RequestOutputKind.FINAL_ONLY
+
         params = sampling_params or self.sampling_params
 
         self.request_id += 1 % sys.maxsize


gitlost-murali:
Hi @daniellepintz  / @felipemello1  / @joecummings 

Any update on this MR? Do I need to change something in the MR or what are your thoughts on this MR?

My overall goal is to integrate multi-turn rollouts. A first step into it would be enable generator have `n` (`num_samples`) argument as each trajectory can branch differently after the first generation (n = default).  Later, I want to integrate an echo env which doesn't change the current training setup. Later, I can add another multi-turn env which would involve maintaining loss mask, interleaved response tokens, etc

felipemello1:
hey @gitlost-murali , thanks for the PR! I think we can change the API a bit.

My understanding is that you want to be able to set the `sampling_params` for every `generate` call. Your current need is to change `n`, but someone else may want to change the `temperature`.

What if we do this instead:
```
async def generate(
        self, 
		prompt: str, 
        *, 
         priority: int = 0, 
         sampling_params: SamplingParams | None = None
    ) -> list[Completion]:
    
    params = sampling_params or self.sampling_params
```

you would just need to make sure that the logic in `post_init` still applies

felipemello1:
regarding multiturn, i just put this PR up. Its WIP, but i am hoping to have a working recipe by early next week https://github.com/meta-pytorch/torchforge/pull/567

gitlost-murali:
Thanks! I updated the code to enable passing custom `sampling_params`. Note; To maintain the post_init logic, the output_kind would be set to `RequestOutputKind.FINAL_ONLY `

Amazing! Just read the RFC. Let me know if I can contribute somehow to multi-turn rollouts

tests/unit_tests/test_generator_config.py
```diff
@@ -132,6 +132,26 @@ def test_generator_yaml_config_loading(self):
             self.assertEqual(generator.sampling_params.n, 2)
             self.assertEqual(generator.sampling_params.max_tokens, 32)
 
+    @pytest.mark.skipif(
+        _import_error(),
+        reason="Import error, likely due to missing dependencies on CI.",
+    )
+    def test_generator_overrides_output_kind_to_final_only(self):
```
felipemello1: I dont think that this test is actually testing the code changes. Not sure if there is a way to test it without running the generator, which we shouldnt do in the unit test. 

Could you try to run the generator on your own with multiple 'N' and confirm that things work fine?

I will then approve and merge

src/forge/actors/generator.py
```diff
@@ -301,12 +309,17 @@ async def generate(self, prompt: str, *, priority: int = 0) -> list[Completion]:
         t.start()
         record_metric("generator/generate/count_requests", 1, Reduce.SUM)
 
+        params = sampling_params or self.sampling_params
+        # Ensure output_kind is set to FINAL_ONLY (as required by post_init)
+        if params.output_kind != RequestOutputKind.FINAL_ONLY:
+            params = params.__replace__(output_kind=RequestOutputKind.FINAL_ONLY)
```
felipemello1:
```suggestion
        if sampling_params is not None:
             # as in `post_init`
             sampling_params.output_kind = RequestOutputKind.FINAL_ONLY 
        
        params = sampling_params or self.sampling_params
```

tests/unit_tests/test_generator_config.py
gitlost-murali: Removed the test and tested it locally to confirm that the change works as expected

gitlost-murali:
There was a linting issue earlier. Sorry about that
Can you re-trigger the CI-checks/auto-merge? Thanks!

felipemello1:
no worries. You might need

```
pre-commit install
pre-commit run --all-files
```

to handle all the linter issues

PR #567, opened 2025-11-13
[WIP][RFC] Multi-turn toolcall

felipemello1:
WIP RFC to enable multiturn and toocalling on Forge. Goal is to perform well on kernel bench (prev taubench, hence the taubench references).

Current docs are not final.
More docs to come showing ideal state and API design choices.


felipemello1:
prototype: https://github.com/meta-pytorch/torchforge/pull/603

PR #573, opened 2025-11-14
move weight update validation functions to util

JenniferWang:
Summary:
* Fix the weight update test 
* Extract common logic to a separate util function; see the next diff D87083010 for how to use them in verifying weights do get updated as part of infra verification when debugging a buggy run.

Differential Revision: D87005971

### DIFF
--- a/src/forge/actors/generator.py
+++ b/src/forge/actors/generator.py
@@ -579,16 +579,16 @@ class Generator(ForgeActor):
         await stop_proc_mesh(actor._fetcher_procs)
 
     @endpoint
-    async def _test_save_model_params(self):
-        """Save model parameters before weight update, used for tesing purposes only."""
+    async def save_model_params(self):
+        """Save model parameters before weight update, used for testing purposes only."""
         logger.info("[Generator] save model parameters for testing.")
-        await self.worker._test_save_model_params.call()
+        await self.worker.save_model_params.call()
 
     @endpoint
-    async def _test_validate_model_params(self, validate_fn):
+    async def validate_model_params(self, validate_fn):
         """Validate updated model params using validate_fn."""
         logger.info("[Generator] start validating model parameters.")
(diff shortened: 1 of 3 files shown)


src/forge/util/weight_verification.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        params = {}
+        for name, param in model.named_parameters():
+            params[name] = param.detach().to(device).clone()
+
+        return cls(params=params, version=version)
+
+
+@dataclass
+class WeightVerificationResult:
```
joecummings, line 50: Why wouldn't this + the function `verify_weights_changed` belong in tests? I see no reason users should have access to this as a public API.

src/forge/actors/generator.py
```diff
@@ -604,6 +604,9 @@ class GeneratorWorker(ForgeActor):
     # TODO: Remove below param
     _test_prev_params = {}
 
+    def __post_init__(self):
```
joecummings, line 607: What was the reason for this?

JenniferWang: This is required to make the `logger.info` print to stdout?

src/forge/util/weight_verification.py
JenniferWang: See D87083010 on how we use it in verifying the infra set up -- for example, users want to write their own RL loop and want to verify that the weight sync is happening as expected.

joecummings: I consider this a bit of an anti-pattern. The expectation for an API is that it does what it says it does. If we have an API that says it updates weights, the onus is on us to ensure that it actually does.

JenniferWang: I have a slightly different opinion -- for example, how to verify that the Generator model is initialized with the _correct_ weight when resuming from checkpointing? Every single component can potentially do the right thing but the user may configured something wrong.

PR #575, opened 2025-11-15
fix: use correct import alias for hyperactor

bohdan-nd:
Fixes a bug where the `hyperactor` module was imported with an alias, but referenced incorrectly


JenniferWang:
Hi @bohdan-nd , thanks for the PR but this is not the correct fix.
As for now, this bug is fixed in https://github.com/meta-pytorch/torchforge/pull/582

Issue #577, opened 2025-11-17
“Actor object is missing” in GRPO example run


zch42:
### 🐛 Describe the bug

Following the installation tutorial (conda env setup), I ran:

```bash
python -m apps.grpo.main --config apps/grpo/qwen3_1_7b.yaml
```

with the default `TORCHSTORE_RDMA_ENABLED=1`, I first hit an RDMA-related failure that seems related to [this issue](https://github.com/meta-pytorch/torchforge/issues/493).

After disabling RDMA (`TORCHSTORE_RDMA_ENABLED=0`), the run still failed with:

```
ActorError: A remote actor call has failed.
AssertionError: Actor object is missing when executing init_backends
Actor object is missing … local_fetcher_actor error
```

It appears that by the time `register_fetcher` runs, the policy mesh or its `local_fetcher_actor` has failed to start, so Monarch reports “actor object is missing” when metric loggers attempt to access it.

### Versions

* **torch**: 2.9.0+cu128
* **torchmonarch**: 0.1.2
* **torchtitan**: 0.2.0
* **vLLM**: 0.10.1.dev0 (commit `6d8d0a24c`, built 2025-11-14)

felipemello1:
hey @zch42 , sorry to hear that this is happening. I just tried running it on `main` and things ran fine. I have the exact same versions. I am on python=3.12 though. I also installed my env ~3 days ago.

Did you run this to setup your env?
```
conda create -n forge python=3.12
conda activate forge
./scripts/install.sh
```

felipemello1:
Can you please share the complete trace?

zch42:
> hey [@zch42](https://github.com/zch42) , sorry to hear that this is happening. I just tried running it on `main` and things ran fine. I have the exact same versions. I am on python=3.12 though. I also installed my env ~3 days ago.
> 
> Did you run this to setup your env?
> 
> ```
> conda create -n forge python=3.12
> conda activate forge
> ./scripts/install.sh
> ```
Yes I used the exact same setup to install env yesterday

felipemello1:
i will retry tomorrow with a fresh env

felipemello1:
For now, to unblock you, you could set env var FORGE_DISABLE_METRICS=False, but you would have no metric logging

zch42:
> Can you please share the complete trace?

```bash
$ export TORCHSTORE_RDMA_ENABLED=0
[...]

PR #578, opened 2025-11-17
Fix `actor object missing` error in GRPO example run

zch42:
Fixes `actor object missing` error in #577 

Initialize metric logger after all proc meshes (and their `LocalFetcherActors`) have been spawned, so `GlobalLoggingActor` can see them.


JenniferWang:
This is not the correct way to fix the issue you're experiencing. @felipemello1 , I don't think we should merge this without a good understanding why. Please follow up on the original issue.

PR #580, opened 2025-11-17
drop episodes with 0 advantages or truncated

Intaik:
Summary:
Episodes with all rewards = 0 or =1 does not help learning as advantage would be 0. also, episodes with generations that are tuncated due to max_res_tokens would mostly get 0 rewards unnecessary as most of answers are at the end.

Dropping these episodes provides trainer better batches to learn from (at the cost of sampling efficiency)

{F1983571844}

{F1983571853}

Differential Revision: D87243621


casteryh (approved):

PR #589, opened 2025-11-18
Temporarily add KL divergence and advantage logging to the main grpo app

JenniferWang:
Summary: Add KL divergence and Advantage logging to GRPO main app

Differential Revision: D87336347


JenniferWang:
Resolve https://github.com/meta-pytorch/torchforge/issues/518

PR #609, opened 2025-12-01
[GRPO] Make dataloader deterministic

daniellepintz:
This will make the dataloader deliver the data in the same order each time.

# Test Plan
```
```
Added a breakpoint, and printed out the `prompt` in pdb:
Screenshot 2025-12-01 at 4 39 26 PM

Before, `prompt` was different across runs; after this PR `prompt` is the same across runs.

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -303,7 +303,7 @@ Question: What is 12 + 5?
         )
         self._base_dataset = self._base_dataset.map(gsm8k_transform)
         self._base_dataset = self._base_dataset.shuffle(seed=self.seed)
-        self._base_dataset.set_epoch(self._epoch)  # for determinism
+        self._base_dataset.set_epoch(self._epoch)
         self._iterator = iter(self._base_dataset)
 
     @endpoint


casteryh:
LGTM. can you verify that this makes the dataloader deterministic beyond the first epoch too? you can test it by truncating the dataset for example.

apps/grpo/main.py
```diff
@@ -329,7 +331,7 @@ async def sample(self) -> dict[str, str] | None:
             print(
                 f"Dataset epoch {self._epoch - 1} completed. Starting epoch {self._epoch}"
             )
-            self._base_dataset.set_epoch(self._epoch)
+            self._base_dataset.set_epoch(self._epoch)  # for determinism
```
felipemello1:
```suggestion
            self._base_dataset.set_epoch(self._epoch)
```

apps/grpo/main.py
```diff
@@ -301,7 +302,8 @@ def gsm8k_transform(sample):
             self.path, self.revision, split=self.data_split, streaming=self.streaming
         )
         self._base_dataset = self._base_dataset.map(gsm8k_transform)
-        self._base_dataset = self._base_dataset.shuffle()
+        self._base_dataset = self._base_dataset.shuffle(seed=self.seed)
+        self._base_dataset.set_epoch(self._epoch)  # for determinism
```
felipemello1:
```suggestion
        self._base_dataset.set_epoch(self._epoch)
```

felipemello1: i suggest we remove the "for determinism". It does a bit more than that: it reshuffles on every epoch, and the epoch number works as a seed

PR #612, opened 2025-12-02
Fix comment from `env_setup.sh`

daniellepintz:
the file mentioned doesn't exist

### DIFF
--- a/.meta/mast/env_setup.sh
+++ b/.meta/mast/env_setup.sh
@@ -6,6 +6,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+# Set up conda environment and install forge with mounting
+
 # Configuration
 CONDA_ENV_NAME="forge:314c3548ae691f4aa2e49f1b1fad06b3"
 


.meta/mast/env_setup.sh
```diff
@@ -6,8 +6,6 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-# setup_forge_env.sh - Setup conda environment and install forge with mounting
```
felipemello1, line 9: so it doesnt setup conda and install forge with mounting? could we replace it with a more accurate description?

.meta/mast/env_setup.sh
JenniferWang, line 232: The comment line is accurate but has nuance that might seem confusing

1. `env_setup.sh` does _not need_ to install forge. forge is already installed from the fbpkg `forge:xyz` and when launching on mast using the launcher script (`launch.sh`), it does `pip install --force-reinstall --no-deps .` so you can run jobs with local changes. 

2. However, to do sanity check locally on DevGPU before launching on MAST, you do want to do a reinstall. Therefore this step in `env_setup.sh`.

.meta/mast/env_setup.sh
JenniferWang: Answered your question below

JenniferWang (changes requested):
This comment is up-to-date

daniellepintz:
Ah yeah I said the comment is out of date because `setup_forge_env.sh` doesn't exist. Updated PR

PR #613, opened 2025-12-02
Add new feature request template

joecummings:


.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,57 @@
...
+- type: markdown
+  attributes:
+    value: >
+      #### Before submitting a feature request, please search through [existing issues](https://github.com/meta-pytorch/forge/issues?q=is%3Aissue+sort%3Acreated-desc+) to see if something similar has already been proposed.
+- type: textarea
+  attributes:
+    label: 🎯 Context/Motivation
+    description: |
+      Describe the problem you're trying to solve or the use case for this feature.
+      What is the motivation behind this request? Why would this feature be valuable?
```
felipemello1: fine to keep as is, but i think it could be a one liner

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,57 @@
...
+    placeholder: |
+      Example: "When training large models, I often need to..."
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: 💡 What does this look like in pseudo-code?
```
felipemello1: not sure if i like this. The intention is good, but maybe too much overhead? e.g. someone requesting multiturn or LoRA. Do they need to come up with a design? Perhaps we could reword to something like "Share any examples or pseudo-code if you have it"

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,57 @@
...
+      ```python
+      # Example pseudo-code showing how you envision using this feature
+      model = MyModel()
+      result = model.new_feature(...)
+      ```
+  validations:
+    required: false
+- type: textarea
+  attributes:
+    label: ✅ Acceptance Criteria
```
felipemello1: Again, intention is good, but it feels like a bit too much overhead. How would an user fill this with "request for LoRA" or "request for multiturn"?

PR #614, opened 2025-12-02
Temporarily use no packing in SFT 

joecummings:
### Context

Our current SFT recipe using packing. However, in order to use packing you need to pass in a block_causal mask in the forward pass. We construct this mask, but it is ignored in titan b/c they only allow additional mask to be passed in *if* the model definition specifies this and denotes that the model is using flex-attn. Since we are unable to control the exact model definitions ourselves, this is a temporary fix to ensure that our training is correct. 

The TRUE fix(es) would be:
1) Move to nightlies everything and push qwen3_flex and llama3_flex versions to titan, then default to using those
2) Work with Titan to allow us to override these model definitions for things like this where we might want to try out Flex *and* normal attention

### Changes

SFT main.py
* Remove references to packing code
* Import and use new padding function
* Add validation to confirm that compile cannot be used

Collate.py
* Add new function that pads to longest seq in the batch

Configs:
* Increased batch size from 1 - 8 now that we are unable to fit multiple sequences in a single sample

### Testing

1. Works with default configs as confirmed by training loss *and* eval loss decreasing

Wandb logs: https://wandb.ai/jcummings/sft-training

2. Validation works correctly

See below output with compile=True
```
[ForgeSFTRecipe-2/8] 2025-12-03 09:47:27 CRITICAL Unhandled exception in actor endpoint
Traceback (most recent call last):
  File "/home/jrcummings/.conda/envs/forge-uv/lib/python3.12/site-packages/monarch/_src/actor/actor_mesh.py", line 935, in handle
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jrcummings/projects/joe-forge/apps/sft/main.py", line 100, in setup
    raise ValueError(
ValueError: training.compile=True is not currently supported. Compile is only supported with flex attention enabled, which requires PyTorch nightly. Please set training.compile=false in your config.
```

### Open questions

* Why is the number of samples processed different between Llama3 8b and Qwen3 8b? Their seq length *should* be the same.

### DIFF
--- a/src/forge/data/collate.py
+++ b/src/forge/data/collate.py
@@ -16,56 +16,63 @@ def collate_padded(batch: list[dict[str, Any]]) -> dict[str, Any]:
     """
     Collate function that pads sequences to the longest sample in the batch.
 
-    Pads 'tokens' with 0 and 'labels' with CROSS_ENTROPY_IGNORE_IDX (-100).
-    Non-tensor fields (like metrics) are collected into lists and flattened
-    if all items are lists.
+    Handles any tensor keys by padding to the longest
+    sequence for that key. Uses 0 as default padding value, and
+    CROSS_ENTROPY_IGNORE_IDX (-100) for 'labels' keys.
+
+    Non-tensor fields are collected into lists. The 'metrics' field is
+    special-cased to be flattened (extended) rather than nested.
 
     Args:
-        batch: List of samples, each containing 'tokens' and 'labels' tensors
+        batch: List of samples, each containing tensor and non-tensor fields
 
     Returns:
-        Batched dict with padded tensors
+        Batched dict with padded tensors and collected non-tensor fields
+
+    Raises:
+        ValueError: If all samples do not have the same keys
     """
     if not batch:
         return {}
 
-    # Find max length in batch
-    max_len = max(sample["tokens"].size(0) for sample in batch)
-
-    # Initialize lists for batched tensors
-    tokens_list = []
-    labels_list = []
-
-    # Pad each sample to max_len
+    # Verify all samples have the same keys
+    first_sample_keys = batch[0].keys()
     for sample in batch:
-        seq_len = sample["tokens"].size(0)
-        pad_len = max_len - seq_len
+        if sample.keys() != first_sample_keys:
+            raise ValueError(
+                f"All samples must have the same keys. Expected {first_sample_keys}, got {sample.keys()}"
+            )
 
-        # Pad tokens with 0
-        padded_tokens = F.pad(sample["tokens"], (0, pad_len), value=0)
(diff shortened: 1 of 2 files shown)


apps/sft/main.py
```diff
@@ -97,6 +95,13 @@ def record_batch_metrics(self, data_metrics: list):
 
     @endpoint
     async def setup(self):
+        # Validate that compile is only used with flex attention
+        if self.job_config.training.compile:
+            raise ValueError(
+                "training.compile=True is not currently supported. "
+                "Compile is only supported with flex attention enabled, which requires PyTorch nightly. "
```
JenniferWang, line 102: Objection to start a main issue tracking the nightly build?

joecummings: Sounds good! But can we first nail down the different subtasks via the Google Doc I just shared? Then we can translate to a GI.

src/forge/data/collate.py
```diff
@@ -7,6 +7,65 @@
...
+        "labels": torch.stack(labels_list),
+    }
+
+    # Collect non-tensor fields (like metrics)
+    for key in batch[0].keys():
+        if key not in ["tokens", "labels"]:
+            result[key] = [sample[key] for sample in batch]
+            # Flatten if all are lists
+            if all(isinstance(item, list) for item in result[key]):
+                result[key] = [item for sublist in result[key] for item in sublist]
```
JenniferWang: Is this a common practice? Feels like unnecessary operation / tribal knowledge

src/forge/data/collate.py
```diff
@@ -7,6 +7,65 @@
...
+from forge.data.utils import CROSS_ENTROPY_IGNORE_IDX
+
+
+def collate_padded(batch: list[dict[str, Any]]) -> dict[str, Any]:
+    """
+    Collate function that pads sequences to the longest sample in the batch.
+
+    Pads 'tokens' with 0 and 'labels' with CROSS_ENTROPY_IGNORE_IDX (-100).
+    Non-tensor fields (like metrics) are collected into lists and flattened
+    if all items are lists.
```
JenniferWang: Is it common practice to assume `tokens` and `labels` are the keys for `collate_padded`?

joecummings: So, yes these are the keys that need to be collated for Titan and Hugging Face's TRL. 

However, you bring up a good point that there's no need to overly commit to that ATM. I changed the function to mirror the generic logic of collate_packed. 

I'm not *thrilled* that we determine which data to stack based on checking if the first element is a Tensor, but it works for now.

src/forge/data/collate.py
joecummings: I simplified - good call

PR #615, opened 2025-12-03
Update `forge` references to `torchforge` in MAST setup

daniellepintz:
This is needed as the repo was renamed.

### DIFF
--- a/src/forge/controller/launcher.py
+++ b/src/forge/controller/launcher.py
@@ -206,7 +206,7 @@ class MastLauncher(BaseLauncher):
         self.timeout_sec = 1 * 60 * 60  # Kill the job if idle for 1 hour
         self.user = getpass.getuser()
         self.work_dir = f"/home/{self.user}"
-        self.edittable_workspaces = ["forge"]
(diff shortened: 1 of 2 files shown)


src/forge/controller/launcher.py
```diff
@@ -206,7 +206,7 @@ def __init__(
         self.timeout_sec = 1 * 60 * 60  # Kill the job if idle for 1 hour
         self.user = getpass.getuser()
         self.work_dir = f"/home/{self.user}"
```
JenniferWang, line 208: This is hardcoded and could you check if it works when launching from `fbsource/fbcode/pytorch/torchforge`? I need to do the following https://www.internalfb.com/diff/D88381552 (see the launcher.py)

JenniferWang (approved):
No objection to updating the name but this is not sufficient for a good experience.

src/forge/controller/launcher.py
daniellepintz: I don't think it will work from `fbsource/fbcode/pytorch/torchforge` because the path is different, but looks like your diff should fix it

PR #622, opened 2025-12-05
Fix example command in `sandbox/rl_trainer/main.py`

daniellepintz:
The current example config does not work because it runs on slurm. Fixing to a command that does work and runs locally.

### DIFF
--- a/tests/sandbox/rl_trainer/main.py
+++ b/tests/sandbox/rl_trainer/main.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-# Usage: python -m tests.sandbox.rl_trainer.main --config apps/grpo/qwen3_32b.yaml
+# Usage: python -m tests.sandbox.rl_trainer.main --config apps/grpo/qwen3_1_7b.yaml
 
 import asyncio
 


felipemello1:
hey, happy to approve, but can you help me understand why 1.7b would run but 30b wouldnt?

halflearned:
hey I saw this issue today. imo this isn't the right fix. the problem is that the `qwen3_32b.yaml` file contains the attributes `hosts:1`, which assumes a scheduler/remote host. just removing those lines from that file fixed the issue for me when running locally.

daniellepintz:
yes indeed the problem is `qwen3_32b.yaml` is configured to use slurm. do we want to change this?

JenniferWang:
@felipemello1  @daniellepintz  this is, generally speaking a UX problem -- what's our default examples to be based on? Local DevGPU with 8 card? 

On the other hand, for now, we can certainly augment the example run command with different mode override. 

```
Local DevGPU: 
python -m tests.sandbox.rl_trainer.main --config apps/grpo/qwen3_1_7b.yaml

Slurm: 
python -m tests.sandbox.rl_trainer.main --config apps/grpo/qwen3_1_7b.yaml provisioner.launcher='slurm' 

MAST:
...
```

Given that the comment is accurate. I suggest closing the PR and include the improvement in the UX work.

PR #628, opened 2025-12-08
Enable online writing to WandB in MAST jobs

daniellepintz:
# Description
In this PR we use the secrets API to extract the WandB API key and add it to the environment. 

# Test Plan
```
./.meta/mast/launch.sh .meta/mast/qwen3_1_7b_mast.yaml
```
From logs:
> [wandb] Attempting to retrieve API key from keychain secret_name='TORCHFORGE_WANDB_API_KEY'
> [wandb] Successfully retrieved API key from keychain.
> wandb: 🚀 View run at https://meta.wandb.io/torchforge/grpo-training/runs/1ql4ar91

### DIFF
--- a/.meta/mast/main.py
+++ b/.meta/mast/main.py
@@ -34,6 +34,9 @@ DEFAULT_CHECKPOINT_FOLDER = "/mnt/wsfuse/teamforge/forge_runs/"
 
 def setup_wandb_api_key() -> None:
     # add wandb API key to the environment
+    if "WANDB_API_KEY" in os.environ:
+        print("[wandb] WANDB_API_KEY already set in environment.")
+        return
     secret_name = "TORCHFORGE_WANDB_API_KEY"
     print(f"[wandb] Attempting to retrieve API key from keychain {secret_name=}")
     try:
--- a/.meta/mast/README.md
+++ b/.meta/mast/README.md
@@ -119,3 +119,7 @@ This ensures that when MAST runs with `HF_HUB_OFFLINE=1`, the transformers libra
 Both cache and model files are stored under:
 - **Cache**: `/mnt/wsfuse/teamforge/hf` (set via `HF_HOME`)
 - **Model weights**: `/mnt/wsfuse/teamforge/hf/<model_name>`
+
+
+#### Weights & Biases
+If you are part of the torchforge team on WandB, then WandB will work out of the box; the link can be found in the MAST logs. If you are not part of the torchforge team on WandB, then you will need to set the "WANDB_API_KEY" environment variable to your WandB API key.


src/forge/controller/launcher.py
```diff
@@ -293,7 +326,8 @@ def build_appdef(self) -> specs.AppDef:
                 "TORCHDYNAMO_VERBOSE": "1",
                 "VLLM_TORCH_COMPILE_LEVEL": "0",
                 "VLLM_USE_TRITON_FLASH_ATTN": "0",
-                "WANDB_MODE": "offline",
+                "WANDB_MODE": "online",
+                "WANDB_BASE_URL": "https://meta.wandb.io",
```
JenniferWang: i wonder if you think moving these MAST controller stuff to a separate `src/forge/controller/fb/` make sense.

JenniferWang:
What's the steps to set up the team keychain for wandb?

src/forge/controller/launcher.py
daniellepintz: yeah I think we can move `MastLauncher` to a separate file, I can do that after this Pr

.meta/mast/main.py
```diff
@@ -55,6 +56,40 @@ async def main(cfg: DictConfig, mode: str = "detached", extra_args: list = None)
...
+        print("[wandb] Successfully retrieved API key from keychain.")
+    except Exception as keychain_exception:
+        print(
+            f"[wandb] Failed to retrieve API key from keychain. {keychain_exception=}"
+        )
+
+    if wandb_api_key is not None:
+        os.environ["WANDB_API_KEY"] = wandb_api_key
+        os.environ["WANDB_BASE_URL"] = WANDB_HOST
+
```
JenniferWang: I think we should terminate the launcher if `wandb_api_key` is `None`.
Otherwise, it just delays the crash in MAST

daniellepintz: Done!

.meta/mast/main.py
```diff
@@ -31,6 +32,41 @@
 DEFAULT_CHECKPOINT_FOLDER = "/mnt/wsfuse/teamforge/forge_runs/"
 
 
+def setup_wandb_api_key() -> None:
+    # add wandb API key to the environment
+    secret_name = "TORCHFORGE_WANDB_API_KEY"
+    print(f"[wandb] Attempting to retrieve API key from keychain {secret_name=}")
```
JenniferWang, line 41: This logic won't work for people outside of the team right? Maybe check if `WANDB_API_KEY` is set first, if not, try resolving it from the keychain. We can add a line in the README to explain the behavior.

daniellepintz: Done!

JenniferWang (approved):

PR #629, opened 2025-12-08
remove metrics

felipemello1:
oops, i merged https://github.com/meta-pytorch/torchforge/pull/606 before i pushed my latest changes


apps/grpo/main.py
```diff
@@ -183,14 +183,19 @@ def simple_grpo_loss(
     loss = -(mean_policy_loss - beta * mean_kl)
 
     # Log metrics
+    # TODO: Better design - have loss function return all metrics as a dict,
```
allenwang28, line 186: do we want to do this now or in a separate PR?

felipemello1: separate PR. I am refactoring the loss today

PR #632, opened 2025-12-09
[MAST] fix `get_tokenizer` error for MAST jobs

JenniferWang:
## Issue

MAST job fails with a stacktrace pointing to tokenizer trying to download from HF, which is not supported in MAST (no network access to HF) 
```
monarch._src.actor.actor_mesh.ActorError: Actor call DatasetActor.setup failed.
  File "/packages/monarch_default_workspace/conda/lib/python3.10/site-packages/monarch/_src/actor/actor_mesh.py", line 1167, in handle
  File "/packages/monarch_default_workspace/workspace/torchforge/apps/grpo/main.py", line 268, in setup
    self._tokenizer = get_tokenizer(self.model)
  File "/packages/monarch_default_workspace/conda/lib/python3.10/site-packages/vllm/transformers_utils/tokenizer.py", line 238, in get_tokenizer
    tokenizer = AutoTokenizer.from_pretrained(
  File "/packages/monarch_default_workspace/conda/lib/python3.10/site-packages/transformers/models/auto/tokenization_auto.py", line 1156, in from_pretrained
    return tokenizer_class.from_pretrained(pretrained_model_name_or_path, *inputs, **kwargs)
  File "/packages/monarch_default_workspace/conda/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 2113, in from_pretrained
    return cls._from_pretrained(
  File "/packages/monarch_default_workspace/conda/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 2395, in _from_pretrained
    tokenizer = cls._patch_mistral_regex(
  File "/packages/monarch_default_workspace/conda/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 2438, in _patch_mistral_regex
    if _is_local or is_base_mistral(pretrained_model_name_or_path):
  File "/packages/monarch_default_workspace/conda/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 2432, in is_base_mistral
    model = model_info(model_id)
  File "/packages/monarch_default_workspace/conda/lib/python3.10/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
  File "/packages/monarch_default_workspace/conda/lib/python3.10/site-packages/huggingface_hub/hf_api.py", line 2660, in model_info
```

## Root cause
Although for the conda-MAST package, as well as the OSS version, we pinned v0.10.0. So there should be no breakage from vllm so long as we don't upgrade the version right?

This is incorrect! As you can see from the stack trace, it's pointing to the hf transformer library. In other words, vllm takes dependency on the transformer library. Although the vllm version is fixed, the dependent library `transformers` is not (https://fburl.com/code/y6ppe22b). So very likely, the breakage is due to some changes in the transformer library.

## Fix
Use the file path instead of model name makes the issue go away.

## Test Plan:
Verify the fix in D88381552


.meta/mast/qwen3_14b_mast.yaml
JenniferWang, line 1: I removed the 14B and 8B config files because they probably don't give us any additional signal on the infra setup.

PR #635, opened 2025-12-09
[refactor 1/n] - move utilities out of main.py to src/

felipemello1:
## Summary

Extract self-contained utility components from the GRPO app into the forge library for reusability.

## Changes

  Moved:
  - Episode dataclass → src/forge/rl/types.py
  - Group type alias → src/forge/rl/types.py
  - Policy type alias → src/forge/rl/types.py
  - collate() → src/forge/rl/collate.py
  - ComputeAdvantages → src/forge/rl/advantage.py
  - RewardActor → src/forge/rl/grading.py
  - drop_weights() → src/forge/util/checkpoint.py

  Deleted:
  - apps/grpo/notebook.ipynb

## Testing

Ran vs main, scores are similar

### DIFF
--- a/src/forge/rl/advantage.py
+++ b/src/forge/rl/advantage.py
@@ -13,6 +13,7 @@ from forge.rl.types import Group
 from monarch.actor import endpoint
 
 
+# TODO: this doesn't need to be an actor
 @dataclass
 class ComputeAdvantages(ForgeActor):
     @endpoint


JenniferWang:
I think we have an existing folder called `src/forge/data_models` -- do you plan to move them over to `rl/types`?

src/forge/rl/advantage.py
```diff
@@ -0,0 +1,25 @@
...
+@dataclass
+class ComputeAdvantages(ForgeActor):
+    @endpoint
+    async def compute(self, group: Group) -> list[float]:
+        # TODO: add batch processing
+        rewards = torch.tensor([[e.reward for e in group]])
+        mean = rewards.mean(1, keepdim=True)
+        std = rewards.std(1, keepdim=True)
+        advantages = (rewards - mean) / (std + 1e-4)
+        return advantages.squeeze(0).tolist()
```
JenniferWang, line 26: This being an actor does not make sense. 
Also should it be merged into `forge/util/ops.py`?

felipemello1:
>  think we have an existing folder called src/forge/data_models -- do you plan to move them over to rl/types?

Probably. I didnt wan't this PR to be hard to review, so i didnt make changes to the functions, i just moved them. Follow up refactoring will have more changes.

src/forge/rl/advantage.py
felipemello1: I dislike how loaded "util" is, but i agree that it doesnt need to be an actor. Do you think i should make the change in this PR? I am ok with it, but my intuition is that i should touch these files in follow up PRs.

JenniferWang: No. But it's generally good practice, as you go through the refactor, add a `TODO` comment so we can grip later to catch any missing refactoring.

felipemello1: will add a todo

PR #638, opened 2025-12-09
[MAST] Commit qwen3_30b_a3b_mast.yaml

JenniferWang:
## Summary:

Initial commit for launching Qwen 30b MOE  on MAST

## Test Plan:
https://fburl.com/msl/hxkn8xb0
Current hitting RDMA crash (passed the checkpoint loading)

### DIFF
--- a/.meta/mast/qwen3_30b_a3b_mast.yaml
+++ b/.meta/mast/qwen3_30b_a3b_mast.yaml
@@ -88,7 +88,7 @@ trainer:
   activation_checkpoint:
     mode: full
   comm:
-    # 30B MoE model can require more time than 1.7B's 1200s
+    # 30B MoE model can require more time to load checkpoint than 1.7B's 1200s
     init_timeout_seconds: 1800
   dcp_path: ${checkpoint_folder}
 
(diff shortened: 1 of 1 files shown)


.meta/mast/qwen3_30b_a3b_mast.yaml
```diff
@@ -0,0 +1,165 @@
...
+    pipeline_parallel_degree: 1
+    context_parallel_degree: 1
+    expert_parallel_degree: 1
+  checkpoint:
+    enable: true
+    initial_load_path: /mnt/wsfuse/teamforge/hf/qwen3_30b_a3b_2
+    initial_load_in_hf: true
+  comm:
+    # Reference model also loads from checkpoint; so setting a higher timeout
+    init_timeout_seconds: 1800
```
JenniferWang, line 126: @daniellepintz the main catch is there are two places loading the checkpoint and we need to loosen the timeout in both places.

.meta/mast/qwen3_30b_a3b_mast.yaml
```diff
@@ -0,0 +1,165 @@
...
+  policy:
+    procs: ${policy.engine_args.tensor_parallel_size}
+    num_replicas: 1
+    hosts: 1
+    with_gpus: true
+    mesh_name: policy
+  ref_model:
+    procs: 4
+    num_replicas: 1
+    hosts: 1
```
JenniferWang, line 139: I also added a designated host for the ref model.

PR #639, opened 2025-12-09
[refactor 2/n] - move dataset and reward to the experiment folder

felipemello1:
## Summary

An experiment is: 
- env, 
- dataset, 
- rewards, 
- train.py putting everything together.

In this PR, i move dataset and reward to their respective files, and delete src/rewards.py, as this should be per experiment.

## Testing

image


apps/grpo/data.py
```diff
@@ -0,0 +1,84 @@
...
+
+from datasets import load_dataset
+from forge.controller.actor import ForgeActor
+from forge.observability.metrics import record_metric, Reduce
+from monarch.actor import endpoint
+from vllm.transformers_utils.tokenizer import get_tokenizer
+
+
+@dataclass
+class DatasetActor(ForgeActor):
```
JenniferWang, line 17: Should this be in a common folder under, like `rl/datasets`?  or just `datasets`?
This is not specific to GPRO right?

JenniferWang:
The PR itself is fine -- it's just harder for the reviewer to see the final picture.

apps/grpo/data.py
felipemello1: yes, but i think we should add a few experiments first and see how it interacts with OpenEnv or other libraries before we settle on a dataset abstraction. For now, keeping it exposed is probably healthier.

E.g. for wordle maybe we dont have a conventional dataset
For search, we may need to setup a database
For "20 questions", we may not have a dataset
For "verifiers" environments, the dataset is defined in the env

PR #644, opened 2025-12-10
: provisioner: explicit monarch hostmesh shutdown

shayne-fletcher:
Summary:
this change extends the explicit-shutdown work from the previous diff into the provisioner layer. the provisioner was still leaving `HostMesh` instances (including the implicit local host) alive at teardown, which meant their underlying processes could still be terminated via `SIGTERM `during interpreter or allocator shutdown. that put us back on the `SIGTERM` path that produces the unwanted folly backtrace on exit.

we now explicitly `await host_mesh.shutdown()` for every `HostMesh` managed by the provisioner, and then call `shutdown_context()` to tear down the remaining actor runtime state. this ensures all meshes and their processes exit cleanly under their own shutdown path rather than being dropped and signaled, keeping the shutdown sequence free of `SIGTERM`-driven backtraces.

i expect this diff together with

Differential Revision: D88856830

### DIFF
--- a/src/forge/controller/provisioner.py
+++ b/src/forge/controller/provisioner.py
@@ -22,7 +22,14 @@ from forge.types import ProcessConfig, ProvisionerConfig
 from monarch._src.actor.actor_mesh import ActorMesh
 from monarch._src.actor.shape import Extent
 
-from monarch.actor import Actor, endpoint, HostMesh, ProcMesh, this_host
+from monarch.actor import (
+    Actor,
+    endpoint,
+    HostMesh,
+    ProcMesh,
+    shutdown_context,
+    this_host,
+)
 
 from monarch.tools import commands
 from monarch.utils import setup_env_for_distributed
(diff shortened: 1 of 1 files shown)


allenwang28:
hey @shayne-fletcher , thanks for the PR! Unfortunately we cannot check this in yet as it breaks CI which is still pinned to the latest Monarch stable cc @JenniferWang

shayne-fletcher:
> hey @shayne-fletcher , thanks for the PR! Unfortunately we cannot check this in yet as it breaks CI which is still pinned to the latest Monarch stable cc @JenniferWang

thanks allen. the diff is still in review but assuming it's accepted, i would appreciate a steer on the procedures for getting it landed.

JenniferWang:
> > hey @shayne-fletcher , thanks for the PR! Unfortunately we cannot check this in yet as it breaks CI which is still pinned to the latest Monarch stable cc @JenniferWang
> 
> thanks allen. the diff is still in review but assuming it's accepted, i would appreciate a steer on the procedures for getting it landed.

Sorry for the delayed response @shayne-fletcher. We're actively working on moving forge nightly build and CI, in the future, shall be run on the Monarch nightly build. Tracked in https://github.com/meta-pytorch/torchforge/issues/625 .

Not sure what's the common practice for OSS but i am thinking merging this diff once we closed the above issue?

shayne-fletcher:
@JenniferWang  @allenwang28  is this ready to go now? can you help me get it merged? please advise.

PR #650, opened 2025-12-16
Move MAST code to internal directory

daniellepintz:
Summary: This code should live internally only.

Differential Revision: D89267782


src/forge/controller/launcher.py
```diff
@@ -214,11 +214,9 @@ def __init__(
         self.sku = "gtt_any"
         self.timeout_sec = 1 * 60 * 60  # Kill the job if idle for 1 hour
         self.user = getpass.getuser()
-        self.work_dir = f"/home/{self.user}"
-        self.edittable_workspaces = ["torchforge"]
         self.remote_work_dir = "/packages/monarch_default_workspace/workspace/"
         self.editable_workspace_paths = [
-            f"{self.work_dir}/{workspace}" for workspace in self.edittable_workspaces
+            f"/data/users/{self.user}/fbsource/fbcode/pytorch/torchforge"
```
JenniferWang, line 219: Add a TODO task to move these logic as well?

daniellepintz: already have a diff for this D89284395 :)

PR #657, opened 2025-12-18
upgrade monarch & torchstore version

JenniferWang:
## tl;dr

The typical release cycles does not meet our need to co-development with the dependency stacks (`monarch`, `torchstore`, `torchtitan`) -- we need the latest fixes, features and we cannot afford to wait/coordinate with all downstream releases, e.g. https://github.com/meta-pytorch/torchforge/pull/644. On the flip side, forge's upstream may also want to upgrade to the latest code.

Therefore we propose the following
* for `torchforge` maintainers to move fast, instead of pinning the stable release, we'll pin the specific nightly version of these libraries:  `monarch`, `torchtitan`, `torchstore`. HUGE thanks to monarch team for removing the torch dependency from monarch, this means the `monarch` nightly build is compatible with the stable torch version; `torchtitan` has no dependency on torch. As for `torchstore`, we've coordinated with @amirafzali on setting up a nightly release for torchstore that has no explicit dependency on torch or monarch. 
    * Additionally, we will set up an integration test that runs continuously against the latest nightly build from the dependencies (not the pinned version) to capture backward incompatible changes. This also serves as a test plan when we need to update the pinned nightly version to pick up new downstream changes. 
    * When we do a stable release, we'll change the nightly pinned versions to use stable version whenever possible. This is probably low ROI for now given how rapidly things are moving. 
* for `torchforge` users to move fast, we already have a nightly flow that's supposed to package the latest code. The remaining questions is why the current nightly branch is divergent from the main branch: https://github.com/meta-pytorch/torchforge/tree/nightly . Titan has the same issue. Asked here https://fb.workplace.com/groups/4571909969591489/posts/25498493553173159
image

## Nuances 
* `torchmonarch` nightly build does not follow the PyTorch nightly naming convention -- the nightly build is published under a different package name `torchmonarch-nightly`. This breaks convention of specifying "<package_name> >= <stable_version>" to allow nightly build.
* `torchstore` _also_ hardcoded monarch and torch in its dependency list. For now we're pinning a specific branch that has no dependency to make both  `uv` and `install.sh` script works. We'll move to nightly once that's ready. 

## Test plan
* clean conda env. use `install.sh` script; the grpo app runs
* verify the dependency does not have `torchmonarch` , torchstore contains latest changes 
```
forge                             0.0.0                            /home/jiyue/torchforge
torch                             2.9.0+cu128
torchmonarch-nightly              2025.12.17
torchstore                        0.0.0.dev0
torchtitan                        0.2.0
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,6 @@ dev = [
     "tomli>=1.1.0",
     "anyio",
     "pytest-asyncio",
-    "matplotlib",
     "multiprocess",
 ]
 docs = [
@@ -57,6 +56,7 @@ docs = [
     "sphinx-design==0.6.1",
     "sphinxcontrib-mermaid==1.0.0",
     "sphinx-gallery==0.19.0",
+    "matplotlib",
     "myst-parser",
     "sphinx-sitemap==2.7.1",
     "sphinx-autodoc-typehints==1.25.3",


allenwang28:
@JenniferWang for the Monarch dependency, is there any issue with keeping the `main` branch fixed to `torchmonarch-nightly` but for each stable release, creating a tagged version that swaps to stable?

felipemello1:
regarding the no-monarch branch, many libraries put dependencies as optional, e.g. `pip install polars[rtcompat]`

would that be an option? e.g. `pip install torchstore` and `pip install torchstore[monarch]`? I am not familiar with using an alternative branch as a solution. Is this common?

JenniferWang:
> regarding the no-monarch branch, many libraries put dependencies as optional, e.g. `pip install polars[rtcompat]`
> 
> would that be an option? e.g. `pip install torchstore` and `pip install torchstore[monarch]`? I am not familiar with using an alternative branch as a solution. Is this common?

I don't know what's the common practice. But i saw @joecummings 's been doing that in, say https://github.com/meta-pytorch/torchforge/blob/main/.github/workflows/build_vllm.yaml#L16 

The issue is there's no CI setup for torchstore -- I need the fixes like this to upgrade monarch https://github.com/meta-pytorch/torchstore/pull/74 so using a branch is the cleanest workaround. @LucasLLC

JenniferWang:
> @JenniferWang for the Monarch dependency, is there any issue with keeping the `main` branch fixed to `torchmonarch-nightly` but for each stable release, creating a tagged version that swaps to stable?

Nope. Love this proposal!

assets/versions.sh
```diff
@@ -10,9 +10,9 @@
 # Stable versions of upstream libraries for OSS repo
 PYTORCH_VERSION="2.9.0"
 VLLM_VERSION="v0.10.0"
-MONARCH_VERSION="0.1.2"
+MONARCH_NIGHTLY_VERSION="2025.12.17"
```
felipemello1, line 13: noob question: what happens in a few weeks or month? do we have to update the version again or is there a way to always use the latest for nightlies?

assets/versions.sh
```diff
@@ -10,9 +10,9 @@
...
 TORCHTITAN_VERSION="0.2.0"
-TORCHSTORE_VERSION="0.1.2"
+TORCHSTORE_BRANCH="no-monarch"
```
felipemello1: i didnt check it on the torchstore side, but is the no-monarch == main?

felipemello1:
the nightly vs main seems like a serious bug, with 206 files containing differences: https://github.com/meta-pytorch/torchforge/compare/main...nightly

assets/versions.sh
JenniferWang: We don't want to automatically point to the latest version for nightlies in the main branch because by doing so, we know the exact version that's working. 

We could automate bumping the version but it shall always be a pinned and tested. I think this can be tracked as a saparate task.

JenniferWang:
> the nightly vs main seems like a serious bug, with 206 files containing differences: [main...nightly](https://github.com/meta-pytorch/torchforge/compare/main...nightly)

JoeC set this up for both titan and forge and they're broken. I don't know how Pytorchbot works so need to wait for him.

assets/versions.sh
JenniferWang: Correct. But the main branch will move forward while the no-monarch branch stays still. I will delete this branch once the torchstore nightly is setup.

PR #666, opened 2025-12-22
Add more fine-grained wandb logging for actual weight update

daniellepintz:
Summary:
Add wandb logging for the actual weight update by itself in `generator_perf/update_weights/actual_weight_update`

Also add more descriptive logging to std out.

Differential Revision: D89575714

### DIFF
--- a/src/forge/actors/generator.py
+++ b/src/forge/actors/generator.py
@@ -37,7 +37,7 @@ from forge.controller import (
 from forge.data_models.completion import Completion
 from forge.data_models.prompt import to_prompt
 from forge.observability.metrics import record_metric, Reduce
-from forge.observability.perf_tracker import Tracer
+from forge.observability.perf_tracker import trace, Tracer
 from forge.types import ProcessConfig
 from forge.util._shared_tensor import SharedTensor, SharedTensorHandle
 from monarch.actor import current_rank, endpoint, ProcMesh, this_host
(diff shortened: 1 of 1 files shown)


src/forge/actors/generator.py
```diff
@@ -477,12 +477,26 @@ async def update_weights(self, version: int) -> None:
             if fetch_fut is not None:
                 fetched_weights = await fetch_fut
                 # Call update_weights on every policy_worker
+                update_weights_start = time.perf_counter()
                 await self.worker.update_weights.call(
                     shared_memory_state_dict=fetched_weights
                 )
```
JenniferWang, line 491: This is a monarch endpoint call and super verbose from user perspective, is there simpler logging from monarch @moonli ?

src/forge/actors/generator.py
```diff
@@ -477,12 +477,26 @@ async def update_weights(self, version: int) -> None:
...
+                    update_weights_duration,
+                    Reduce.MEAN,
+                )
                 await self._drop_shared_memory(fetched_weights)
             else:
+                update_weights_start = time.perf_counter()
                 await self.worker.update_weights.call(version=version)
+                update_weights_duration = time.perf_counter() - update_weights_start
+                record_metric(
+                    "generator_perf/update_weights/actual_weight_update",
```
JenniferWang: What is `actual` mean here? Is it about vLLM updating weight? If so, could you make it clear? Also do you want to distinguish the two branches?

src/forge/actors/generator.py
daniellepintz: I think we can use this:

https://github.com/meta-pytorch/torchforge/blob/e940fd89ad9b262e690988f2f1a5ee6cb0d25574/src/forge/observability/perf_tracker.py#L396

src/forge/actors/generator.py
daniellepintz: I basically want to measure the call to `GeneratorWorker.update_weights`, the next version of this PR should make that more clear

daniellepintz: I am ok to not distinguish the two branches, and instead let the std out logging do that; wdyt?

PR #667, opened 2025-12-22
remove hard-coded versions from assets/versions.sh and fix CUDA path in scripts/install.s

JenniferWang:
Summary: There's no need to specify the package version for `torchmonarch-nightly` and `torchtitan` in two separate places. We leave the canonical version in `pyproject.yaml` and remove the ones from `assets/versions.sh`

Differential Revision: D89667000

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -198,12 +198,6 @@ main() {
     pip install "setuptools<80"
     python -m pip install vllm --no-cache-dir --index-url https://download.pytorch.org/whl/preview/forge
 
-    # Install monarch nightly
-    pip install torchmonarch-nightly==$MONARCH_NIGHTLY_VERSION
-
-    # Install torchtitan
-    pip install torchtitan==$TORCHTITAN_VERSION
-
     # Install torchstore without monarch or torch dependency
     pip install "git+https://github.com/meta-pytorch/torchstore.git@${TORCHSTORE_BRANCH}"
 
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -260,14 +250,15 @@ EOF
...
+    # Priority: system CUDA driver > CUDA toolkit > conda libs
     local py_shim_activate="${conda_env_dir}/etc/conda/activate.d/python_ld_shim.sh"
     cat > "$py_shim_activate" << 'EOF'
-# Define python wrappers that only set LD_LIBRARY_PATH for the launched process
-python()  { LD_LIBRARY_PATH="${CONDA_PREFIX}/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" command python  "$@"; }
-python3() { LD_LIBRARY_PATH="${CONDA_PREFIX}/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" command python3 "$@"; }
+# Define python wrappers that set LD_LIBRARY_PATH only for the launched process
+# Use system CUDA driver (newer) instead of compat (outdated stub that may be incompatible)
+python()  { LD_LIBRARY_PATH="/usr/lib64:/usr/local/cuda-12.8/lib64:${CONDA_PREFIX}/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" command python  "$@"; }
+python3() { LD_LIBRARY_PATH="/usr/lib64:/usr/local/cuda-12.8/lib64:${CONDA_PREFIX}/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" command python3 "$@"; }
```
felipemello1, line 261: what if the user doesnt use cuda-12.8?

scripts/install.sh
```diff
@@ -198,12 +198,6 @@ main() {
     pip install "setuptools<80"
     python -m pip install vllm --no-cache-dir --index-url https://download.pytorch.org/whl/preview/forge
 
-    # Install monarch nightly
-    pip install torchmonarch-nightly==$MONARCH_NIGHTLY_VERSION
-
-    # Install torchtitan
-    pip install torchtitan==$TORCHTITAN_VERSION
```
felipemello1, line 205: how do these get installed now?

JenniferWang, line 205: @felipemello1  to your question above.... pip should just install from the pyproject from 

` pip install -e ".[dev]"` below

scripts/install.sh
JenniferWang, line 225: @felipemello1  CUDA version has always been fixed to 12.8

JenniferWang: @felipemello1  there are multiple places where we hardcode CUDA 12.8. (you can search for `cu128`) . We could potentially revisit this restriction after removing removing this custom vLLM build https://github.com/meta-pytorch/torchforge/blob/main/.github/workflows/build_vllm.yaml#L35 (https://github.com/meta-pytorch/torchforge/pull/668/changes)

JenniferWang:
> changes look reasonable, but i didnt test it on my machine. If you think we need someone else to test it before merging, let me know and i can stop what i am doing to run the install script

I've tested on my machine. No rush!

PR #680, opened 2025-12-26
add generator offline throughput benchmark (#675)

JenniferWang:
Summary:

Add offline throughput benchmark for the generator for iterating on vLLM refactoring.
The benchmark can indeed serve two purposes
1. `random` dataset for testing throughput
2. `fixed` dataset with fixed prompt for sanity check the response

Validate that regression test is setup correctly by manually introducing a regression via setting `max_num_seqs` to be 32 in model config.

Differential Revision: D89768647

### DIFF
--- a/benchmarks/generator/throughput.py
+++ b/benchmarks/generator/throughput.py
@@ -1,0 +1,320 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Offline throughput benchmark for torchforge generators.
+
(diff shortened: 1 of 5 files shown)


benchmarks/generator/throughput.py
```diff
@@ -0,0 +1,320 @@
...
+        raise ValueError(f"Unknown dataset type: {dataset_type}")
+
+    requests: list[BenchmarkRequest] = dataset.generate()
+
+    print(f"\nRunning throughput benchmark ({num_requests} requests)...")
+    prompts = [req.prompt for req in requests]
+    request_ids = [req.request_id for req in requests]
+
+    start = time.perf_counter()
+    # TODO: here we're measuring two things together: vllm generation and data transmition.
```
daniellepintz: what is meant exactly by data transmition?

PR #682, opened 2025-12-27
Add `qwen3_30b_a3b.yaml` config for test_policy_update.py

daniellepintz:
Summary:
Add a config to test weight sync for Qwen3 30B MoE model
Note: the test will fail with this config until we update the torchtitan version.

Differential Revision: D89833611


JenniferWang:
This test is used for two purposes
- Correctness check for onboarding new model and therefore the test can run on any model config
- Correctness check for changing infra code (e.g. Generator) and therefore it's a test and shall use a small model config. 

In general, 
1. Please do not introduce new test fixture that fails the test.
2. Do not introduce new test fixture unless it tests new aspect of our infra. 

Also, this test is not configured to run continuously yet ... so it's not used to test infra regression :(

PR #686, opened 2025-12-29
remove titan commit hash for MAST

JenniferWang:
Summary:
See the previous diff for context. Split the logic into two diffs to make GH
export work.

Reviewed By: daniellepintz

Differential Revision: D89903684


daniellepintz (approved):

PR #693, opened 2026-01-02
Add integration test workflow to CI

daniellepintz:
## Summary:
Adding a GH action workflow to test integration tests continuously. For now just testing `test_policy_update.py`. Will add other integration tests in next PRs.

Runs on every push to main, and on workflow_dispatch (manual trigger).

Also, needed to modify config of `qwen3_1_7b_tp.yaml` to use fewer GPUs because the CI machines only have 4 GPUs.

## Test Plan:
Test by adding `on: pull_request:` to the yaml file; then remove it before landing.

Job succeeds: https://github.com/meta-pytorch/torchforge/actions/runs/20723487277/job/59493218607?pr=693

Differential Revision: D90031998

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -1,0 +1,46 @@
+name: Integration Tests (8 card)
+
+on:
+  push:
+    branches: [ main ]
+  workflow_dispatch:
+
+concurrency:
+  group: integration-test-${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  id-token: write
+  contents: read
+
+defaults:
+  run:
+    shell: bash -l -eo pipefail {0}
+
+jobs:
+  integration_test:
+    if: github.repository_owner == 'meta-pytorch'
+    runs-on: linux.g4dn.metal.nvidia.gpu
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v4
+      - name: Setup conda env
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-update-conda: true
+          miniconda-version: "latest"
+          activate-environment: test
+          python-version: '3.12'
+      - name: Update pip
+        run: python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


tests/integration_tests/fixtures/qwen3_1_7b_tp.yaml
```diff
@@ -1,4 +1,4 @@
-# trainer tp = 2, policy tp = 4
+# trainer tp = 1, generator tp = 2
```
JenniferWang: The point of this test is exactly about the sharding. If we reduce the sharding to 1, the test is going to be less effective. 

Is there a way to get a larger CI machine?  Or you should just use `qwen3_1_7b_no_tp.yaml`

daniellepintz: Sg, I updated to use `qwen3_1_7b_no_tp.yaml` for now. In parallel will see if we can get a larger CI machine

JenniferWang:
Great! A separate question is -- should we reuse the same workflow as the GPU test or a new workflow?

daniellepintz:
Ah, great news, I just tested on a different CI machine with 8 GPUs (`linux.g4dn.metal.nvidia.gpu`) as per Amir's suggestion, and the config with tp works now!

> Great! A separate question is -- should we reuse the same workflow as the GPU test or a new workflow?

I think we should create a new workflow because a) this is a somewhat expensive test, no need to run it on every push to a PR like the gpu test and b) now we are using a different machine than the gpu test

tests/integration_tests/fixtures/qwen3_1_7b_tp.yaml
daniellepintz: I was able to get a larger CI machine, so we can use the `qwen3_1_7b_tp.yaml` config without any changes :)

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,46 @@
+name: Integration Tests (GPU)
```
JenniferWang: Maybe `Integration Tests (8 card)`

PR #699, opened 2026-01-08
Loss refactor

felipemello1:
image
image
image

NOTE1: ratio starts at 0.8, which is odd. It should start closer to 1.0. I think that the main culprit here is that we PAD left, but this will be addressed in another PR
NOTE2: this was done with bsz=groupsize=4. This is very small. Other setups would probably be more efficient/stable.

## Goal

This PR modernizes the losses in Forge, bringing SOTA implementations and providing guidance on how new losses should be added moving forward.

It is part of a bigger ongoing effort, which is to improve UX and prepare Forge for multiturn/tool call.

## Context

Currently Forge uses `simple_grpo_loss`. GRPO has come a long way and many variations have been implemented. This PR implements dr-GRPO, DAPO, SAPO, GSPO and CISPO. A proper doc will be created, but losses are self documenting for now.

## What is being changed

Mainly, only three things:
1. Addition of these losses
2. Using loss_mask instead of slicing the response.
e.g., instead of doing: 
```loss(input[-response_len:], targets[-reponse_len:])```
We do:
```loss(input, targets, loss_mask```)

which is used as 
```(loss_per_token*loss_mask).sum()```

This is more robust, but also necessary for multiturn, where we have interleaved trainable / not-trainable sequences

3. inference logprobs is added to the episode and made available to the losses to compute importance sampling (a weight of logprob_current / log_prob_old to correct for offpoliciness)

## Design choices

This is what a loss looks like:
```
class GRPOLoss(PolicyGradientLoss, BaseLossConfig):
    clip_low: Annotated[float, Field(ge=0, le=1)] = 0.2
    clip_high: Annotated[float, Field(ge=0, le=1)] = 0.2
    beta: Annotated[float, Field(ge=0)] = 0.1
    agg_type: AggType = "fixed_horizon"

    def __call__(
        self,
        logits: torch.Tensor,  # (B, S, V)
        target_ids: torch.Tensor,  # (B, S)
        advantages: torch.Tensor,  # (B, S)
        old_logprobs: torch.Tensor,  # (B, S)
        loss_mask: torch.Tensor,  # (B, S)
        ref_logprobs: torch.Tensor | None = None,  # (B, S) or None
        loss_scale: torch.Tensor | None = None,
        *args,
        **kwargs,
    ) -> LossOutput:
        logprobs, lp_m = compute_logprobs(logits, target_ids)
        entropy, ent_m = compute_entropy(logits, loss_mask)  # logging only
        ratio, log_ratio, ratio_m = compute_ratio(
            logprobs, old_logprobs, loss_mask, ratio_type="token"
        )
        pg_loss, clip_m = pg_ppo_clip(
            ratio, advantages, loss_mask, self.clip_low, self.clip_high
        )

        kl_m: list[Metric] = []
        if self.beta > 0:
            if ref_logprobs is None:
                raise ValueError("ref_logprobs required when beta > 0")
            kl, kl_m = compute_kl(logprobs, ref_logprobs, loss_mask)
            pg_loss = pg_loss + self.beta * kl

        loss, agg_m = aggregate(pg_loss, loss_mask, self.agg_type, loss_scale)

        return LossOutput(loss, lp_m + ent_m + ratio_m + clip_m + kl_m + agg_m)
```

Key decisions:
1. The loss is the config. It inherits a pydantic class (BaseLossConfig). The main motivation is 
- Field validation e.g. `clip_low: Annotated[float, Field(ge=0, le=1)] = 0.2`. We dont have to manually check anything.
- Auto complete (e.g. for enums)

Other options would be a) Use the traditional __init__ + args, losing the advantages of pydantic; b) Pass a `DAPOLossConfig`, duplicating the number of classes;

2. Inherit from `PolicyGradientLoss`. It does virtually nothing. Its just so we can do `isinstance(loss, PolicyGradientLoss)`. In the future, we will have packed dataset, and this will require PackedPolicyGradientLoss

3. The loss is expressed in utilities, e.g. `compute_ratio`, `aggregate`, etc. Many of those are shared across losses. Some are unique.

4. Each loss has its own class, e.g. `DAPOLoss`, `GRPOLoss`, etc. Another alternative would be to have an uber loss with all flags under the sun. This is, in fact, very tempting and is the strategy adopted by multiple frameworks such as `VeRL`, `TRL` and `Prime-RL.` 

The motivation is simple: There are many useful tricks that are not loss-specific. Lets say that there is some `EntropyTrick`. If we want to add it to Forge, we either have to a) add it to every loss; b) create a `GRPOWithEntropyTrick`.

In this proposed design, we will have to do `GRPOWithEntropyTrick`, and let the users change/combine losses and tricks as they prefer.

Pros:
- Understanding and maintaining a loss is simple
- Users can easily reuse and combine components
- Since there is a new trick every couple of weeks, our library is shielded from having a loss with more if/else than atoms in the universe.

Cons:
- Add a new loss just to expose a trick
- Current losses do NOT benefit from the trick
- If users want to use trick X, Y, Z, they have to manually understand and combine them. This could be mitigated by us offering some loss that combines the main ones, for example, or by providing solid documentation + guidance.

## WIP for this PR

[x] Unit tests - I have tested all losses vs other framework implementations. But actual unit tests with numerics that dont rely on external imports is missing
[x] running every experiments with every loss for llama 8b
[x] Delete old files/unit tests
[x] Some rough edges, like naming of logged metrics

## Follow for other PRs

[ ] Many...There are many things i want to change, but the current changes is just enough to get it to work. Examples:
- Loss normalization (take into account valid tokens across all dp_ranks, not only local rank)
- Advantage computation
- Refactor of the Episode class
- Collate refactor
- Rollout refactor
- Type hints for the trainer
- Expose the loss in the config (hydra? Registry?)

### DIFF
--- a/tests/rl/loss/test_gspo.py
+++ b/tests/rl/loss/test_gspo.py
@@ -42,3 +42,56 @@ class TestGSPOLoss:
 
         output.loss.backward()
         assert_close(logits.grad.norm(), torch.tensor(0.158423))
+
+    def test_zero_advantages(self, inputs):
+        d = inputs
+        advantages = torch.zeros_like(d["advantages"])
+
+        loss_fn = GSPOLoss()
+        output = loss_fn(
+            logits=d["logits"],
+            target_ids=d["target_ids"],
+            advantages=advantages,
+            generator_logprobs=d["generator_logprobs"],
+            loss_mask=d["loss_mask"],
+        )
+
+        assert output.loss.isfinite()
+
+    def test_empty_mask(self, inputs):
+        """Loss should be finite (zero) when mask is all zeros (no trainable tokens)."""
+        d = inputs
+        empty_mask = torch.zeros_like(d["loss_mask"])
+
+        loss_fn = GSPOLoss()
+        output = loss_fn(
+            logits=d["logits"],
+            target_ids=d["target_ids"],
+            advantages=d["advantages"],
+            generator_logprobs=d["generator_logprobs"],
+            loss_mask=empty_mask,
+        )
+
+        assert output.loss.isfinite()
+        assert output.loss == 0.0
+
+    def test_empty_sequence(self):
+        """Loss should be zero when sequence length is 0."""
+        B, V = 2, 10
+        logits = torch.empty(B, 0, V)
+        target_ids = torch.empty(B, 0, dtype=torch.long)
+        advantages = torch.empty(B, 0)
+        generator_logprobs = torch.empty(B, 0)
+        loss_mask = torch.empty(B, 0)
+
+        loss_fn = GSPOLoss()
+        output = loss_fn(
+            logits=logits,
+            target_ids=target_ids,
+            advantages=advantages,
+            generator_logprobs=generator_logprobs,
+            loss_mask=loss_mask,
+        )
+
+        assert output.loss.isfinite()
+        assert output.loss == 0.0

### CODE REVIEW

File: apps/grpo/main.py [hunk 1]
```diff
@@ -181,7 +140,31 @@ async def continuous_rollouts():
                 (group_size, max_req_tokens + max_res_tokens),
                 dtype=torch.long,
             )
+            seq_len = max_req_tokens + max_res_tokens
```
felipemello1, line 169: the code below adds `response_mask` and `inference_logprobs` to the episode, so it can be used by the loss

src/forge/actors/trainer/titan.py
```diff
@@ -124,16 +125,36 @@ def forward_backward(
         model_parts = self.engine.model_parts
         parallel_dims = self.engine.parallel_dims
         optional_context_parallel_ctx = None
+
+        # Create shifted target_ids for next-token prediction
+        # target_ids[i] = input_ids[i+1], with loss_mask applied
+        targets["target_ids"] = create_shifted_targets(
+            inputs["tokens"], targets.get("loss_mask")
+        )
```
felipemello1, line 131: before we had this as an option of `compute_logprobs`. Now this is a responsibility of the `trainer`. Another option is to put this in the `Episode` itself. Happy to change it.

src/forge/actors/trainer/titan.py
```diff
@@ -124,16 +125,36 @@ def forward_backward(
...
+                    loss = loss_output.loss
+
+                # Record metrics from loss output
+                for metric in loss_output.metrics:
+                    value = (
+                        metric.value.item()
+                        if isinstance(metric.value, torch.Tensor)
+                        else metric.value
+                    )
+                    record_metric(metric.key, value, metric.reduction, metric.timestamp)
```
felipemello1, line 150: we avoid calling 'record_metrics' inside of the loss so that the loss is compile friendly. Instead, we just collect all the Metrics, and log them later.

src/forge/actors/reference_model.py
```diff
@@ -99,9 +98,9 @@ def __post_init__(self):
         self.rank = current_rank().rank
         self.size = math.prod(current_size().values())
 
-        self.compute_log_probs = compute_logprobs
+        self.compute_logprobs = compute_logprobs
```
felipemello1, line 98: this is a mess in our code, in some places its logprobs, others are log_probs. I prefer with the "_", but will leave it for another PR

src/forge/actors/reference_model.py
```diff
@@ -131,23 +130,20 @@ async def setup(self):
 
     @endpoint
     async def forward(
-        self, input_ids: torch.Tensor, max_req_tokens: int, return_logprobs: bool
+        self, input_ids: torch.Tensor, return_logprobs: bool = True
```
felipemello1, line 130: we dont need max_req_tokens anymore. We compute the logprob for the whole sequence, not just the response.

It is not as optimal, but its more straight forward for every tensor to always be [bsz, seq_len].

src/forge/actors/reference_model.py
```diff
@@ -178,24 +174,17 @@ async def forward(
                 with self.engine.maybe_enable_amp:
                     with torch.inference_mode():
                         logits = self.model(input_ids)
-        self.step += 1
```
felipemello1, line 174: The logic stays the same as before, but:

1. we do create_shifted_targets outside of the loss function
2. We call the loss under `self.engine.train_context`, which already have `loss_parallel`. I just simplified it.

Please review

src/forge/observability/metrics.py
```diff
@@ -87,22 +87,20 @@ def accumulator_class(self):
...
+    """Container for metric data including key, value, reduction type, and timestamp."""
 
     key: str
     value: Any
     reduction: Reduce
     timestamp: float | None = None
 
-    def __post_init__(self):
-        if self.timestamp is None:
-            self.timestamp = time.time()
```
felipemello1, line 101: had compile issues. Moved it to `record_metric`, so that we can store `Metric` without issues

src/forge/rl/collate.py
```diff
@@ -27,21 +27,27 @@ def collate(
...
 
-        pad_id = batch[0].pad_id
-        mask = response != pad_id
+        old_logprobs = torch.stack([e.old_logprobs for e in batch])
+        loss_mask = torch.stack([e.loss_mask for e in batch])
 
-        input = {"tokens": torch.cat([request, response], dim=1)}
+        input = {"tokens": input_ids}
         target = {
-            "response": response,
```
felipemello1, line 40: no more slicing logic. All tensors for our losses are of shape [bsz, seq_len], and loss_mask is the way to handle responses / trainable tokens / padding

src/forge/rl/collate.py
```diff
@@ -27,21 +27,27 @@ def collate(
...
 
-        ref_logprobs = [e.ref_logprobs for e in batch]
-        ref_logprobs = torch.stack(ref_logprobs).squeeze()  # [b x s]
+        input_ids = torch.cat([request, response], dim=1)
+        seq_len = input_ids.shape[1]
+
+        # ref_logprobs is optional - only stack if all episodes have it
+        ref_logprobs = None
+        if all(e.ref_logprobs is not None for e in batch):
+            ref_logprobs = torch.stack([e.ref_logprobs for e in batch])
```
felipemello1, line 35: all losses., except dr-grpo, dont expect ref_logprobs.

src/forge/rl/types.py
```diff
@@ -13,6 +13,12 @@
 from forge.data_models.completion import Completion
 
 
+# TODO: Deprecated fields - will be removed in Episode refactor PR
```
felipemello1: TODO for myself: remove this comment. Will address in another PR.

tests/sandbox/toy_rl/sumdigits.py
felipemello1, line 1: note for myself: i edited a bunch of log_probs -> logprobs here by mistake. Only change here should be to import compute_logprobs from the new losses

src/forge/rl/losses.py
felipemello1, line 1: bulk of this PR, have fun

src/forge/rl/losses.py
```diff
@@ -0,0 +1,1049 @@
...
+
+        Returns:
+            LossOutput
+        """
+        pass
+
+    @property
+    def name(self) -> str:
+        """Return the loss name (class name without 'Loss' suffix)."""
+        return self.__class__.__name__.replace("Loss", "")
```
felipemello1: note for myself: delete this

src/forge/rl/losses.py
```diff
@@ -0,0 +1,1049 @@
...
+        )
+        pg_loss, gate_m = pg_soft_gate(
+            ratio, advantages, loss_mask, self.tau_pos, self.tau_neg
+        )
+        loss, agg_m = aggregate(pg_loss, loss_mask, self.agg_type, loss_scale)
+
+        return LossOutput(loss, lp_m + ent_m + ratio_m + gate_m + agg_m)
+
+
+class REINFORCELoss(PolicyGradientLoss, BaseLossConfig):
```
felipemello1: will probably delete this

apps/grpo/main.py
```diff
@@ -191,7 +174,10 @@ async def continuous_rollouts():
                     request=prompt,
                     response=response.text,
                     completion=response,
+                    old_logprobs=old_logprobs,
```
JenniferWang: I remember there was a PR that adds `old_logprobs` -- but what is this?

felipemello1: You have 3 logprobs:
1. Current model
2. Inference model --> this is the old_logprob. Its only old if you are doing off-policy
3. Reference model (never trained)

You can use them to get ratios, e.g. current_logprob / inference_logprob or current_logprob/ref_logprob, which is used by the losses (check `pg_ppo_clip` or KL divergence (check `compute_kl`)

image

felipemello1: i saw one library naming it inferece_logprobs. Perhaps we should do that too. But in general libraries call it old_logprob

krammnic:
Let me start with points related to adding new features (the "entropy trick" part of the RFC.

Solutions that require:

a) creating a new class for each added feature;
b) introducing a general “god” loss object (this is especially problematic and leads to a TRL-style design);
c) adding feature-specific checks inside existing loss implementations;

violate the Open–Closed Principle and therefore should not be used in a proper design.

Instead, I propose introducing either:
a) middleware-style hooks, where at each stage of the loss computation pipeline all hooks registered for that stage are applied;
b) or a decorator pattern, where loss behavior is extended via composable wrappers.

A small set of pre-built combinations may be provided for convenience, while all other compositions should be supported via documentation rather than additional code paths.

krammnic:
General comment on `losses.py`.  I think it should be splited on several parts. 

UPD: I actually hope that the current split via comments inside it is actually future split by files.

krammnic:
I think it is cool idea to add this also: https://arxiv.org/abs/2601.05242

krammnic:
We need the possibility to apply loss only to X% highest entropy tokens. Where currently best X = 20. (NeurIPS 2025, Wang et al.)

joecummings:
> Let me start with points related to adding new features (the "entropy trick" part of the RFC.
> 
> Solutions that require:
> 
> a) creating a new class for each added feature; b) introducing a general “god” loss object (this is especially problematic and leads to a TRL-style design); c) adding feature-specific checks inside existing loss implementations;
> 
> violate the Open–Closed Principle and therefore should not be used in a proper design.
> 
> Instead, I propose introducing either: a) middleware-style hooks, where at each stage of the loss computation pipeline all hooks registered for that stage are applied; b) or a decorator pattern, where loss behavior is extended via composable wrappers.
> 
> A small set of pre-built combinations may be provided for convenience, while all other compositions should be supported via documentation rather than additional code paths.

For your proposed changes, can you provide a small code sample to help understand the exact approach you're advocating for?

felipemello1:
hey @krammnic , thanks for the review! 

1. Regarding using hooks and decorators, I think that decorators would not be feasible because usually you have a -> b, and if you try to introduce a -> c -> d with a decorator, e.g. some mask input that relies on 'a', i dont see how a decorator could handle that.

hooks on the other hand could work, but maybe it would get ugly and be hard to debug? Either way, i think that we can handle those in a follow up PR. Is that reasonable?

2. Regarding GDPO (https://arxiv.org/abs/2601.05242), this is a change in the advantage computation, so we can handle it in another PR as well. I do like the idea though!

3. Regarding the entropy mask, i like it too. In fact, there are many masking ideas....Entropy (qwen), gibberish filter (meta), negative advantage + high ratio (deepseek). I am thinking that i need a follow up PR just for masks, and maybe those could be handled with a flag? Because then we can do something like:

```mask = compute_all_masks(list_mask_fn, loss_mask)```

but even this would probably be tricky.

File: src/forge/rl/__init__.py [hunk 14]
```diff
@@ -7,6 +7,31 @@
 from forge.rl.advantage import ComputeAdvantages
 from forge.rl.collate import collate
 from forge.rl.grading import RewardActor
+from forge.rl.losses import (
```
joecummings: Let's not import this at top level. If people want to use losses then can call `from forge.rl.loss import XYZ`

src/forge/rl/collate.py
```diff
@@ -27,22 +27,29 @@ def collate(
...
 
         advantages = [e.advantage for e in batch]
         advantages = torch.tensor(advantages).unsqueeze(-1)  # [b x 1]
+        advantages = advantages.expand(-1, seq_len)  # [b x s]
 
-        pad_id = batch[0].pad_id
-        mask = response != pad_id
+        inference_logprobs = torch.stack([e.inference_logprobs for e in batch])
```
joecummings: nit: Can we call this `generator_logprobs` to match the convention we've chosen for Generator protocol?

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+
+class BaseLossConfig(BaseModel):
+    """Base configuration for all policy gradient losses."""
+
+    # extra="forbid": Raises error if user passes unknown fields (catches typos).
+    # arbitrary_types_allowed=True: Allows torch.Tensor and other non-JSON types in fields.
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
+
+
+class PolicyGradientLoss(ABC):
```
joecummings: If there's not meaningful structure or methods here, let's remove this ABC.

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+
+    return targets
+
+
+def compute_logprobs(
+    logits: torch.Tensor,
+    target_ids: torch.Tensor,
+    temperature: float = 1.0,
+    ignore_index: int = CROSS_ENTROPY_IGNORE_IDX,
+) -> tuple[torch.Tensor, list[Metric]]:
```
joecummings: Why does this *need* to return list[Metric] as well?

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+                key="loss/entropy/mean",
+                value=masked_mean(entropy, mask),
+                reduction=Reduce.MEAN,
+            ),
+        ]
+
+    return entropy, metrics
+
+
+# -----------------------------------------------------------------------------
```
joecummings: We don't need these barriers if it's just one function

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+# Ratio Computation
+# -----------------------------------------------------------------------------
+
+
+def compute_ratio(
+    logprobs: torch.Tensor,
+    inference_logprobs: torch.Tensor,
```
joecummings: again, might opt for `generator_logprobs` here.

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+      rewards are assigned (per-response). Lower variance for long sequences.
+      Uses reparameterization trick to maintain per-token gradient flow.
+
+    Reference (sequence): Zheng et al., "GSPO" (arXiv:2507.18071, 2025).
+
+    Args:
+        logprobs (torch.Tensor): Log probs from current policy (B, S).
+        inference_logprobs (torch.Tensor): Log probs from sampling policy (B, S).
+        mask (torch.Tensor): Valid token mask (B, S).
+        ratio_type (RatioType): "token" for per-token ratio, "sequence" for sequence-level.
```
joecummings: Include default here.

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+
+    return ratio, log_ratio, metrics
+
+
+# -----------------------------------------------------------------------------
+# Policy Gradient Strategies
+# -----------------------------------------------------------------------------
+
+
+def pg_ppo_clip(
```
joecummings: Why the policy gradient prefix?

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+                key="loss/clip/low_fraction",
+                value=masked_mean((clipped_low & neg_adv).float(), mask),
+                reduction=Reduce.MEAN,
+            ),
+        ]
+
+    return pg_loss, metrics
+
+
+def pg_dual_clip(
```
joecummings: Same question on all these about `pg` prefix

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+
+
+# -----------------------------------------------------------------------------
+# KL Divergence
+# -----------------------------------------------------------------------------
+
+
+def compute_kl(
+    log_policy: torch.Tensor,
+    log_ref: torch.Tensor,
```
joecummings:
```suggestion
    ref_logprobs: torch.Tensor,
```

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+    return pg_loss, metrics
+
+
+# -----------------------------------------------------------------------------
+# KL Divergence
+# -----------------------------------------------------------------------------
+
+
+def compute_kl(
+    log_policy: torch.Tensor,
```
joecummings:
```suggestion
    policy_logprobs: torch.Tensor,
```

src/forge/rl/losses.py
```diff
@@ -0,0 +1,983 @@
...
+    Estimator properties (for KL[policy, ref]):
+    - k1: Unbiased KL estimate, but E[grad k1] = 0 (useless for optimization).
+    - k2: Biased KL estimate, but E[grad k2] = grad KL (unbiased gradient).
+    - k3: Unbiased KL estimate with low variance. E[grad k3] = grad KL[ref, policy].
+
+    k3 is preferred for monitoring KL value. k2 is preferred when using KL as a
+    regularizer (gradient flows correctly). k1 is rarely used in practice.
+
+    Args:
+        log_policy (torch.Tensor): Log probs from current policy (B, S).
```
joecummings: Update here as well.

src/forge/rl/types.py
joecummings: Was Episode refactored? If so, just remove it in this PR.

src/forge/rl/types.py
```diff
@@ -24,10 +30,12 @@ class Episode:
     response: str | None = None
     # Processed data
     completion: Completion | None = None
-    ref_logprobs: torch.Tensor | None = None
+    inference_logprobs: torch.Tensor | None = None  # [seq_len]
```
joecummings: generator logprobs

src/forge/rl/losses.py
joecummings: Can we call this `loss.py`

tests/unit_tests/losses/test_rl_losses.py
```diff
@@ -0,0 +1,498 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Expected values generated using P2120280632"""
```
joecummings: Can we make this a gist so everyone can see it?

joecummings (commented):
Overall, this is an awesome step in the right direction for making our code cleaner and more reusable. I especially like all the small components that can be utilized to create common loss classes for RL functions.

Couple things:
* I will push back on the use of Pydantic and the PolicyGradient base class. The former might be a good call, but I don't want to be in a state where it's used some places and not the other. For the latter, I don't see the value yet of abstracting to such a concept. If/when other losses come into play, we can address it then. 
* Make sure naming is consistent all over between `generator` and `inference`. I would suggest sticking with generator since that's the convention we used for our actual vLLM instance.
* We REALLY need to think about how we want users to interact with this library. The main thing in my mind is what happens when a user wants to set up their own new RL thing with a slightly different loss and slightly different data. Right now, it's not clear how they would do that. It's out of scope for this PR, which provides some basic building blocks and common patterns for the general RL public, but if we want to serve researchers well, this should be top of mind.

🫡

src/forge/actors/trainer/titan.py
befelix: It would be nice if the trainer didn't impose too many specific requirements for variables names on the collate, so that it's easy to iterate. In our code, we do this alignment of the target logits as part of the loss function. This setup is quite nice for testing, since you can directly pass the output of the collate to the loss without having to instantiate a full trainer.

src/forge/rl/losses.py
felipemello1: just to keep the same pattern for everything. All other functions return it. I dont have a strong opinion here, but my intuition is that we should keep. I am ok with removing.

src/forge/rl/losses.py
felipemello1: easier to identify these functions. They have different names, but they play the same role.

src/forge/rl/losses.py
joecummings: Make sense - can you just add a small comment?

src/forge/actors/trainer/titan.py
felipemello1: thanks for the input @befelix ! I am keeping it in mind in a follow up refactor of the data flow that i am working on. My guiding principle here is that loss should focus on calculating loss, and not formatting the input data.

But we should make it very straight forward for you to format the input data however you want, without having to change N different functions.

src/forge/rl/loss/types.py
```diff
@@ -0,0 +1,37 @@
...
+    loss: torch.Tensor
+    metrics: list[Metric]
+
+
+class BaseLossConfig(BaseModel):
+    """Base configuration for all policy gradient losses."""
+
+    # extra="forbid": Raises error if user passes unknown fields (catches typos).
+    # arbitrary_types_allowed=True: Allows torch.Tensor and other non-JSON types in fields.
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
```
joecummings: What is model_config?

src/forge/rl/loss/primitives.py
```diff
@@ -0,0 +1,402 @@
...
+            default: "token".
+
+    Returns:
+        tuple[torch.Tensor, torch.Tensor, list[Metric]]: (ratio, log_ratio, metrics). Both
+            ratio and log_ratio are (B, S). For sequence type, values are broadcast from
+            per-sequence computation.
+    """
+    if ratio_type == "token":
+        log_ratio = logprobs - generator_logprobs.detach()
+        log_ratio = torch.clamp(log_ratio, min=-20.0, max=20.0)
```
joecummings: Clamping should happen after at the explicit direction of the loss, not here

src/forge/rl/loss/primitives.py
```diff
@@ -0,0 +1,402 @@
...
+        ratio = torch.exp(log_ratio)
+
+    elif ratio_type == "sequence":
+        token_log_ratio = logprobs - generator_logprobs.detach()
+        seq_lengths = mask.sum(dim=-1).clamp(min=1)
+        seq_log_ratio = (token_log_ratio * mask).sum(dim=-1) / seq_lengths
+
+        # Reparameterization: forward uses seq ratio, backward uses token grads
+        log_ratio = logprobs - logprobs.detach() + seq_log_ratio.detach().unsqueeze(-1)
+        log_ratio = torch.clamp(log_ratio, min=-20.0, max=20.0)
```
joecummings: clamp comment

src/forge/rl/loss/primitives.py
```diff
@@ -0,0 +1,402 @@
...
+        tuple[torch.Tensor, list[Metric]]: Per-token KL (B, S) and loss/kl/mean metric.
+    """
+    log_ratio = policy_logprobs - ref_logprobs.detach()  # log(π_θ / π_ref)
+
+    if kl_type == "k1":
+        kl = log_ratio
+    elif kl_type == "k2":
+        kl = 0.5 * log_ratio.square()
+    elif kl_type == "k3":
+        log_ratio_clamped = torch.clamp(-log_ratio, min=-20.0, max=20.0)
```
joecummings: k3 is not clamped by default

src/forge/rl/loss/primitives.py
```diff
@@ -0,0 +1,402 @@
...
+        ratio = torch.exp(log_ratio_clamped)  # π_ref / π_θ
+        kl = ratio - log_ratio_clamped - 1
+        kl = torch.clamp(kl, min=0.0, max=10.0)
```
joecummings: clamped

src/forge/rl/loss/ops.py
joecummings, line 1: nit: slight preference for calling this `ops` or something b/c I could consider types as primitives too.

tests/unit_tests/losses/test_rl_losses.py
joecummings, line 1: Technically, this test structure should mirror the loss folder exactly meaning one file per test.

tests/rl/loss/test_common.py
joecummings, line 1: Why is this it's own file and not just within each loss file? We don't need to consolidate tests - much better to just have the testing be clear. I imagine it's easier for CLAUDE to grep in a standard layout as well.

PR #700, opened 2026-01-13
Use Monarch Jobs API in launchers

daniellepintz:
## Description
Migrate `provisioner.py` and `launcher.py` to use the Monarch Jobs API. This offloads a lot of work we were previously doing to Monarch, since they have this functionality already in their Jobs API.

The "Launcher" is now a very thin wrapper over the Monarch Jobs API. This is needed because different Monarch "Job"s e.g. "SlurmJob" have different APIs for initialization.

### Slurm Launching
Run with controller locally:
```
python -m apps.grpo.main --config experimental/slurm/qwen3_8b.yaml
```

Run with the controller remotely:
```
./experimental/slurm/submit.sh qwen3_8b
```

### DIFF
--- a/experimental/slurm/README.md
+++ b/experimental/slurm/README.md
@@ -3,7 +3,7 @@
 First make sure you have followed the environment setup instructions in
 https://github.com/meta-pytorch/torchforge/blob/main/README.md.
 
-When running GRPO on Slurm, the "controller" (main script, grpo/main.py) can either run
+When running GRPO on Slurm, the "controller" (main script in grpo/main.py) can either run
 locally on your login node, and launch the workers to Slurm; or the controller can
 run on a remote node along with the workers.
 


experimental/slurm/README.md
```diff
@@ -1,10 +1,23 @@
...
 
-To run GRPO on Slurm, please use the `submit.sh` script in this directory.
+First make sure you have followed the environment setup instructions in
+https://github.com/meta-pytorch/torchforge/blob/main/README.md.
+
+When running GRPO on Slurm, the "controller" (main script, grpo/main.py) can either run
+locally on your login node, and launch the workers to Slurm; or the controller can
+run on a remote node along with the workers.
+
+To run with controller locally (good for debugging):
```
allenwang28: to be precise, let's call this "interactive"

experimental/slurm/README.md
```diff
@@ -1,10 +1,23 @@
...
+```
+python -m apps.grpo.main --config experimental/slurm/qwen3_8b.yaml
+```
+
+To run with the controller remotely use the `submit.sh` script:
```
allenwang28: this would be called "batch"

Ilia asked for a mode where we could `salloc` a set of nodes and then run repeatedly across those nodes though - we should see if submit.sh can be adapted to run on a set of pre-allocated nodes

experimental/slurm/README.md
```diff
@@ -1,10 +1,23 @@
...
 
-Usage:
 ```
 ./experimental/slurm/submit.sh qwen3_8b
 ./experimental/slurm/submit.sh qwen3_32b
 ./experimental/slurm/submit.sh qwen3_30b_a3b
 ```
+
+Note, the controller and workers will be launched in two different Slurm jobs.
```
allenwang28: actually I think if we can figure out the salloc path, that will make this not be the case.

src/forge/controller/launcher.py
```diff
@@ -31,45 +31,128 @@ def __init__(
...
-                "qos": self.cfg.qos,
-                "time": "72:00:00",
-            },
-            appdef=appdef,
-            workspace=monarch.tools.config.workspace.Workspace(dirs=[temp_workspace]),
+        # Collect all mesh requirements from config
+        meshes = {}
+
+        # Add services that need remote hosts
+        for service_name, service_cfg in self.cfg.services.items():
```
allenwang28: for service and actor extraction, we should write a dedicated function that takes the service/actor definition and converts it into meshes, i.e. `get_meshes_from_config` or something similar.

src/forge/controller/launcher.py
```diff
@@ -31,45 +31,128 @@ def __init__(
...
+            if hosts and hosts > 0:
+                mesh_name = actor_cfg.mesh_name or actor_name
+                meshes[mesh_name] = hosts
+
+        # If no remote resources needed, skip job creation
+        if not meshes:
+            return
+
+        # Prepare slurm_args from config (only for args without dedicated parameters)
+        slurm_args = []
```
allenwang28: Right now this is implemented Slurm specific,  but this functionality would be useful across other Job definitions too. We should try and find the right abstraction point that lets us basically end up with a factory method like this:

```
if self.cfg.provisioner.type == "slurm":
    self._job = SlurmJob(...)
elif self.cfg.provisioner.type == "mast":
    self._job = MastJob(...)
```

I think for the slurm_args we should put that in the top level YAML definition

src/forge/controller/provisioner.py
```diff
@@ -251,34 +250,31 @@ async def initialize(self):
         if self.launcher is not None:
             await self.launcher.initialize()
 
-    async def create_host_mesh(self, name: str, num_hosts: int) -> HostMesh:
-        """Creates a remote server and a HostMesh on it."""
+    async def create_host_mesh(self, name: str, num_hosts: int) -> tuple[HostMesh, Any]:
```
allenwang28: maybe this can become `get_host_mesh` since we've now pre-allocated all of our hosts

src/forge/controller/provisioner.py
```diff
@@ -251,34 +250,31 @@ async def initialize(self):
...
+        """Creates a remote HostMesh using the launcher.
+
+        Args:
+            name: Name for the host mesh
+            num_hosts: Number of hosts to allocate
+
+        Returns:
+            A tuple of (HostMesh, allocation_handle) where:
+            - HostMesh is the allocated resource ready to use
+            - allocation_handle is an opaque handle for cleanup (managed by launcher)
```
allenwang28: if the launcher now contains the job, I think the launcher can handle all clean ups which should simplify the provisioner logic

generally I think we can just ask Claude, "now that the launcher has access to the Job, how can we simplify the role of the provisioner?"

experimental/slurm/README.md
daniellepintz: yea, this is not ideal but was what i was able to get working currently. will def continue iterating though

src/forge/controller/launcher.py
daniellepintz: > We should try and find the right abstraction point that lets us basically end up with a factory method like this

do you mean not even have the individual launcher classes like `SlurmLauncher`?

> I think for the slurm_args we should put that in the top level YAML definition

currently it is here: https://github.com/meta-pytorch/torchforge/blob/7078ed359ffc4f774605bc210e15cffa9915ed5d/experimental/slurm/qwen3_30b_a3b.yaml#L17-L18

do you mean more top level than this?

and are you saying the user should be expected to define `slurm_args` entirely? would be something like:
```
[
                f"--account={slurm_account}",
                f"--qos={slurm_qos}",
            ]
```

allenwang28: > do you mean not even have the individual launcher classes like SlurmLauncher?

I think the provisioner should work against any Job definition provided by Monarch, the only difference is how the Jobs are instantiated

```
provisioner:
  launcher: slurm
  slurm_args:
    account: ...
    qos: ...
  memMB: 2047962
  cpu: 192
  account: agentic-models
  qos: h200_capabilities_shared
```

something like that?

thedavekwon: Only tricky thing right now is that Monarch Job APIs configuration differ from scheduler to scheduler, and information needed it to create Job instances.  For example, MastJob and KubernetesJob both have `add_mesh` API to configure each mesh compared to SlurmJob that initialize all meshes in `__init__.py`. Also, MastJob and KubernetesJob allows configuring resource per mesh unlike SLURM. 

I would layer it so that job creation has specific launcher (SlurmLauncher, KubernetesLauncher, MastLauncher), but after job creation, we should have uniform layer for interacting with these jobs to get constructed hostmeshes.

src/forge/controller/launcher.py
```diff
@@ -31,45 +31,128 @@ def __init__(
...
+        job_state = self._job.state(cached_path=None)
+        logger.info(
+            f"Job state received! Extracting HostMeshes for {list(meshes.keys())}"
         )
-        alloc = RemoteAllocator(
-            world_id=name,
-            initializer=TorchXRemoteAllocInitializer(server_info.server_handle),
+
+        # Store all HostMeshes by name (like node0_host, node1_host in the example)
+        for mesh_name in meshes.keys():
```
thedavekwon: why not directly use `self._job.state`? I don't think we need `self._host_meshes` any more.

src/forge/controller/launcher.py
daniellepintz: Thanks @thedavekwon! Yea I noticed that difference between SlurmJob and MastJob APIs :/ do you know why the difference exists, e.g. with `meshes` vs `add_mesh`? Is there any chance of changing this on the monarch side?

romilbhardwaj:
Just dropping in to say we ([SkyPilot](https://github.com/skypilot-org/skypilot)) are excited to see this PR! 🙌

We already have a [JobTrait implementation](https://github.com/meta-pytorch/monarch/tree/main/examples/skypilot) that enables running Monarch on Kubernetes. Once this PR lands, integrating SkyPilot with TorchForge becomes straightforward - we'd essentially need to add a `SkyPilotLauncher` that mirrors the `SlurmLauncher` pattern and swap `SlurmJob` for `SkyPilotJob`.

This would allow TorchForge users to run on any cloud/infra that SkyPilot supports, including Kubernetes clusters.

src/forge/controller/launcher.py
thedavekwon: We can make it uniform, but I doubt that it is useful since configuration for each scheduler is very different. I would say layer it so that you only require methods in JobTrait and make each scheduler launcher (SlurmJob, MastJob, KubernetesJob) responsible to provide Job.

src/forge/controller/provisioner.py
daniellepintz: I kind of ended up doing the opposite, simplifying the launcher api a lot (e.g. removing `get_allocator`) and moving more logic into provisioner calling the `job` since the calls to jobs api should be shared across slurm and other launchers

lmk what you think!

src/forge/controller/launcher.py
daniellepintz: done! thank you!

experimental/slurm/README.md
```diff
@@ -1,8 +1,21 @@
...
+locally on your login node, and launch the workers to Slurm; or the controller can
+run on a remote node along with the workers.
+
+## To run "interactively":
+(controller runs locally, good for debugging)
+```
+python -m apps.grpo.main --config experimental/slurm/qwen3_8b.yaml
+```
+
+## To run "in batch":
```
allenwang28:
```suggestion
## To run in batch mode:
```

experimental/slurm/README.md
```diff
@@ -1,8 +1,21 @@
...
 
-To run GRPO on Slurm, please use the `submit.sh` script in this directory.
+First make sure you have followed the environment setup instructions in
+https://github.com/meta-pytorch/torchforge/blob/main/README.md.
+
+When running GRPO on Slurm, the "controller" (main script, grpo/main.py) can either run
+locally on your login node, and launch the workers to Slurm; or the controller can
+run on a remote node along with the workers.
+
+## To run "interactively":
```
allenwang28:
```suggestion
## To run in interactive mode:
```

(admittedly a nit)

src/forge/controller/launcher.py
```diff
@@ -6,70 +6,108 @@
...
+        slurm_args = [f"--{key}={value}" for key, value in self.cfg.slurm_args.items()]
+
+        # Create a single SlurmJob with all meshes
+        logger.info(f"Creating SlurmJob with meshes: {meshes}")
+        job = SlurmJob(
+            meshes=meshes,  # e.g., {"generator": 1, "trainer": 2, "ref_model": 1}
+            gpus_per_node=self.cfg.gpus_per_node,
+            cpus_per_task=self.cfg.cpus_per_task,
+            mem=self.cfg.mem,
+            time_limit="72:00:00",  # Default to 72 hours
```
allenwang28: should this additionally be configurable?

src/forge/controller/provisioner.py
```diff
@@ -310,12 +308,17 @@ async def get_proc_mesh(
...
                 If this is None, it uses the local host.
                 This behavior may change in the future.
+            mesh_name: Name of the pre-allocated mesh to use.
+                Required for remote allocations (when num_hosts > 0).
+                Must match a mesh name defined in the launcher config.
             host_mesh: The host mesh to allocate the process on.
-                If None, a new host mesh will be created.
-            port: The distributed port to use.
-                If None, a port will be detected.
+                If None, a new host mesh will be created from the pre-allocated mesh.
```
allenwang28:
```suggestion
                If None, will use an available host mesh from the pre-allocated pool.
```

src/forge/controller/provisioner.py
```diff
@@ -327,15 +330,15 @@ async def get_proc_mesh(
         is_remote = num_hosts is not None and num_hosts > 0
 
         async with self._lock:
-            server_name = None
             if is_remote:
                 if mesh_name is None:
-                    created_hosts = len(self._server_names)
-                    mesh_name = f"alloc_{created_hosts}"
+                    # TODO: make mesh_name a required arg
```
allenwang28: is doing this too invasive? Wondering if we can do this now or if we can fast follow this PR

experimental/slurm/README.md
daniellepintz: thank you, appreciate the nits!! i wasn't sure of the exact terminology so thank you

daniellepintz: are these terms Monarch terms?

src/forge/controller/launcher.py
daniellepintz: I think I am going to remove this for now, added this when was debugging something. will remove for now and it will use slurm's default time limit. can always add back later

src/forge/controller/provisioner.py
daniellepintz: my understanding is mesh_name will never be None, at least in how we are currently using forge.

yeah I can def do this now (make mesh_name required), kind of wanted to check with you first

experimental/slurm/README.md
allenwang28: nah I think they are general "Job" terms

src/forge/controller/launcher.py
```diff
@@ -6,70 +6,107 @@
...
+
+        # Register cleanup handler
+        atexit.register(job.kill)
+
+        # Wait for job allocation
+        logger.info("Getting job state (this will block until nodes are allocated)...")
+        job_state = job.state(cached_path=None)
+
+        logger.info("SlurmLauncher initialization complete.")
+        return job, job_state
```
thedavekwon, line 110: Q: Why does it have to return both `job` and `job_state`? Why not just `job`?

thedavekwon (approved):
Looks good! Thanks!

PR #704, opened 2026-01-16
Hn/sft recipe bugfix

hieuchi911:
# Description

The current SFT recipe `ForgeSFTRecipe` does not apply gradient clipping, causing the training loss to start increasing after some point due to unstable training. Plus, `seq_len` is not being applied during tokenization.

## Test Plan

Comparing training runs with and without gradient clipping, the run with gradient clipping shows a stabilized training loss curve, while the run without gradient clipping shows increasing or unstable loss: https://api.wandb.ai/links/p1ai/qrnczlpx

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ cython_debug/
 # SLURM logs
 slogs/
 slurm-*
+slurm_*
 
 # Celery stuff
 celerybeat-schedule
@@ -196,11 +197,6 @@ cover/
 wandb/
 
 assets/wheels/vllm*.whl
-
-# DCP artifacts
-forge_dcp_tmp/
-demo_top_down.md
-
 
 # enroot / sqsh
 *.sqsh
--- a/apps/grpo/llama3_8b.yaml
+++ b/apps/grpo/llama3_8b.yaml
@@ -110,6 +110,7 @@ ref_model:
     context_parallel_degree: 1
     expert_parallel_degree: 1
   checkpoint:
+    enable: true
     initial_load_path: hf://${model}
     initial_load_in_hf: true
 
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -97,8 +97,14 @@ async def main(cfg: DictConfig):
     # ---- Global setups ---- #
     provisioner = None
     if cfg.get("provisioner", None) is not None:
+        # Create launcher config with services and actors for pre-allocation
+        launcher_config = LauncherConfig(
+            **cfg.provisioner,
+            services=cfg.get("services", {}),
+            actors=cfg.get("actors", {}),
+        )
         provisioner = await init_provisioner(
-            ProvisionerConfig(launcher_config=LauncherConfig(**cfg.provisioner))
+            ProvisionerConfig(launcher_config=launcher_config)
         )
     else:
         provisioner = await init_provisioner()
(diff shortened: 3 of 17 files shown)


src/forge/data/tokenizer.py
```diff
@@ -229,6 +229,7 @@ def __init__(
         generation_config_path: str | None = None,
         chat_template_path: str | None = None,
         truncation_type: str = "right",
+        max_seq_len: int | None = None,
```
felipemello1, line 233: please add to the docstring

src/forge/data/tokenizer.py
```diff
@@ -259,6 +260,8 @@ def __init__(
         for token in self.special_tokens:
             self.special_tokens_mapping[token] = self.base_tokenizer.encode(token)
 
+        self.max_seq_len = max_seq_len
```
felipemello1: lets put this at the top of the function

src/forge/data/tokenizer.py
```diff
@@ -294,6 +297,10 @@ def tokenize_messages(
         add_eos: bool = True,
         max_seq_len: int | None = None,
     ) -> tuple[list[int], list[bool]]:
+
+        if max_seq_len is None and self.max_seq_len is not None:
+            max_seq_len = self.max_seq_len
```
felipemello1: since the function takes max_seq_len as an arg, lets remove this check and pass the max_seq_len in the _call_ instead

src/forge/data/tokenizer.py
felipemello1:
```suggestion
        tokens, mask = self.tokenize_messages(messages, max_seq_len=self.max_seq_len)  # TODO: add_end_tokens
```

apps/sft/main.py
```diff
@@ -281,9 +287,20 @@ def train_step(self, batch) -> None:
...
 
+        grad_norm = dist_utils.clip_grad_norm_(
+            [p for m in self.model_parts for p in m.parameters()],
+            self.job_config.training.max_norm,
+            foreach=True,
+            pp_mesh=(
+                parallel_dims.world_mesh["pp"] if parallel_dims.pp_enabled else None
+            ),
+            ep_enabled=parallel_dims.ep_enabled,
+        )
```
felipemello1, line 298:
```suggestion
        grad_norm = dist_utils.clip_grad_norm_(
            [p for m in self.model_parts for p in m.parameters()],
            self.job_config.training.max_norm,
            foreach=True,
            pp_mesh=parallel_dims.get_optional_mesh("pp"),
            ep_enabled=parallel_dims.ep_enabled,
        )
```

apps/sft/main.py
```diff
@@ -114,6 +114,11 @@ async def setup(self):
         # Load training datasets
         logger.info("Setting training datasets")
         train_datasets_config = self.job_config.training.datasets
+        if self.job_config.training.seq_len is not None:
+            self.max_seq_len = self.job_config.training.seq_len
+        else:
+            self.max_seq_len = None
```
felipemello1: i think that we can delete this and use `self.job_config.training.seq_len` directly. Wdyt?

apps/sft/main.py
hieuchi911: Hi @felipemello1 I found that `parallel_dims.get_optional_mesh` is not available in `torchtitan==0.2.0` but `torchtitan==0.2.1`, and `torchtitan==0.2.1` requires `torch==2.11.0`? Either way I'm getting dependencies conflict when trying to install these versions

apps/sft/main.py
hieuchi911: That makes sense, this extra `self.max_seq_len` is not necessary, it's more of a tokenizer thing so having that in `HuggingFaceModelTokenizer` should be enough, I removed it and everything works fine

apps/sft/main.py
felipemello1: sry, i was not aware! we are updating the torchtitan version, but we can keep as it was for now. my bad

felipemello1:
can you please do the meta CLA check so i can merge when ready? will finish reviewing it on monday. Ty for the changes!

PR #705, opened 2026-01-16
Generic Replay Buffer

befelix:
## Description

Currently the replay buffer only works for `Episode` inputs, which makes it rather inflexible for downstream adaptations of `torchforge`. This PR proposes to make the `ReplayBuffer` generic in terms of what data is stored.

The main downside of this generality is that one needs to explicitly specify the policy age when adding data.

## Test plan

n/a

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -269,7 +269,9 @@ async def main(cfg: DictConfig):
             advantages = await compute_advantages.compute.call_one(episodes)
             for episode, advantage in zip(episodes, advantages):
                 episode.advantage = advantage
-                await replay_buffer.add.call_one(episode)
+                await replay_buffer.add.call_one(
+                    episode, policy_version=episode.policy_version
+                )
 
                 sample = episode.to_dict(exclude=["ref_logprobs", "completion"])
                 sample["score"] = sample["reward"]
--- a/src/forge/actors/replay_buffer.py
+++ b/src/forge/actors/replay_buffer.py
@@ -9,7 +9,7 @@ import random
 from collections import deque
 from dataclasses import dataclass
 from operator import itemgetter
-from typing import Any, Callable
+from typing import Any, Callable, cast, Generic, TypeVar
 
 from forge.controller import ForgeActor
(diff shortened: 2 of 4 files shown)


felipemello1:
thanks for the PR! This is good feedback. I need to look into it a bit more in detail, but i imagine that this would also require changes in the collate fn.  I am working on refactoring some components, including the collate and episode class. Do you mind if i get back to you next week?

befelix:
> I need to look into it a bit more in detail, but i imagine that this would also require changes in the collate fn.

No rush, but at least in the current PR the collate for the GRPO app should be the same; it only adds the flexibility to do something else.

>  I am working on refactoring some components, including the collate and episode class. 

Regarding collates; we also have a custom implementation of both that remove `torchforge`'s left-padding to reduce log-prob mismatches between vllm and torch titan (at least for the default qwen models). This is also something we could upstream if you'd like, but it's a larger change since log_prob extraction doesn't work as conveniently anymore. Not sure if you'd want this or would prefer to directly switch to sequence packing.

felipemello1:
Thanks @befelix ! I became aware yesterday that we are doing left padding -.-`. Working on it now. Also, fyi,
fixed this today: https://github.com/meta-pytorch/torchforge/pull/706
And am going to merge our loss PR soon: https://github.com/meta-pytorch/torchforge/pull/699

Since you guys are doing so much work with Forge, it would be nice to touch bases and gather some feedback, see what we can prioritize to help you guys. (its fine if you guys are busy and cant at the moment). Just let me know and I can shoot you an email and schedule some 30min meeting

felipemello1:
> Not sure if you'd want this or would prefer to directly switch to sequence packing.

I am a big fan of sequence packing, but there are so many other bottlenecks that we haven't been able to get there yet. We will though.

befelix:
It would be nice to chat. Can you ping me at felix.berkenkamp@aleph-alpha-research.com to  schedule a call?

befelix:
quick ping @felipemello1 - we'd be happy to chat.

felipemello1:
hey @befelix , sorry about the delay! Sent the email. Ty :)

PR #717, opened 2026-01-20
Hotfix: grpo local host mesh fallback when no launcher

Xinyu-Kang:
## Summary
Restore single-node GRPO runs when no launcher is configured. Regression introduced by commit ef3c162 caused local GRPO runs to fail.

## Changes
- Use `this_host()` as the HostMesh fallback when `provisioner.launcher` is None.

### DIFF
--- a/apps/grpo/main.py
+++ b/apps/grpo/main.py
@@ -14,6 +14,7 @@ import torchstore as ts
 import yaml
 from apps.grpo.data import DatasetActor
 from apps.grpo.grading import MathReward, ThinkingReward
+from monarch.actor import this_host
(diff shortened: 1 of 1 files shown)


daniellepintz:
Hi @Xinyu-Kang thanks so much for catching this and sending this PR! I have a fix in https://github.com/meta-pytorch/torchforge/pull/718 which should allow us to still use the `get_host_mesh` API

Xinyu-Kang:
Hi @daniellepintz, thanks for the quick fix! If #718 covers this, I’m happy to close this PR to avoid duplication. Appreciate it.

PR #746, opened 2026-01-30
Create a HostMesh per replica

daniellepintz:
Summary: In the [Jobs API refactor](https://github.com/meta-pytorch/torchforge/pull/700) we created one HostMesh per service, but it should really be one HostMesh per replica.

Differential Revision: D91890958


allenwang28 (approved):

